PR #146, opened 2014-06-27
Introduce CompareObject matcher

dmalikov:
It would be nice to be able to do something like `@(2,1,3) | Should CompareObject @(1,2,3)`

### DIFF
--- a/Functions/Assertions/CompareObject.Tests.ps1
+++ b/Functions/Assertions/CompareObject.Tests.ps1
@@ -1,0 +1,15 @@
+Set-StrictMode -Version Latest
+
+InModuleScope Pester {
+    Describe "CompareObject" {
+        It "return true  for things that are compared as equals" {
+            PesterCompareObject @() @() | Should Be $true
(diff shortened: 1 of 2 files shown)


it-praktyk:
Guys,
please re-read what you wrote previously and take a decision.

If not I'll close this pull request after 14 days. Is it OK?

nohwnd:
My older self was stupid and assertions that do too much suck. Comparing objects properly is way more difficult that what is implement in this PR so I would close this and rather work on my Assert module to get it object comparison to working state.

it-praktyk:
@dmalikov, thank you for your contribution.

Based on opinion @nohwnd I'm closing your pull request now.

nohwnd:
@dmalikov re-reading my answer I see it might be understood as arrogant, and downputting, but that was not my intention. It was supposed to have a jokeful tone to it. We had few bugs with Should Be doing weird things based on context, that is why I am trying to restrict assertions to do just one thing.  Also I tried implementing object comparison and it has 200+ tests and ton of code and I am still not done with it. :)

PR #201, opened 2014-08-30
Tweak Contain and Contain Exactly

Jaykul:
Fix Should Contain and Should ContainExactly
- You should be able to specify more than one line for the test, as in `| Should Contain "line`nanother"`
- You shouldn't have to think about RegEx in a "contains" test
- You should be able to do: `"Some string" | Should Contain "string"`

After doing that, I also tweaked Match so it could be used on file content if you pass a FileInfo object (not just a path): `get-item TestDrive:\Test.txt | Should match "the contents"`

### DIFF
--- a/Functions/Assertions/Contain.Tests.ps1
+++ b/Functions/Assertions/Contain.Tests.ps1
@@ -3,7 +3,7 @@ Set-StrictMode -Version Latest
 InModuleScope Pester {
     Describe "PesterContain" {
         Context "when testing file contents" {
-            Setup -File "test.txt" "this is line 1`nrush is awesome"
+            Setup -File "test.txt" "this is line 1`nis rush awesome?`none last line"
             It "returns true if the file contains the specified content" {
                 Test-PositiveAssertion (PesterContain "$TestDrive\test.txt" "rush")
             }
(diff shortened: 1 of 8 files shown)


Jaykul:
Hey @nohwnd how much stuff have we broken in Pester4 -- do we have room for a breaking change?
Put another way, can we just merge this or delete it?

it-praktyk:
Is it still relevant or can be closed?

it-praktyk:
> This is technically a breaking change (the [regex]::Escape() part in the Contains operators). 
Is it last call to include the pull request in Pester v4?

nohwnd:
I think we should make this change. I suggest we do it this way:

- replace `Contain*` with `Match*`
- do not re-introduce the Contain assertion to make it break after update 
- provide regex to migrate current scripts

I suggest we do not provide the FileInfo functionality, assertions should be unambiguous, and I totally see someone banging on our door tomorrow that he provided a file but expected the object to be serialized to string and matching on the path.

@dlwyatt  @Jaykul ?

nohwnd:
Looking at how much code would need to be generated to update from `Contains` to just `Match` I think we should just rename it to `FileContentMatch`. That way update will be simpler for those who are affected. And the assertion will finally do what it's name says.

There is also [this Twitter poll](https://twitter.com/PSPester/status/901799465278951424).

Jaykul:
As in `$Path | Should FileContentMatch pattern`?

nohwnd:
@Jaykul yes, in that fashion, but I would probably name it `FileContentMatches`, what do you think?

Jaykul:
I think `FileContentMatch` makes more sense than `FileContentMatches` because in English we would say "it should match" not "it should matches" -- and also because the regex operator in PowerShell is `-match` ...

But I'm not going to worry much about the exact name of the operator... I mean, since Get-Content works on things other than files, maybe it should just be: `ContentMatch` ... but it's just not that bit a deal, even if you want to name it

```posh
$PathVariable | Should BeThePathToAFileWithContentMatching $Pattern
```

😁

nohwnd:
BePathToAFileWithContentLoadedWithUTF8EncodingMatchingExactly  :D 

Okay let's stick with FileContentMatch.

PR #228, opened 2014-11-18
Include stack in test result output

ghostsquad:


alx9r:
@nohwnd @dlwyatt @ghostsquad Can anyone comment on the fate of this PR?  I've got a growing body of integration tests which, when fail, would be debugged faster if a stack trace was shown.

ghostsquad:
@alx9r Wow. It's been awhile. Sorry, I've not been doing much in the way of PowerShell development or testing as of recent, so this has dropped off my radar.

In order for Pester to really move forward, I think newer versions need to drop PowerShell 2.0 support. PowerShell has come such a long way, and it would be a real shame to keep this framework from being really awesome because of the strict need to stay backwards compatible to such an early version of PowerShell.

PR #229, opened 2014-11-18
Include the file the test came from in the Test Output

ghostsquad:


dlwyatt:
This will be addressed in v4

PR #240, opened 2014-12-16
Allows a user to provide templates for New-Fixture

gpduck:
Searches $ENV:UserProfile\Documents\WindowsPowerShell\Pester for NewFixtureTestTemplate.ps1 or NewFixtureFunctionTemplate.ps1 to use in place of the default templates.

### DIFF
--- a/Functions/New-Fixture.Tests.ps1
+++ b/Functions/New-Fixture.Tests.ps1
@@ -96,7 +96,7 @@ Describe "New-Fixture" {
 
             $functionFilePath | Should Exist
 
-            (Get-Content -Path $functionFilePath -Raw) | Should Be ("function #name# {`r`n`r`n}`r`n" -replace "#name#",$name)
+            [System.IO.File]::ReadAllText( (Get-Item $functionFilePath).FullName ) | Should Be ("function #name# {`r`n`r`n}`r`n" -replace "#name#",$name)
         }
 
         It "Creates the default test template" {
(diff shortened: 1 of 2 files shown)


it-praktyk:
I think that generally it's a good idea to have option to use a custome template. What about using additional parameter with an user declarable path to file?

PeterMTaylor:
Got ideas how the template would look so I know this works under New-Fixture and to be compliant? 
My concern is making sure I could also fill in the required help details at the same time.

gpduck:
The idea would be that you could put whatever you wanted in the template and then you could customize the individual test with whatever you needed.

it-praktyk:
the related issue #783

nohwnd:
@gpduck Thank you for working on this, but I won't merge it. It's been sitting here for many years and I think it's time to close it. :) This could still be provided by an external script, but I think it's more suited to be just an interactive snippet, then anyone can customize it to their likings.

randyram-at-ms:
FWIW, my team is constantly asking, "what is $sut?" in code reviews, asking me to change it to be more meaningfully named like $systemUnderTest.  @nohwnd what do you mean by "could still be provided by an external script"?

nohwnd:
@randyram-at-ms I mean that you can write a function that will generate any code that you like, so we don't have to provide it. It is hard to make a templating function that will satisfy everybody, and it does not seem to be needed now that every editor has templating support.

You can name $sut any way you like, you can call it $script or $code or whatever you like I personally don't use new-fixture at all.

Issue #246, opened 2014-12-23
Pester shoud understand arrays


vors:
This test fails on a4244c3c3825f4a108be5283a98bb6eb7a3e3508 (v4 branch)

``` powershell
Describe "Pester shoud understand arrays" {
    It "two arrays are different" {
        @(1) | Should not be @(1,2)
    }
}
```

francoislg:
I hit the same wall with arrays today and I think it's a wonderful idea to support it.

I've come up with a simple solution that works quite well for my context (thought probably not elegant enough) to compare two arrays regardless of their order.

function Assert-ArrayEquals {
	param($array, $expected)
	for($i = 0; $i -lt $expected.length; $i++){
		$expected -contains $array[$i] | Should Be $true
	}
}

The only problem is that it isn't quite verbose enough.

Hope that helps others

it-praktyk:
Please check @stuartleeks projects
- [PesterMatchArray](https://github.com/stuartleeks/PesterMatchArray)
- [PesterMatchHashtable]https://github.com/stuartleeks/PesterMatchHashtable

More Pester's related links on my blog -> https://blog.it-praktyk.pl/2016/05/29/Pester-related-links/

DarkLite1:
Stumbled upon this too and tried to use the suggested solution with `PesterMatchArray`, but it doesn't test the content [deeply enough](https://github.com/stuartleeks/PesterMatchArray/issues/5) . So in the end I reverted to [this post](https://dscottraynsford.wordpress.com/2015/08/23/comparing-objects-using-json-in-powershell-for-pester-tests/) and used the following test:

```
$Result = @(
    [PSCustomObject]@{
        'First Name' = 'Chuck'
        'Address'    = 'California'
    }
    [PSCustomObject]@{
        'First Name' = 'Jean-Claude'
        'Address'    = 'Brussels'
    }
)

$ExpectedResult = @"
[
    {
        "First Name":  "Chuck",
        "Address":  "California"
    },
    {
        "First Name":  "Jean-Claude",
        "Address":  "Brussels"
    }
]
"@

[String]::Compare(($Result | ConvertTo-Json),$ExpectedResult) | Should Be 0
```

nohwnd:
Maybe this could help you: https://github.com/nohwnd/Assert#comparing-whole-objects

DarkLite1:
Perfect! Even better! Thank you very much for the helpful module. Love it already!

it-praktyk:
[...]

PR #252, opened 2015-01-15
Added array comparison support for Be / BeNullOrEmpty / BeExactly

dlwyatt:
This will be to address issue #246 .  It may not be ready to merge as-is yet; I just dusted off the old code I was working on to fix this problem a few months ago, but had thrown out due to difficulties telling the difference between an empty array and a $null value when the pipeline is involved.

This is a first attempt at improving the behavior of Should when multiple objects are piped in.  The Should Be, Should BeExactly, and Should BeNullOrEmpty assertions will now be performed against the entire input set, instead of once for each input object.  Also, arrays will no longer be fed straight to the -eq / -ceq operators (which carries some tricky baggage in PowerShell).  Instead, PowerShell will enumerate the arrays together and make sure their scalar elements are equal.

Other operators still behave as they always have, since passing an array to something like Should Exist doesn't make sense (unless your intention is to make sure that every file name in the array exists, in which case the old behavior is correct.)

I was running into some performance problems with the new code; calling Get-Command multiple times in our Should code seems to slow it down quite a bit, particularly if the command doesn't exist.  I've updated the Should code to have each operator register itself in a hashtable at module load time, instead of relying on a naming convention and Get-Command.  This keeps our performance at a consistent level regardless of how we tweak Should operator features.

There were a few old tests that worked, but only due to some odd buggy behavior (piping objects of an arbitrary non-boolean type to "should be $true"; they started to fail in this update.  I've changed those tests to use "Should Not BeNullOrEmpty" instead.

### DIFF
--- a/Functions/Assertions/Be.Tests.ps1
+++ b/Functions/Assertions/Be.Tests.ps1
@@ -25,6 +25,35 @@ InModuleScope Pester {
 
             $array | Should Be $arrayWithCaps
         }
+
+        It 'Handles reference types properly' {
+            $object1 = New-Object psobject -Property @{ Value = 'Test' }
+            $object2 = New-Object psobject -Property @{ Value = 'Test' }
+
+            $object1 | Should Be $object1
+            $object1 | Should Not Be $object2
+        }
+
+        It 'Handles arrays with nested arrays' {
+            $array1 = @(
+                @(1,2,3,4,5),
+                @(6,7,8,9,0)
+            )
+
+            $array2 = @(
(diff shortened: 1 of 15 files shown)


dlwyatt:
Weird... I rebased this onto master and ran all the tests locally on PSv2 and v5.  Build server has it failing on v3, with an error that for the moment is making me scratch my head.  Will be setting up a VM with PowerShell v3 on it to see if I can reproduce the problem.

dlwyatt:
Merged into DevelopmentV4; will be released in that version.

PR #254, opened 2015-01-15
Allow NuGet package to be build with module in root

gpduck:
PowerShellGet packages are currently built expecting the module to be in the root of the NuGet package, so this change allows the build script to specify a custom base for the target in the NuSpec so it can build either Chocolatey or PowerShellGet style packages.

I've left the default behavior to match the current folder structure, but a PowerShellGet module can be built using:

`Invoke-PSake .\build.psake.ps1 -properties @{targetBase = "" }`

### DIFF
--- a/Pester.nuspec
+++ b/Pester.nuspec
@@ -13,15 +13,14 @@
     <licenseUrl>http://www.apache.org/licenses/LICENSE-2.0.html</licenseUrl>
   </metadata>
   <files>
-    <file src="Pester.psm1" target="tools" />
-    <file src="Pester.psd1" target="tools" />
-    <file src="Pester.Tests.ps1" target="tools" />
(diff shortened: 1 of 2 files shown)


dlwyatt:
What does this gain us?  PowerShellGet takes care of its own NuGet packaging behind the scenes when you run Publish-Module; I'm not aware of a scenario where we have to do that ourselves.

gpduck:
Flexibility?  I'm actually trying to setup my own builds internally at work
but I can't use the project's psake script because I need my output in the
other format.  I was trying to use PowerShellGet as a reason that it would
apply to people outside my company.

nohwnd:
@gpduck @dlwyatt currently we push to NuGet, Chocolatey and PowerShellGallery and as far as I know all the packages work just fine. But if this works without changing the configuration of the build server I see no problem merging it.

PR #263, opened 2015-01-23
Missing snippets directory causes Import to fail

ToddBowles:
When attempting to Import the module when running a script through Powershell ISE, a missing snippets directory will cause the import to fail.

### DIFF
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -260,9 +260,10 @@ function Get-ScriptBlockScope
     [scriptblock].GetProperty('SessionStateInternal', $flags).GetValue($ScriptBlock, $null)
 }
 
-if (($null -ne $psISE) -and ($PSVersionTable.PSVersion.Major -ge 3))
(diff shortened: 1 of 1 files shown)


nohwnd:
@ToddBowles  @dlwyatt I don't get the fix. That directory is part of the module distribution, why it should suddenly became missing? And if it does, why the import should not fail?

todd-bowles-onthehouse:
That's a good point. I wrote the fix because I thought the lack of the snippets distribution might have been intended (in which case the directory was optional and loading should just be skipped if it didn't already exist).

I suppose I expected either the root issue to be fixed (missing snippets) or the pull request to be merged.

Not both.

I can see your confusion.

I suggest reverting the pull request in that case, if the snippets are a non-optional part of the distribution.

nohwnd:
@todd-bowles-onthehouse Okay we agreed that the fix stays.

Issue #271, opened 2015-01-27
Allow user to specify arguments to be passed to test files


dlwyatt:
Example usage:

```posh
$tests = @(
    @{ Path = "$root\test1.Tests.ps1"; Parameters = @{ Parameter1 = 'Value1'; Parameter2 = 'Value2' } }
    "$root\test2.Tests.ps1"
)
Invoke-Pester -Path $tests
```

vors:
I have another usage proposal:
```powershell
Invoke-Pester -Path .\foo.Tests.ps1 -Parameters = @{ Parameter1 = 'Value1'; Parameter2 = 'Value2' } 
```
If somebody want to run more then 1 test file, he can call `Invoke-Pester` twice and merge results.

vors:
I tried to merge results, and I cannot find an API to do it. Does it exist?

dlwyatt:
Nope, not at the moment.

ghostsquad:
I don't think this is a great idea. I also think it's the author's responsibility to encourage proper testing patterns.

http://googletesting.blogspot.com/2014/07/testing-on-toilet-dont-put-logic-in.html

> Unlike production code, simplicity is more important than flexibility in tests. Most unit tests verify that a single, known input produces a single, known output. Tests can avoid complexity by stating their inputs and outputs directly rather than computing them. Otherwise it's easy for tests to develop their own bugs.

My suggestion would be to setup Data-Driven/Parameterized tests like what is described in #179

dlwyatt:
The author knows jack squat about proper testing patterns.  ;)  I just make the thing work.  This particular feature request came from VMware.

vors:
Request also came from Microsoft, PowerShell team. We want run same test suite against different VMs in parallel (different VMs for different OS versions). Currently we workaround it with passing global variable, which is ugly.

ghostsquad:
@vors - That's an interesting use case. Though I would argue that a test config file would be better, and there would be no need to try to support the originally requested feature. Regarding your comment on global variables being ugly: yep, they certainly are, if the global scope is cluttered with variables. That's generally bad practice. There's nothing _truely_ wrong with a global variable, as long as they are used properly. A `$global:Variable` is essentially equivalent to a Static Property in C#. There's quite a few static/global classes used in C# all the time, namely ConfigurationManager comes to mind, as it does exactly what we are talking about here.

```Xml
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
<startup> 
[...]

Issue #270, opened 2015-01-27
Allow execution of test files that don't end with .Tests.ps1


dlwyatt:
If a full path to a file name is passed to the -Path argument of Invoke-Pester, just treat that as a test file and run it.  The *.Tests.ps1 filter should be applied if -Path is given a folder name rather than a file.

PR #272, opened 2015-01-27
Path overhaul

dlwyatt:
Proposed fix for issues #270 and #271 .

As a side effect of these changes, we lose a little bit of console output ("Executing all tests in $Path...") and the Path property is gone from the PesterState object.  We intend to add path information about each test script that is executed into the PesterState object in v4 (as a test suite, similar to Describe / Context), which is a better place for that anyway.


Pester.psm1
```diff
@@ -221,6 +230,85 @@ about_pester
     if ($EnableExit) { Exit-WithCode -FailedCount $pester.FailedCount }
 }
 
+function ResolveTestScripts
+{
+    param ([object[]] $Path)
```
vors: Should we rename `$Path` to something different to represent that it may contain both path and arguments? Maybe  create a separate Parameter set for it

dlwyatt:
What do you mean by "add line new line"?

vors:
I'm talking about this:
image

[Why it's impotent for git](http://stackoverflow.com/questions/5813311/no-newline-at-end-of-file).

dlwyatt:
Ah, gotcha.  I can check that as well.  Just added some tests to make sure there's no trailing whitespace in any file... build blew up, as you might imagine.  :P

vors:
Yeah, we just need to fix all triling spaces, line ending once and never have this problem again :)

dlwyatt:
Will do that later tonight as part of this PR.  It'll mean there's a mountain of changes in the diff, but oh well.

vors:
About teamcity github hook again:
image

Gray cross usually indicates "aborted". For "failed" common practice is to use red cross. (see https://github.com/Azure/azure-powershell/pull/102 for example). I don't know how to express it on github hook level, so I just talking in terms of user interface. :)

It will be beneficial to align reporting mechanism with common practices, so new contributors will have a consistent experience.

dlwyatt:
I might be able to change that on the TeamCity server; will see what I can do.

dlwyatt:
Hmm, nevermind.  I thought maybe I had set it up to "cancel build" if the dependencies failed, but it's actually set to "run build, but log problem."  Beyond that, I don't seem to have any control over the GitHub hook.

It's interesting that the X is red down by the Merge pull request button, but is gray on the line with the commit itself.

dlwyatt:
Trying one thing... instead of reporting on build start / finish, I switched it to only report on finish.  Right now the commits have both a Pending and a Failure status, and maybe that's somehow messing with the icons.

vors:
Reporting only on finish will probably mean, that it will not report 'in progress status'.

dlwyatt:
That's probably fine.  It's only a couple minutes from start to finish in our build anyway.

Pester.psm1
dlwyatt: I'm okay with that, but we'd need to keep a Path alias to maintain backward compatibility.  Maybe something like `-Script` or `-TestScript`?

vors:
Maybe preserving `-Path` as `string[]` and adding `-Target` as a `hashtable[]`.
It will be much harder to rename it, once people take dependencies.

Also, we definitely need an example in `Invoke-Pester` function definition.

dlwyatt:
I don't think I like the idea of having separate parameters.  It's not consistent with the -CodeCoverage parameter (which already accepts strings or hashtables), as well as what people are already accustomed to in Select-Object and Format-*.

vors:
Ok. Please, update Pester.psm1 help.

dlwyatt:
Done, but it's kind of wordy for a Parameter entry.  By contrast, here's what Select-Object's help has to say about the Property parameter:

`Specifies the properties to select. Wildcards are permitted.`

Not a single mention of calculated properties or hashtables until you get down to the examples.  Would that be better?

vors:
LGTM

PR #284, opened 2015-02-19
resolves #282

### LINKED ISSUE #282: Throw Message Assertion is not working in all cases
Repro script
```Powershell
Describe 'Test' {
    It 'Test a Throw 1' {
        $o = new-object psobject
        $o | Add-Member -MemberType ScriptMethod -Name 'Test' -Value {
            throw (new-object System.Exception('my message'))
        }

        { $o.Test() } | Should Throw 'Assert failed on "Test" with "0" argument(s): "my message"'
    }

    It 'Test a Throw 2' {
        $o = new-object psobject
        $o | Add-Member -MemberType ScriptMethod -Name 'Test' -Value {
            throw (new-object System.Exception('my message'))
        }

        { $o.Test() } | Should Throw 'Exception calling "Test" with "0" argument(s): "my message"'
    }
}
```

Output (slightly altered to get expected vs actual messages to line up)
```
Describing Test
 [-] Test a Throw 1 36ms
   Expected: the expression to throw an exception with message 
{Assert failed on "Test" with "0" argument(s): "my message"}, an exception was raised, message was 
{Assert failed on "Test" with "0" argument(s): "my message"}
       from C:\temp\pesterthrow.tests.ps1:8 char:11
       +         { $o.Test() } | Should Throw 'Assert failed on "Test" with "0" argument( ...
       +           ~~~~~~~~~
   at line: 8 in C:\temp\pesterthrow.tests.ps1
 [+] Test a Throw 2 13ms
Tests completed in 50ms
Passed: 1 Failed: 1 Skipped: 0 Pending: 0
```

Notice how the _actual_ exception message is actually different that that which Pester is showing us? If Pester were showing us the *real* exception message, the first test would pass, and the second test would fail. Clearly those 2 strings in the first test are identical.

ghostsquad:
I still think there should be a test for this change, I just don't know exactly how to test pester from pester. :(

### DIFF
--- a/Functions/It.ps1
+++ b/Functions/It.ps1
@@ -122,7 +122,7 @@ function ItImpl
 
         [Parameter(ParameterSetName = 'Skip')]
         [Switch] $Skip,
-        
+
         $Pester,
         [scriptblock] $OutputScriptBlock
     )
@@ -134,7 +134,7 @@ function ItImpl
     {
         throw 'No test script block is provided. (Have you put the open curly brace on the next line?)'
     }
-    
+
     #the function is called with Pending or Skipped set the script block if needed
     if ($null -eq $test) { $test = {} }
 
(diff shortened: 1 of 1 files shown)


dlwyatt:
"I just don't know exactly how to test pester from pester."

That is sometimes tricky.  I'll add a test after merging this.

PR #288, opened 2015-03-05
Adds a way to set the xml output from the batch

cosmo0:


cosmo0:
Welp, just found out that you can pass any argument. Sorry !

Issue #321, opened 2015-04-24
Pester should expose what test is currently running, today it doesn't.


theficus:
Pester doesn't tell you what test is currently running. This means that when you're watching the output results all you know is what previously ran and what its result was, but there's no visibility into the test that's currently running.

I have some tests that take 30+ seconds to run and it would be nice to know what's currently executing.

Some ideas for how to solve this:
* Use Write-Progress to show what's currently running (I like this one)
* Use Write-Verbose to log an additional line like ```[*] MyTestName starting ...```. The nice thing about this approach is you can suppress it by disabling verbose logging.
* Use Write-Output to log an additional line

dlwyatt:
We could add some Write-Progress calls, though they'd go from 0% to Completed all at once.  No way for us to predict how long a particular test is going to take, but at least you'd have an indication of what's running.

theficus:
Yeah I think that would be fine though. A nice indicator as to what it is running.

--
Sent from my phone. Expect brevity and typos.

On Apr 24, 2015, at 8:35 PM, Dave Wyatt <notifications@github.com<mailto:notifications@github.com>> wrote:


We could add some Write-Progress calls, though they'd go from 0% to Completed all at once. No way for us to predict how long a particular test is going to take, but at least you'd have an indication of what's running.

-

nohwnd:
In console rewriting the current line of output is pretty easy, and it actually looks nice. 
```powershell
Write-Host "[*] test running..." -ForegroundColor Yellow -NoNewline
Start-Sleep -Seconds 1
Write-Host "`r[+] test succeeded " -ForegroundColor Green
```
Problem is that it does not work in ISE. 

Writing out progress, suggested by Dave, works in both console and ISE (and likely other hosts), and looks quite good in my opinion: 
```powershell 
1..10 | foreach {
    $r = (Get-Random -Minimum 10 -Maximum 1000)
    Write-Progress -Activity testing -PercentComplete 0 -CurrentOperation $_
    Start-Sleep -Milliseconds $r
    Write-Progress -Activity testing -PercentComplete 100 -CurrentOperation $_
    
    if ($r % 2)
    {
        Write-Host "[+] Test completed" -ForegroundColor Green
    }
    else 
    {
        Write-Host "[+] Test failed" -ForegroundColor Red        
    }
    Start-Sleep -Milliseconds 10 #sleep for few milliseconds to see the progress go to 100 %
}
```
[...]

PR #322, opened 2015-04-25
Adding some Write-Progress output to It

dlwyatt:
Potential fix for #321 

I added a few blank lines to the console output at the beginning of `Invoke-Pester`, to make room for the Progress display.  Tests that finish quickly may not wind up displaying any progress output at all, but long-running tests will provide the user with some feedback about what's happening.

### DIFF
--- a/Functions/It.ps1
+++ b/Functions/It.ps1
@@ -230,6 +230,8 @@ function Invoke-Test
     }
     else
     {
+        Write-Progress -Activity "Running test '$Name'" -Status Processing
+
         $errorRecord = $null
         try
         {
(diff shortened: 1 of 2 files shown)


alx9r:
This change seems to be causing some pretty significant side-effects -- at least on my system.  Here are the downsides:

* Much of the run-time the result of the latest few test runs are out of site below the bottom of the ISE console.
* The entire console contents jerks up and down as each test runs.  The jerking is harsh enough that I have to avert my eyes to avoid discomfort.

dlwyatt:
I'm not sure what you're referring to.  I don't notice any jerking up and down in the ISE or powershell.exe.  In both cases, the progress indicator appears at the top of the console / pane, while the most recent test results are at the bottom.

In any case, if you don't want to see that stuff, you can set $ProgressPreference to 'SilentlyContinue' before running Invoke-Pester.

alx9r:
Hmm...I just restarted ISE and the symptoms aren't there.  I think it might have something to do with multiple powershell sessions in ISE (i.e. using `CTRL-T`) or ISESteroids, or both.  I noticed it throughout multiple restarts today, but I didn't realize it wasn't constant until now.

Issue #315, opened 2015-04-20
Is there a terse way to assert that a mock was called with certain parameters and not otherwise?


alx9r:
Consider the following files which demonstrate asserting that Write-Error was called only once with a particular parameter and _not otherwise_.  I find this to be a common scenario for testing basic validation logic and error reporting of functions.

## Module.Tests.ps1
```Powershell
Import-Module Module

InModuleScope Module{
Describe f1 {
    Mock Write-Error -Verifiable 

    It 'writes exactly "message 1"' {
        f1 one | Should be $false
        Assert-MockCalled Write-Error -Exactly -Times 1 -Scope It -ParameterFilter {
            $Message -eq "message 1"
        }
        Assert-MockCalled Write-Error -Exactly -Times 0 -Scope It -ParameterFilter {
            $Message -ne "message 1"
        }
    }
    It 'writes exactly "message 2"' {
        f1 two | Should be $false
        Assert-MockCalled Write-Error -Exactly -Times 1 -Scope It -ParameterFilter {
            $Message -eq "message 2"
        }
        Assert-MockCalled Write-Error -Exactly -Times 0 -Scope It -ParameterFilter {
            $Message -ne "message 2"
        }
    }
}
}
```
## Module.psm1

This is the module we are working on.  It has mistakes and the tests aren't passing yet.

```Powershell
function f1
{
    param($which)

    if ( $which -eq 'two' )
    {
        Write-Error 'message 1'
    }
    if ( $which -eq 'two' )
    {
        Write-Error 'message 2'
        return $false
    }
    return $true
}
```

## Is there a terser way to achieve the same mock-called assertions?

What I don't like about how I am doing this is the repetition in the two calls to `Assert-MockCalled` in each `It` block.  The first call always has `-Exactly -Times 1` parameters and some filter, while the second call always has `-Exactly -Times 0` and the inverse of the first calls' filter.  I have found that maintaining this duplication is rather time-consuming and error-prone.

Normally I would deal with this kind of duplication with a helper function that makes both calls.  I have experimented with this a bit but haven't found an improved result.  I ran into two problems:

* The Pester error messages (e.g. `Expected Write-Error in module Module to be called...`) became even more confusing to interpret.
* I did not find a universal way to transform the `ParameterFilter` scriptblock to one that evaluates to its boolean inversion.

I find myself wishing that `Assert-MockCalled` had a `-NotOtherwise` switch that could be used with `-ParameterFilter`.

[...]

Issue #269, opened 2015-01-26
Improve error output from Assert-MockCalled


dlwyatt:
Right now this is reporting a line number in Mock.ps1.  We should be able to make this throw an assertion like the ones from `Should`, which reports the calling line from the test script instead.  Sample of current output:

```
 [-] Does something 55ms
   Expected Set-Content to be called at least 1 times but was called 0 times
   at line: 518 in C:\Program Files\WindowsPowerShell\Modules\Pester\Functions\Mock.ps1
```

alx9r:
Implementing this would mean a significant improvement in productivity for me.  As it stands, I spend a significant amount of time stepping through code just so I can make a sensible guess at (and it usually does involve guessing) which call to Assert-MockCalled failed and for what reason.  

This is probably my biggest wish for Pester because of the amount of time it has the potential to save.

If this is not likely to be implemented soon, do you have some advice on how to converge on the particular failing Assert-MockCalled quicker?

dlwyatt:
Thanks for reminding me; I'd forgotten about this one.  I'm at the PowerShell Summit at the moment, but I'll get this done some time next week.

alx9r:
Oh right.  The summit.  Thanks and have a good time!

PR #324, opened 2015-04-25
Fixes for #315 and #269

dlwyatt:
The new parameter for Assert-MockCalled is called `-NotOtherwise` for now, but I think we may want to come up with a better name.  Either way, the functionality is implemented now.

Assert-MockCalled now throws the same type of ErrorRecord objects as the Should command, which gets us better output as noted in #269.


dlwyatt:
I've ditched the -NotOtherwise switch in favor of a different parameter set (-ParameterFilter and -ExclusiveFilter; both take script block filters, and -ExclusiveFilter means that all calls to the mock must match the filter.)

Issue #212, opened 2014-09-18
Should generate coverage report files 


Jaykul:
It would be really great if we could export code coverage reports to a reusable format to show progress, and document state.

There are several known formats already (e.g. gcov, lcov, even a json format) -- which are used by projects like codecov.io to generate pretty web-based code-coverage reports, can we pick one of those?

See here, for docs and examples: https://codecov.io/docs

DarqueWarrior:
Has there been anymore thought on this? I would really like to publish my code coverage results as part of my CI. What would it take to generate a JaCoCo or Cobertura?

dlwyatt:
I'll have to look into those formats and see what they support.  It's not difficult to take the information we have in memory and dump it out to a file, but Pester's code coverage analysis looks at PowerShell _commands_, regardless of whether they're one to a line, one command spanning multiple lines, or multiple commands on a single line.  Where it gets awkward is translating that to a format that expects a report on what _lines_ were hit.

stevepeak:
@DarqueWarrior Can you upload a Jacoco report to me. I'm curious if if contains complexity data because Codecov will support Jacoco Complexity next month.

> Both Jacoco and Cobertura are acceptable upload formats to Codecov.

DarqueWarrior:
@stevepeak Here you go. These are from a very simply Java project created with [yo vsts](https://github.com/DarqueWarrior/generator-vsts).
[CoberturaCodeCoverageReport.zip](https://github.com/pester/Pester/files/615111/CoberturaCodeCoverageReport.zip)
[JaCoCoCodeCoverageReport.zip](https://github.com/pester/Pester/files/615112/JaCoCoCodeCoverageReport.zip)

dlwyatt:
Both of those formats have assumptions from Java (things like "Packages" and "Classes" must exist, etc), but we could probably still just fudge that and put Pester's useful information in there.

stevepeak:
@DarqueWarrior very nice 👍 both those formats work.

DarqueWarrior:
I have used them for C# which does not have Packages but still was useful.

TravisEz13:
The DSC resource has a Module we use to generate reports for `codecov.io` here
https://github.com/PowerShell/DscResource.Tests/tree/dev/DscResource.CodeCoverage

RangerDjanger:
Has anyone been able to generate code coverage results to publishing into TFS/VSTS?

it-praktyk:
[...]

PR #341, opened 2015-05-05
Add hit commands property to CodeCoverage object

TravisEz13:
Add hit commands property to CodeCoverage object, by exposing the data pester already has.

I believe this helps address #212 by allowing the creation of adapter/translators that can create code coverage reports rather than making pester proper responsible for it.

This would not prevent pester for generating reports itself in the future, but different users may have different needs and I cannot find a clear solution to which report is best. This method would allow for easier experimentation.


vors:
:+1:

Jaykul:
Beauty
https://codecov.io/github/PoshCode/PSGit/src/PSGit.psm1?ref=dev

PR #348, opened 2015-05-12
Fix Should Exist assertion to handle paths with [ ]

### LINKED ISSUE #343: Special characters do not work with Should Exist
The first It works but the second does not. There is a need of ExistExactly.

```powershell
Describe "Testing Exists" {
    It "works with test-path" {
        $file = New-Item -Path "TestDrive:\[test].txt" -ItemType File         
        Test-Path -LiteralPath $file | Should Be $true
    }

    It "should work with Exists" {
        $file = New-Item -Path "TestDrive:\[test].txt" -ItemType File 
        $file | Should Exist        
    }
}
```

nohwnd:
Fixing bug which caused that paths that contain [ ] wildcard were not
handled correctly by the Should Exist assertion.

It's morning am not 100 % sure that this won't break anything. But my reasoning is that any path that is provided is expanded by PowerShell (like path with environment variables) so no special handling is needed there. Same goes for PS(Provider)Paths... nothing else comes to mind. 

Review please. :)

Fixes #343

### DIFF
--- a/Functions/Assertions/Exist.ps1
+++ b/Functions/Assertions/Exist.ps1
@@ -1,6 +1,7 @@
 
 function PesterExist($value) {
-    return (Test-Path $value)
+    $resolvedPath = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($value)
+    Test-Path -LiteralPath $resolvedPath
 }
 
 function PesterExistFailureMessage($value) {
@@ -10,5 +11,3 @@ function PesterExistFailureMessage($value) {
 function NotPesterExistFailureMessage($value) {
     return "Expected: ${value} to not exist, but it was found"
 }
-
-
--- a/Functions/Assertions/Exist.ps1
+++ b/Functions/Assertions/Exist.ps1
@@ -1,7 +1,6 @@
 
 function PesterExist($value) {
-    $resolvedPath = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($value)
-    Test-Path -LiteralPath $resolvedPath
+    Test-Path $value
 }
 
 function PesterExistFailureMessage($value) {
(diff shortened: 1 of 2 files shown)


nohwnd:
So obviously I was thinking about files and registry only and forgot about function and env drives. Looking into possible fixes.

nohwnd:
@dlwyatt Any fix I can come up with, pollutes the assertion just to solve this edge case that did not surface for years. I suggest to document this problem and suggest to escape the path with two back-ticks like this: ````"TestDrive:\``[test``].txt"````. I will also add tests for the function and env drives. What do you think? Any better ideas?

Functions/Assertions/Exist.ps1
```diff
@@ -1,6 +1,7 @@
 
 function PesterExist($value) {
-    return (Test-Path $value)
+    $resolvedPath = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($value)
```
dlwyatt: This line is going to break things (because it strips off the provider name from PS paths), and isn't necessary anyway.  Just the call to `Test-Path -LiteralPath` should be sufficient, I think.

nohwnd: Yup, I realized that as well after the test failed on CI. How about just documenting it, adding more tests and keeping the assertion as it? https://github.com/pester/Pester/pull/348#issuecomment-101235064

dlwyatt:
Honestly, I think the value of `Should Exist` is way down now that our failed assertions show the content of the line in the test code.  `Test-Path -LiteralPath $something | Should Be $true` is now a perfectly acceptable test, where before I would have preferred `Should Exist`.  So I wouldn't bother with any convoluted escaping of paths in the workaround or documentation.  :)

nohwnd:
Good point. I will document that, including your "work around" and close the issue.

PR #365, opened 2015-06-07
New assertion syntax

dlwyatt:
Should operators can now be accessed by dynamic parameter sets, instead of only the old `Should Be Value` syntax, which couldn't be easily expanded.  Now we can add things like `Should -Throw -ExceptionType ([Exception])`, etc.  Should operators can also be given aliases in the new parameter sets, such as `Should -Be` which can also be written as `Should -EQ`.


nohwnd:
Awesome!

Issue #363, opened 2015-05-29
Should `It [-Skip]` also have an alias [-ignore] assigned to the property?


ChaseFlorell:
Just working my way through pester, and I've got a test that I want to ignore. In nUnit and xUnit, it's possible to have an [ignore] attribute.

I noticed that in pester it's called `-skip`, and I'm wondering if we should add an alias `-ignore`?

    [CmdletBinding(DefaultParameterSetName = 'Normal')]
    param(
        [Parameter(Mandatory = $true, Position = 0)]
        [string]$name,

        [Parameter(Position = 1)]
        [ScriptBlock] $test = {},

        [System.Collections.IDictionary[]] $TestCases,

        [Parameter(ParameterSetName = 'Pending')]
        [Switch] $Pending,

        [Parameter(ParameterSetName = 'Skip')]
        [alias('Ignore')]
        [Switch] $Skip
    )

ChaseFlorell:
Along this same line, when the xml output is generated, nunit will use `result="Ignored"` whereby pester uses `result="Skipped"`. When  this happens, our build runner (TeamCity) reports the `-skipped` tests as "Success" instead of "Ignored".

I think the output also needs to be changed to match nunit, if the nunit style output is to be used for the xml results.

ChaseFlorell:
@dlwyatt any chance of having a discussion around this? I'd be happy to fork and implement, but I'd like to discuss it first.

Is there a specific reason for using `skip` instead of `ignore`?
I can definitely implement it as an alias so that it doesn't break existing functionality, however changing the nunit xml would obviously change things, I'm pretty sure (not 100% mind you) that nunit doesn't support the `skip` verb in xml reports.

nohwnd:
@ChaseFlorell Looks like a good idea. Changing the xml output should not be a problem if it was broken till now. Do they show the correct format of the output somewhere in the documentation? Are you sure it's  not fixing the problems "just" for TC?

ChaseFlorell:
Hey @nohwnd, I can't say with 100% certainty that it's not just a TeamCity fix, however, when I look at the results of an nunit run, I've got results that look like this.



    <test-results name="C:\Dev\ECIM\build-artifacts\output\TransCanada.ECIM.Test.dll" total="977" errors="0" failures="0" not-run="1" inconclusive="0" ignored="1" skipped="0" invalid="0" date="2015-06-12" time="09:07:22">
[...]

PR #368, opened 2015-06-12
Feature/issue 363 - Fixes nunit reports generating the incorrect data and also enables an -ignore flag.

ChaseFlorell:
This pull request addresses Issue #363. I've left the "skip" syntax in place, but have added the alias "ignore" for those who are more familiar with nunit.
I've also modified the nunit reports being generated to not only change from skipped to ignored, but also to fix the run count, removing the number of skipped/ignored count from the number of total tests run.

Please see the screenshot to compare my libraries master branch with that of my pester-tests branch. As you can see, the total number of tests represented is still the same, but it now reports the number of skipped/ignored tests as well.

Also, [here is a discussion](https://groups.google.com/forum/#!topic/nunit-discuss/GH5Zp_WWOdw) on the topic of ignore vs skip.

### DIFF
--- a/Functions/It.Tests.ps1
+++ b/Functions/It.Tests.ps1
@@ -51,6 +51,11 @@ InModuleScope Pester {
             $scriptBlock | Should Not Throw
         }
 
+        It 'Does not throw an error if the -Ignore switch is used, and no script block is passed' {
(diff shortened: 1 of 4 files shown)


dlwyatt:
Looks good to me.  We might want to change some of our internal property names (`SkippedCount` to `IgnoredCount`, etc), but otherwise this is fine.

PR #378, opened 2015-06-29
Fix Unicode issue in `Contain` and `ContainExactly`.

### LINKED ISSUE #377: Contain doesn't support Unicode
This test fails:

    Set-StrictMode -Version Latest

    InModuleScope Pester {
        Describe "PesterContain" {
            Context "when testing file contents" {
                Setup -File "test.txt" "this is line 1`nrush is awesome`nAnd this is Unicode: ☺"

                It "returns true if the file contains the specified UTF8 content" {
                    Test-PositiveAssertion (PesterContain "$TestDrive\test.txt" "☺")
                }
            }
        }
    }

Likely other assertions have this problem too.  I see two ways to fix this:

1.  Assume files to be checked is in UTF8.
2.  Provide `-Encoding` option for `Contain` assertion and for other assertions if needed.

I like first option because it works for almost everyone, it is simple and intuitive and it has backward compatibility with ASCII. Also, it is trivial to implement if I don't miss some hidden difficulties.

Second option is more flexible but it is a question if somebody will ever need this flexibility.

tkirill:
fix #377

These are the only places where `Get-Content` is used in assertions.

### DIFF
--- a/Functions/Assertions/Contain.Tests.ps1
+++ b/Functions/Assertions/Contain.Tests.ps1
@@ -1,18 +1,24 @@
-Set-StrictMode -Version Latest
+﻿Set-StrictMode -Version Latest
 
 InModuleScope Pester {
     Describe "PesterContain" {
         Context "when testing file contents" {
-            Setup -File "test.txt" "this is line 1`nrush is awesome"
+            Setup -File "test.txt" "this is line 1`nrush is awesome`nAnd this is Unicode: ☺"
+
             It "returns true if the file contains the specified content" {
                 Test-PositiveAssertion (PesterContain "$TestDrive\test.txt" "rush")
             }
+
             It "returns true if the file contains the specified content with different case" {
                 Test-PositiveAssertion (PesterContain "$TestDrive\test.txt" "RUSH")
             }
 
             It "returns false if the file does not contain the specified content" {
                 Test-NegativeAssertion (PesterContain "$TestDrive\test.txt" "slime")
+            }
+
+            It "returns true if the file contains the specified UTF8 content" {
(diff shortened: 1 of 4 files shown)


tkirill:
How I can fix this error?  Locally I have all tests passed on powershell v4.

dlwyatt:
Those timeouts indicate a problem on the build server, don't worry about it for now.

As for the PR, I'd be concerned with what happens if the code encounters a file that isn't ASCII or UTF-8.  (Unicode / UTF-16, for instance, which are created by default when you use Out-File or the redirection operators.)

tkirill:
@dlwyatt I took the character `レ` which takes 3 bytes `0xE3 0x83 0xAC` in UTF-8.  I saved the character in a file in UTF-16 with BOM and this test passed:

    Describe "Get-Content" {
        It "magically works with UTF-16" {
            'test.txt' | Should Contain 'レ'

            $utf8 = Get-Content -Encoding UTF8 'test.txt'
            $utf16 = Get-Content -Encoding Unicode 'test.txt'
            $oem = Get-Content -Encoding Oem 'test.txt'

            $utf8 | Should Be 'レ'
            $utf16 | Should Be 'レ'
            $oem | Should Be 'レ'
        }
    }

The test passes even without my changes but I don't know why.  Maybe when Powershell sees BOM it assumes UTF-16 and doesn't look at `-Encoding` parameter?  This is strange.

But if save the character in UTF-8 then test fails without this PR and passes with this PR.

tkirill:
@dlwyatt If you don't like these changes I would be happy if you provide small instruction how to implement custom `should` assertions.  It would solve my little problem too :+1:

dlwyatt:
Sorry about that, I lost track of this one.  I'll look it over now and merge if everything seems okay

Issue #47, opened 2013-03-30
Test Time Elapsed should Measure Entire Describe


mwrock:
Currently the time measurements are around the It blocks. This is misleading since these are typically very simple comparisons of basic values and the bulk of the time is in the context or describe blocks. I suggest we measure the describes and print the times after each of those.

PR #380, opened 2015-07-03
Two simple fixes

Remco-74:
The calls to Write-Progress are now simplified, and the progress bar now displays incremental progress.
Timing measurements are now more accurate for the first tests in a context as well.

### DIFF
--- a/Functions/It.ps1
+++ b/Functions/It.ps1
@@ -241,8 +241,6 @@ function Invoke-Test
     }
     else
     {
-        Write-Progress -Activity "Running test '$Name'" -Status Processing
-
         $errorRecord = $null
         try
         {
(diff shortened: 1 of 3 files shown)


Remco-74:
Added a small fix, retry

dlwyatt:
The problem with this timer change is that you're basically reverting the fix for #47 .  Now the time shown for the `It` blocks is accurate, but the total time taken is not.  (For example, running the entire Pester test suite on my computer claims that the tests completed in 13.49s, using this PR's version, but actually took 22.24s).

Fixing this in the console output is easy, but we also need to make sure that the proper times are reflected in the -PassThru and NUnitXml output.  For now, we just accepted that the first test in each Context / Describe would have slightly skewed times.

Issue #373, opened 2015-06-19
Improvements to the progress reporting


theficus:
The addition of adding a progress indicator when running a unit test is great. I have some suggestions to make it even better:

* Remove the "insert blank lines" code. This is really an implementation detail of the PSHost and just looks silly most of the time. I don't think I've seen anything else that uses the progress reporting that does this.

* Show a progress indicator when running BeforeEach/AfterEach/BeforeAll/AfterAll. Ideally when BeforeEach/AfterEach is run, it shows the test name. BeforeAll/AfterAll could show the describe name.

Remco-74:
It would be very nice to update the progressbar incrementally (moving from left to right) as well.

theficus:
Unfortunately PowerShell doesn't support indeterminate progress bars so there's really no good way to get it to move. Either way it's an improvement to know what's running as its running. It can just be made even better. :)

Remco-74:
This can be fixed with a single line of code. I have written a patch for this, but i'm not sure to how to submit this. What is the best way to do that?

Remco-74:
I've made a pull request to improve the progress reporting, but it fails during build (since i didn't add any new tests).

http://teamcity.codebetter.com/viewLog.html?buildId=194548&buildTypeId=Pester_TestPullRequests&guest=1

I'm not quite sure how to write a test that invokes [Invoke-Pester] itself. I didn't find other testcases that do that as well. 

How to proceed from here?

dlwyatt:
The problem isn't that you didn't add any new tests.  The build log contains this error:

```
[Step 1/1] Z:\Builds\temp\buildTmp\powershell6628862289246688128.ps1 : The variable '$cnt'
[06:25:57][Step 1/1]  cannot be retrieved because it has not been set.
[06:25:57][Step 1/1]     + CategoryInfo          : InvalidOperation: (cnt:Token) [Write-Error], Run 
[06:25:57][Step 1/1]    timeException
[06:25:57][Step 1/1]     + FullyQualifiedErrorId : VariableIsUndefined,powershell662886228924668812 
[06:25:57][Step 1/1]    8.ps1
[06:25:57][Step 1/1]  
[06:25:57][Step 1/1] Process exited with code 1
[06:25:59][Step 1/1] ##teamcity[buildStatisticValue key='buildStageDuration:buildStepRUNNER_1278' value='2169.0']
[06:25:59][Step 1/1] Step PowerShell v2 Tests (Powershell) failed
```

[...]

PR #382, opened 2015-07-04
Fix for Progress Reporting - Issue #373

Remco-74:
Pester now uses an incrementing progress bar, it wil increment for each script file.

### DIFF
--- a/Functions/It.ps1
+++ b/Functions/It.ps1
@@ -241,8 +241,6 @@ function Invoke-Test
     }
     else
     {
-        Write-Progress -Activity "Running test '$Name'" -Status Processing
-
         $errorRecord = $null
         try
         {
@@ -277,7 +275,6 @@ function Invoke-Test
         $result = Get-PesterResult -ErrorRecord $errorRecord
         $orderedParameters = Get-OrderedParameterDictionary -ScriptBlock $ScriptBlock -Dictionary $Parameters
         $Pester.AddTestResult( $result.name, $result.Result, $null, $result.FailureMessage, $result.StackTrace, $ParameterizedSuiteName, $orderedParameters )
-        Write-Progress -Activity "Running test '$Name'" -Completed -Status Processing
     }
 
     if ($null -ne $OutputScriptBlock)
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -185,8 +185,6 @@ about_pester
     $pester = New-PesterState -TestNameFilter $TestName -TagFilter ($Tag -split "\s") -ExcludeTagFilter ($ExcludeTag -split "\s") -SessionState $PSCmdlet.SessionState -Strict:$Strict -Quiet:$Quiet
     Enter-CoverageAnalysis -CodeCoverage $CodeCoverage -PesterState $pester
 
-    Write-Screen "`r`n`r`n`r`n`r`n"
-
     $invokeTestScript = {
         param (
             [Parameter(Position = 0)]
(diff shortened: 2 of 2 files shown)


dlwyatt:
Why did you remove the per-test calls to Write-Progress in It.ps1?  That's where the most value is (because there's nothing happening at the console while an It block is running.)

I'm okay with adding additional Write-Progress calls for the test files (as well as Describe / Context), if people find that valuable, but you shouldn't be removing the code that are already there.

Also, this line was there for a reason:  `Write-Screen "`r`n`r`n`r`n`r`n" `.  It's to make room for the Write-Progress output so it's not obscuring the console output for the first tests (see @nohwnd 's comments in #321 ).  Adding additional calls to Write-Progress in other places would make that effect even worse.

Remco-74:
Dave,

I wanted to make the progress reporting more smooth, to improve the user experience. 

A single progress bar that displays how far the tests have run seems a good way to do that. Incrementing the progress per script file seems to work well enough for typical test scripts (like the ones used in Pester itself). The other calls to Write-Progress in *It.ps1* caused screen flicker, since it essentially creates and removes a progress bar for every test run. That makes the test unreadable, and does not display total progress either.

I removed the *Write-Screen* line because it didn't make much sense to me. That only works correctly when you start with a clean PowerShell window, and you do not have any previous output already there. It doesn't make sense in other hosts either (for example in the PowerShell ISE), since a host can choose to display the progress bar in another way, or not at all.

dlwyatt:
That's all basically true, but the only reason we added those Write-Progress calls at all was so that there would be some indication of what test was in progress.  When there are long-running `It` blocks, Pester could appear to freeze.

If we were not going to have those particular Write-Progress calls (even though they do go by very fast for quick-running tests), we would probably just go back to not having any Write-Progress calls at all.

Remco-74:
Dave, some options;

An alternative solution may be to replace the Write-Progress calls in *It.ps1* by *Write-Verbose*, so that way users can use the *-Verbose* switch to get more feedback when a test takes long to execute.

The new call to *Write-Progress* can be made optional by implementing a new *-ShowProgress* switch on *Invoke-Pester*.

What would you suggest?

dlwyatt:
Write-Verbose could be an option, though that might also create a lot of other noise in the form of verbose output from commands that the tests are running.  (Import-Module, etc.)

As for adding a new switch to Pester, I don't think that's necessary when there's a `ProgressPreference` variable.  People who really don't want to see the Progress output can already just set that variable to 'SilentlyContinue'

@nohwnd , thoughts on this?

Remco-74:
I did some tests with a call to *Write-Verbose* in *Invoke-Test*, but it doesn't seem to be very useful. It it just creates an extra line for every TestResult line that Pester writes itself. I also noticed that Pester doesn't use Verbose output much in other code either (only some output from code coverage). I would not recommend adding the *Write-Verbose* here.

dlwyatt:
That's why we went with Write-Progress at first.  The value is in being able to see what test is running _before_ Pester writes the test result, but by using Write-Progress, that information only stays visible until the test is finished and the normal console output is displayed anyway.

At this point, I think we should just leave this alone.  The current calls to Write-Progress do add value for some situations, and can be easily disabled with a preference variable if they're annoying in others.

theficus:
I just wanted to add that adding the 3 lines is really just covering an implementation detail. It assumes that when a script is run that the text will always be aligned to the top of the screen which simply isn't true. Literally nothing else I'm aware of that uses progress reporting uses this hack. Most of the time I run it just looks silly to see those three empty lines appended to my output.

I conceptually understand why this was done but I just don't think it's necessary.

PR #420, opened 2015-10-15
Fix XML reporting on .NET Core

andschwa:
Create and pass an `IO.FileStream` to `XmlWriter.Create()`, since on .NET Core, it does not have an overload for a file path in a string.


dlwyatt:
Thanks!

Issue #413, opened 2015-09-29
build.bat throws quotation errors


prikkeldraad:
When I run Build.bat in a directory containing spaces I get the following error:
dir : Cannot find path 'P:\Mijn' because it does not exist.
At line:1 char:22
+ $psakeDir = ([array](dir P:\Mijn Documenten\test\trunk\pester\Vendo ...
+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (P:\Mijn:String) [Get-ChildItem], ItemNotFound
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand

I rewrote part of the script to fix this:
```
@echo off

if '%1'=='/?' goto help
if '%1'=='-help' goto help
if '%1'=='help' goto help
if '%1'=='-h' goto help

"%~dp0vendor\tools\NuGet.exe" install "%~dp0vendor\packages.config" -o "%~dp0vendor\packages"
@SET cmd=$psakeDir = ([array](dir """%~dp0vendor\packages\psake.*"""))[-1]; ".$psakeDir\tools\psake.ps1" build.psake.ps1 %* -ScriptPath "$psakeDir\tools" ; if ($psake.build_success -eq $false) { exit 1 } else { exit 0 }
powershell -NoProfile -ExecutionPolicy Bypass -Command ^ "%cmd%"
goto :eof

:help
@SET cmd=$psakeDir = ([array](dir """%~dp0vendor\packages\psake.*"""))[-1]; ".$psakeDir\tools\psake.ps1 build.psake.ps1" -docs -ScriptPath "$psakeDir\tools"
powershell -NoProfile -ExecutionPolicy Bypass -Command ^ "%cmd%"
```

Hope it helps

nohwnd:
@prikkeldraad  Great, thanks for the help! Could you possibly create a PullRequest for your changes? In case you need any help with that, please let me know.

PR #424, opened 2015-11-03
#413 quotations added for paths which include spaces

prikkeldraad:


Build.bat
```diff
@@ -5,11 +5,11 @@ if '%1'=='-help' goto help
...
-  "$psakeDir = ([array](dir %~dp0Vendor\packages\psake.*))[-1]; .$psakeDir\tools\psake.ps1 build.psake.ps1 %* -ScriptPath $psakeDir\tools ; if ($psake.build_success -eq $false) { exit 1 } else { exit 0 }"
+"%~dp0vendor\tools\NuGet.exe" install "%~dp0vendor\packages.config" -o "%~dp0vendor\packages"
+@SET cmd=$psakeDir = ([array](dir """%~dp0vendor\packages\psake.*"""))[-1]; ".$psakeDir\tools\psake.ps1" build.psake.ps1 %* -ScriptPath "$psakeDir\tools" ; if ($psake.build_success -eq $false) { exit 1 } else { exit 0 }
+powershell -NoProfile -ExecutionPolicy Bypass -Command ^ "%cmd%"
 goto :eof
 
 :help
-powershell -NoProfile -ExecutionPolicy Bypass -Command ^
-  "$psakeDir = ([array](dir %~dp0Vendor\packages\psake.*))[-1]; .$psakeDir\tools\psake.ps1 build.psake.ps1 -docs -ScriptPath $psakeDir\tools"
+@SET cmd=$psakeDir = ([array](dir """%~dp0vendor\packages\psake.*"""))[-1]; ".$psakeDir\tools\psake.ps1 build.psake.ps1" -docs -ScriptPath "$psakeDir\tools"
```
dlwyatt: I'm a little confused by the extra quotation marks around the argument to the dir command in this line (and the same on line 9).  Are they necessary?  Quick test seemed to work fine with just a single pair of quotes.

It's also not necessary to put the quotation marks around the PowerShell tokens such as $psakeDir, because they don't get expanded until after they're passed to a command, and spaces in those values don't matter.  (But they also don't hurt anything, so I'm fine with that part either way.)

nohwnd:
@dlwyatt Any more comments on this, can it be merged? Or you'd prefer if the extra quotes were removed?

dlwyatt:
Just want to understand why they're there.  The code would be a bit easier to read without a bunch of extra punctuation if it's not necessary.

PR #426, opened 2015-11-17
Improve stack trace and exception console output

alx9r:
*Note: This is my first PR.  While I'm fairly confident about PowerShell and Pester, I'm a git/github novice.*

------

This pull request improves the stack traces and exceptions that Pester outputs to the console.  This is similar to PR #228 except that it (hopefully) addresses the concerns with that PR.

Stack trace outputs are complicated by the fact that `ErrorRecord.ScriptStackTrace` is not available in PowerShell 2.  I did not find a good substitute for `.ScriptStackTrace` under PowerShell 2.  Rather, I opted to use `ErrorRecord.InvocationInfo` to produce a single-line "stack trace" for PowerShell 2.  This is the same technique as is already used in `Get-PesterResult` to produce `TestResult.stackTrace`.

The improvements of this PR over master and #228 are as follows:
 * Console Output:
     * When `ErrorRecord` is a result of `PesterAssertionFailed`:
        * If `.ScriptStackTrace` is **not** available: No changes. The "Expected...But was" messages stay the same and there are no changes to the stack trace because `.ScriptStackTrace` is unavailable.
        * If `.ScriptScriptTrace` is available: The typical single line stack trace changes from eg. `at line: 36 in C:\temp\scriptpath.ps1` to `at <ScriptBlock>, C:\temp\scriptpath.ps1: line 36`.  The rest of the output is "Expected...But was", etc and remains unchanged.
     * When an exception other than `PesterAssertionFailed` produces the `ErrorRecord`:
        * Messages from exceptions are prefixed with their type name.
        * The full chain of nested exceptions are output instead of just the outer exception.
        * When `.ScriptStackTrace` is available the full unit-under-test stack trace is output.  The lines of the stack trace that reference Pester internals are omitted.
 * Implementation:
    * Adds plumbing to get `.ErrorRecord` into `TestResult` for later use.  This aspect is similar to #228 which didn't seem to be a concern at that time.
    * Implements `ConvertTo-FailureLines` - This does the heavy-lifting of extracting, re-ordering, and formatting just the right script trace and exception details from `ErrorRecord`.
    * Adds call to `ConvertTo-FailureLines` to Write-PesterResult.

### Console Output Comparison

These comparisons are using PowerShell 4.

image

On the left is prior to this PR.  On the right is with this PR.  Note the following:
 * The first `throws` shows the full three lines of stack trace from the unit under test.
 * The second `throws` shows the nested exceptions and name prefixes `ArgumentException` and `FormatException`.
[...]

PR #439, opened 2015-12-15
Spelling change and new example

lboening:
I found a spelling error. I added one example.

### DIFF
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -177,7 +177,12 @@ This will only run the Describe block named "Add Numbers"
 .Example
 Invoke-Pester -EnableExit -OutputXml "./artifacts/TestResults.xml"
 
-This runs all tests from the current directory downwards and writes the results according to the NUnit schema to artifatcs/TestResults.xml just below the current directory. The test run will return an exit code equal to the number of test failures.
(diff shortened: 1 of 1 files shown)


nohwnd:
Hello, thanks a lot for the PR. :+1:  I saw that you made the PR from your master so I wanted to merge and close it as fast as I could before you start making more changes to your master, because those changes would be also seen in the PR. For that reason I cherry picked a single commit and merged it.

PR is not a snapshot of the branch, it's just a link to the branch, so a better way to do PR is to create a separate "feature branch" where you put the changes that you want in our codebase. We then have time to investigate the changes, while you can happily hack away in some other branch or on your master. It is also good to PR only a single commit without any fast forwarding info in it. You can achieve that by rebasing and squashing all your commits. See [this thread](https://github.com/pester/Pester/pull/426#issuecomment-160067554) and the links for more info.

nohwnd:
Merged without including the merge info.

lboening:
Oh, that's just fine. I had a feeling I was rushing it a bit, but Github will let you press all sorts of buttons without hesitation. I have read the thread, it makes more sense to create a feature branch.

PR #464, opened 2016-01-31
Run on nano server (and Pi 2 with Windows IoT)

nohwnd:
There were two changes that needed to be made to make Pester work on
Nano server. First I had to conditionally import Get-CimSession instead
of Get-WmiObject that is not available on nano server.  The other thing
that I replaced was the BraceFinder that was written and compiled via
Add-Type. The PowerShell implementation is about 10 times slower (but
it's 1 vs 10 milliseconds).

The rest of the fixes needed to be done in test files.
1) DirectoryInfo serializes to just Name instead of FullName which
caused the most of the failed tests, so I had to resort to storing the
full name in the Global:TestDrive variable instead of the DirectoryInfo
object. This is a breaking change, but I don't see a better option.

2) I had to conditionally skip some of the xml tests because the
required properties and methods to handle schema validation are not
available on .net core.

### DIFF
--- a/Functions/Assertions/PesterThrow.Tests.ps1
+++ b/Functions/Assertions/PesterThrow.Tests.ps1
@@ -55,7 +55,7 @@ InModuleScope Pester {
     }
 
     Describe 'PesterThrowFailureMessage' {
-        $testScriptPath = Join-Path $TestDrive.FullName test.ps1
+        $testScriptPath = Join-Path $TestDrive test.ps1
 
         It 'returns false if the actual message is not the same as the expected message' {
             $unexpectedErrorMessage = 'unexpected'
@@ -75,7 +75,7 @@ InModuleScope Pester {
     }
 
     Describe 'NotPesterThrowFailureMessage' {
-        $testScriptPath = Join-Path $TestDrive.FullName test.ps1
+        $testScriptPath = Join-Path $TestDrive test.ps1
 
         It 'returns false if the actual message is not the same as the expected message' {
             $unexpectedErrorMessage = 'unexpected'
--- a/Functions/Context.Tests.ps1
+++ b/Functions/Context.Tests.ps1
@@ -8,7 +8,7 @@ Describe 'Testing Context' {
         $parameter = $command.Parameters['Fixture']
         $parameter | Should Not Be $null
 
-        $attribute = $parameter.Attributes | Where-Object { $_.TypeId -eq [System.Management.Automation.ParameterAttribute] }
+        $attribute = $parameter.Attributes | Where-Object { $_ -is [System.Management.Automation.ParameterAttribute] }
         $isMandatory = $null -ne $attribute -and $attribute.Mandatory
 
         $isMandatory | Should Be $false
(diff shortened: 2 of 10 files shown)


nohwnd:
I traced back the issue with BaseName vs. FullName serialization and turns out it's a reported issue, which is fixed in newer version of the framework. 
https://github.com/dotnet/corefx/issues/2632
https://github.com/dotnet/corefx/pull/5437

nohwnd:
@dlwyatt Would it be possible to add Nano server to your test agent pool on the community CI server? What do you think about rolling back the BraceFinder implementation from C# to PowerShell? I don't remember the results of the speed tests that originally motivated that change. But from the test run duration on the CI server I can't tell any notable difference, (that is if I am reading that correctly): 

image

dlwyatt:
It's something I'd like to support eventually, but it won't happen right away.  The build.powershell.org environment is hosted on AWS, which doesn't have any server 2016 or Nano AMIs yet; I'd have to build one out.  Will also have to get the TeamCity build agent set up on Nano (which requires the Oracle JRE), along with Amazon's EC2Config agent.

nohwnd:
I understand. I suggest to merge this PR and release it as a new minor version. This concrete version will be able to run on Nano, but future release won't be guaranteed to run on Nano server, as there is no automatic process to notify us when something broke. What do you think?

dlwyatt:
Works for me, though if they're already going to fix the FileInfo/DirectoryInfo to String code, then we probably don't need to worry about changing those parts of Pester.

nohwnd:
Right. I'll just keep this PR hanging here then.

kilasuit:
@dlwyatt were you not looking at moving the Build agents to Azure at one point? Is there anything that you would need to get done to create a Nano build agent that I could help out with at all?

dlwyatt:
I did look into that, but it turned out that Azure VM pricing made it more cost effective to stay on AWS for now, even with Azure's per-minute billing for build agents.  Because the TeamCity server itself is up 24/7, and AWS has reserved instance pricing, we still wind up paying less overall (even with each build charging for 1 hour of time on the agents that only use 5-10 minutes.)

The tipping point where Azure would become more cost-effective than AWS would happen at over 3700 builds per month (and we're nowhere close to that utilization yet.)

I did investigate hosting the TeamCity server on AWS and the build agents on Azure, but there are extra challenges there:

- TeamCity doesn't support HTTPS for server-to-agent communication.  It's a hot feature request, but they just say "put them on the same private network, or use VPN."
- I was not able to get a site-to-site VPN connection between Azure and AWS working without introducing a new VM on one side or the other.  I really, really wanted to be able to just use their built-in VPN features and get them talking to one another, but they both make certain assumptions about the encryption / protocols, and those assumptions don't agree.  An extra VM with 100% uptime wipes out any cost savings we would gain by moving the build agents into Azure.

TL;DR - AWS costs less for now.  :)

it-praktyk:
@dlwyatt, @nohwnd what about this PR?

Windows Server 2016 Nano is available on AWS now - [Windows AMIs](https://aws.amazon.com/windows/resources/amis/).

nohwnd:
I cannot do much about this till Dave adds the build agents to build.powershell.org, I have no power over that service, I administrate only the Pester build, not the whole service 🙂 . Once that is done I can go and configure the builds if needed. It would be also be really nice to be able to run against Linux or docker hosted on Linux. (and MacOS but that's not gonna happen right?)

it-praktyk:
I did a progress for #639 (Linux compatibility) last weekend - I just need to achieve Windows compatibility too ;-)

PR #471, opened 2016-02-15
Fix of $args handling

PetSerAl:
`$args` automatic variable is not used by advanced functions, so it is valid to use `$args` as normal parameter for advanced function. Pester should allow to mock advanced function with such parameter. Same applies to cmdlets with `Args` parameter. Also there is an error when mock for advanced function called with dot operator.

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -1533,3 +1533,49 @@ Describe 'Mocks with closures' {
         TestClosure -Closure | Should Be 'Variable resolved from closure'
     }
 }
+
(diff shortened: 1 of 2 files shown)


dlwyatt:
Thanks!  What was the error regarding the dot operator?

PetSerAl:
See my first test `It 'Advanced function mock should be callable with dot operator'` it fails without fix.
When you call advanced function with dot operator from simple function, then `$args` from simple function happens to be in local scope of advanced function. That cause that mocked function incorrectly capture `$args` from local scope.

PR #473, opened 2016-02-17
Added support for ErrorId to PesterThrow

TeflonJim:
Speculative support for ErrorId handling in PesterThrow.
Significant increase in complexity of generated messages from PesterThrow, switched to StringBuilder for greater control.
Extended tests to verify backward compatibility and added features.


TeflonJim:
Looks like there's more (white space) failing than I've fixed. I'll sort it tonight, too hard to do from here.

PR #474, opened 2016-02-18
Allow mocking commands with single quote in name or module name

PetSerAl:
Single quote does not properly escaped, when command name and module name values got embedded into `MockPrototype`.

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -1,4 +1,4 @@
-Set-StrictMode -Version Latest
+﻿Set-StrictMode -Version Latest
 
 function FunctionUnderTest
 {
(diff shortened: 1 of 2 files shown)


dlwyatt:
The CodeGeneration stuff was only added in PSv5, if I remember correctly; Pester needs to be compatible down to v2.  Also, who the heck is naming their functions with those goofy characters anyway?  :)

PetSerAl:
It have backup path for pre v5. For me all tests passed on v2.

dlwyatt:
Ah, missed that when I glanced at the change, sorry.  :)  Still not sure that anyone would actually encounter this problem, but I suppose there's no reason not to merge it.

PR #477, opened 2016-02-21
Mocking cmdlet without positional parameters

PetSerAl:
If advanced function does not specify any positional parameters, then PowerShell make all function parameters to be positional implicitly. This is not true for cmdlets. So that, mock of cmdlet can have different parameter binding behavior compared to original cmdlet. In v3+ you can disable implicit positional parameter binding behavior by specifying `PositionalBinding=$false` in `CmdletBinding` attribute.

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -1597,3 +1597,58 @@ Describe 'Single quote in command/module name' {
     }
 
 }
+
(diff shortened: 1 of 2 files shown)


dlwyatt:
For the record, I'm impressed!  :)  The Mocking code is by far the messiest, hardest to follow part of Pester, and you keep finding these little improvements that I'd never even noticed.

Issue #470, opened 2016-02-13
Adjust failure output to be VSCode problemMatcher friendly


rkeithhill:
I'm trying to add `problemMatcher` support to a pester task for the PoweShellEditorServices extension for VSCode.  The problem is that the `problemMatcher` uses line-oriented regexes to extract the following required fields: `file`, `line`, `message`.  You can specify a `problemMatcher` in a task like so:
```json
"tasks": [
    {
        "taskName": "Pester",
        "isTestCommand": true,
        "problemMatcher": {
            "owner": "powershell",
            "fileLocation": ["absolute"],
            "severity": "error",
            "pattern": [
                {
                    "regexp": "^\\s*\\[-\\](.*)$",
                    "message": 1
                },
                {
                    "regexp": "^.*$"
                },
                {
                    "regexp": "^.*$"
                },
                {
                    "regexp": "^\\s+at line: (\\d+) in (.*)$",
                    "line": 1,
                    "file": 2,
                    "message": 2
                }
            ]
        }
    }
]
```
This will match the following type of Pester output:
```
 [-] Should accept pipeline input to Path 111ms
   Expected: {C:\Users\Keith\.vscode-insiders\extensions\ms-vscode.PowerShell-0.3.1\examples\foo[1].txt}
   But was:  {}
   at line: 47 in C:\Users\Keith\.vscode-insiders\extensions\ms-vscode.PowerShell-0.3.1\examples\PathProcessing.Tests.ps1
   47:         Get-ChildItem -LiteralPath "$pwd\foo[2].txt" | Import-FileNoWildcard | Should Be "$PSScriptRoot\foo[1].txt"
```
This results in a reasonable VSCode experience:

![vscodepesterproblemmatcher](https://cloud.githubusercontent.com/assets/5177512/13025044/82fe84f2-d1bc-11e5-8110-8ed658002008.png)
Here is where it falls down.  The above handles the case when there are two lines of error info between the initial `[-]` line and the line with the filename/line number info.  Here's a different error message that has four lines between those two locations:
```
Describing Verify Path Processing for Non-existing Paths Allowed Impl
 [-] Processes non-wildcard absolute path to non-existing file via -Path param 483ms
   Expected string length 92 but was 91. Strings differ at index 88.
   Expected: {C:\Users\Keith\.vscode-insiders\extensions\ms-vscode.PowerShell-0.3.1\examples\READMENew2.md}
[...]

PR #479, opened 2016-02-21
Initial fix for #470

dlwyatt:
@rkeithhill , start with this and let me know what you think.  This was just about the quickest and laziest way I could implement this, just to get the ball rolling.  If the new option is specified, the first line of the stack trace gets output in between the failed test header line and the error message.  We can tweak this as needed.  Here's how the usage works (same as you'd expect with New-PSSessionOption / etc):

```posh
$option = New-PesterOption -IncludeVSCodeMarker
Invoke-Pester -PesterOption $option
```

### DIFF
--- a/Functions/PesterState.ps1
+++ b/Functions/PesterState.ps1
@@ -6,7 +6,8 @@ function New-PesterState
         [String[]]$TestNameFilter,
         [System.Management.Automation.SessionState]$SessionState,
         [Switch]$Strict,
-        [Switch]$Quiet
+        [Switch]$Quiet,
+        [object]$PesterOption
     )
 
     if ($null -eq $SessionState) { $SessionState = $ExecutionContext.SessionState }
@@ -18,7 +19,8 @@ function New-PesterState
             [String[]]$_testNameFilter,
             [System.Management.Automation.SessionState]$_sessionState,
             [Switch]$Strict,
-            [Switch]$Quiet
+            [Switch]$Quiet,
+            [object]$PesterOption
         )
 
         #public read-only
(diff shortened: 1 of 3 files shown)


rkeithhill:
Cool. I believe this will work just fine. 

BTW as I've gotten more experience with the errors and warnings list in VSCode, I've found that it's not just a list but VSCode also decorates the failing test source with red squiggles and displays the `problemMatcher` `message` field text in a hover tip.  At first I was using the `It` description of the test as the `message` but that is a bit redundant when you pop the hover tip in the test source e.g.:

vscodepestermessagetitle

That got me thinking that maybe I always grab the message from the third line.  VSCode only allows you to pick the `message` text from a single line but I think the third line might work OK.  Check it out:

vscodepestermessage

What do you think?

dlwyatt:
Only problem with that is that the most common test by far is `Should Be`, and the error message for that is always spread across at least two lines.  (More if the values are strings.)

```
Describing Testing
 [-] Compares booleans 50ms
   at <ScriptBlock>, C:\Users\dlwya_000\OneDrive\Source\Temp\test.Tests.ps1: line 15
   Expected: {False}
   But was:  {True}
   15:         $true | Should Be $false
   at <ScriptBlock>, C:\Users\dlwya_000\OneDrive\Source\Temp\test.Tests.ps1: line 15
 [-] Compares strings 67ms
   at <ScriptBlock>, C:\Users\dlwya_000\OneDrive\Source\Temp\test.Tests.ps1: line 19
   String lengths are both 10. Strings differ at index 7.
   Expected: {String two}
   But was:  {String one}
   ------------------^
   19:         'String one' | Should Be 'String two'
   at <ScriptBlock>, C:\Users\dlwya_000\OneDrive\Source\Temp\test.Tests.ps1: line 19
```

rkeithhill:
Yeah, that is less than ideal.  OK, I'll stick with the `It` description for now.  I think that is more handy - at least in the errors and warnings list.

daviwil:
Awesome, thanks so much for working this out!  I think people will be pretty happy to have good Pester integration in VS Code :)

dlwyatt:
This will be 3.4.0.  Just need to tie up the loose ends, add some documentation and I'll get it released tonight.

PR #480, opened 2016-02-23
Make paths to be provider qualified

PetSerAl:
This change make Pester work even if someone remove build-in PowerShell drives: `Remove-PSDrive Alias,Function`. It seems, that `configuration` keyword in PowerShell is not happy if `function` drive is not exists. So the test involve it fails if you actually remove `function` drive, but other tests pass successfully.

### DIFF
--- a/Functions/Assertions/Exist.Tests.ps1
+++ b/Functions/Assertions/Exist.Tests.ps1
@@ -15,10 +15,10 @@ InModuleScope Pester {
             "TestDrive:\``[test``].txt"  | Should Exist
         }
 
-        It 'returns correct result for function drive' {
+        It 'returns correct result for function provider' {
             function f1 {}
 
-            'function:f1' | Should Exist
+            'function::f1' | Should Exist
         }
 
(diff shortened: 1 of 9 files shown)


dlwyatt:
I think I'm going to leave this one alone.  As you say, even DSC itself doesn't appreciate having the default PSDrives taken away.  Just file this one under "don't do that".  :)

PetSerAl:
Personally, I think it is bug in `configuration` keyword handling. It also does not work property if you mock `Set-Item`: `function Set-Item {'Mocked Set-Item'};configuration C{}`.
It does not harm to be more robust for such unusual cases. Especially, in most cases it just replace `:\` to `::` or `:` to `::`.
Just one case where I change `$ExecutionContext.InvokeProvider.Item.Set` to `$ExecutionContext.InvokeProvider.Item.New`: it seems that PowerShell v2 have a bug. It does not set item if path is literal, provider qualified and not yet exists. But, as you can see, this does not break any test.

PR #484, opened 2016-03-02
update SafeCommands to use Get-CimInstance if it exists, then try Get…

JamesWTruher:
…-WmiObject. If neither exist, then warn the user that the test results may not have all the system information. Updated TestResult to try harder to get system information (based on whether Get-CIMInstance or Get-Wmi exists). Also, change a couple of tests to be more portable when it comes to CoreCLR because TypeId does not exist as property on Parameter.

### DIFF
--- a/Functions/Context.Tests.ps1
+++ b/Functions/Context.Tests.ps1
@@ -8,7 +8,8 @@ Describe 'Testing Context' {
         $parameter = $command.Parameters['Fixture']
         $parameter | Should Not Be $null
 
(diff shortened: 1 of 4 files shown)


JamesWTruher:
This targets PowerShell on Nano, but is actually useful generally. Get-WMIObject is not available on Nano, but Get-CimInstance is available. Also, changes in .NET Core eliminated the TypeId property on Parameter, but an easy alternative is present. 
There is an existing .NET Core bug https://github.com/dotnet/corefx/pull/5437 which should address issues of Pester with regard to TESTDRIVE (which I saw in a different pull request)

dlwyatt:
Thanks!  :)

JamesWTruher:
thank you!

PR #485, opened 2016-03-06
Fix test failed sample

urasandesu:
I tried the sample in [README](https://github.com/pester/Pester/blob/2da66db16123f45c97193f775afbde0e3ce5874d/README.md), but it always failed: 
default

I investigated that, and I found that `Get-Version` and `Get-NextVersion` have to define at somewhere. I recommend that they are also in README with mocking reason.


dlwyatt:
Thanks!  :)

PR #502, opened 2016-04-02
Formatted About topics

juneb:
Applied standard formatting rules to Pester About topics. Among other things, this allows the About topics to be included in post-processing tools that look for standard elements in these otherwise unstructured text files.

I also corrected just a few grammatical errors. Really, just a few.


dlwyatt:
Thanks!  :)

juneb:
Completely self-serving. :)

Our help display tool couldn't display them properly with the non-standard
formatting.

I'll get back in and do some more substantive stuff, too.

PR #508, opened 2016-04-14
$count was null

adbertram:
$count was $null forcing -First on line 415 to error with a $null param arg. I'm creating the $count variable to ensure $count is always 1.


dlwyatt:
Looking at the code, I think it should just be initialized to zero, regardless of what the value of `$traceLines` happens to be.

adbertram:
That works too.

PR #511, opened 2016-04-17
fix #510

### LINKED ISSUE #510: The error produced by nested `It` blocks is obfuscated by an error in `ConvertTo-FailureLines`
### Repro

````Powershell
Describe describe {
    It outer {
        It inner { $true }
    }
}
````

Results in this error:

image

### Analysis

The [originating block of code](https://github.com/pester/Pester/blob/d656dc7a5f737f9e217161f9d0c6fcfab4ab3f2e/Functions/PesterState.ps1#L400-L410) uses variable `$count`. That line was originally written as part of #426 in 77b84c9517ab4ea0eeade99728ea5bfb83320fa7 like this:

````Powershell
Select-Object -First $count
````

I expected this to work because `-First` is typed as `[int]` and `[int]$null` evaluates to `0` in powershell.  But [these tests demonstrate](https://github.com/alx9r/ToolFoundations/blob/a661f5e41949ed2be768cd08660752aedcda9835/EnvironmentTests/casting.Tests.ps1#L28-L118) that implicit casting of parameters does not behave in that manner nor consistently.

That line happens to have been changed in 344aee669706bfdb2c686af12b52a6b65f43d7aa to this:

````Powershell
& $SafeCommands['Select-Object'] -First $count
````

That change is inconsequential to this problem, though, and this problem existed prior to that change in the original 77b84c9517ab4ea0eeade99728ea5bfb83320fa7.

### Fix

I have a working fix which is simply a matter of initializing `$count`.  This restores the helpful output error message as follows:

image

I'll create a PR as soon as I (re)figure out the right way to do that.

alx9r:


alx9r:
This PR should fix issue #510.

dlwyatt:
This line should already be in the master branch from a day or two ago.  :)  It was reported in #508

alx9r:
D'oh.  I guess I should have been working from the head instead of the last release.

alx9r:
Yep.  This was already fixed in 900d02880bfe11cdfb7693a59cc469536a1df0f4.

PR #521, opened 2016-04-24
Fix error messsage of Should Throw when null input is provided

### LINKED ISSUE #498: Error message not helpful when missing scriptblock
I forgot to enclose my input to Should Not Throw in curly braces. The message that I got about an ampersand looks like some code fails to handle a very predictable exception.

Can we please fix this?

    [-] Error occurred in Context block 403ms
      Expected: the expression not to throw an exception. Message was {The expression after '&' in a
 pipeline element produced an object that was not valid. It must result in a command name, a script
block, or a CommandInfo object.}...

nohwnd:
Fixes #498


juneb:
If this is too long, let me know and I'll fix it.

Input to 'Should [Not] Throw' is $null (invalid). Always enclose input to Throw in a scriptblock: 
{<input>} | Should [Not] Throw.

PR #536, opened 2016-05-19
Gherkin 3

Jaykul:
The final refactoring of Gherkin ...

- Added some more tests (and fixed them)
- Switched the gherkin library
- Rewrote the script to work with those new types.


Jaykul:
Ok, so ... the Gherkin.dll from NuGet requires .Net 4, so it's not going to work in PowerShell 2.
Personally, I'm ok with that... but I guess I need to wrap all the gherkin stuff in version checks?
That or we can ship the IKVM version for downlevel?

dlwyatt:
If we're going to ship the IKVM version at all, we may as well use it on all versions rather than having branches in the code.  But let's think on it a bit, and see what @nohwnd says as well.

Once again I'm wondering if this is the right time to just drop PSv2 support from Pester.  Those of us (myself included) who still write modules that are v2-compatible would have to also write our tests to be Pester v3-compatible, but I suspect that pool of people is shrinking every day.

Windows 7 / Server 2008 R2 (the only currently-supported OSes that can even run PSv2) are already out of mainstream support, but extended support will go on for another few years, and I'm not really that thrilled at the thought of supporting PSv2 until 2020.

Jaykul:
I dug a little, and making Gherkin compile against .Net 3.5 was trivial, so ... I did.

https://github.com/Jaykul/gherkin-dotnet/tree/35_Fix

Jaykul:
Ok, I literally have no clue what that build (non)output even means. I can't find a log or anything, is the module just not importing for some reason?

dlwyatt:
Don't worry about it, that's probably nothing to do with what you've changed.  I'll sort it out when I go through all the latest merge conflicts with master :)

Do you want to wait and see if the Gherkin people will take a PR for this, or just ship Pester with the version in your fork?

Jaykul:
I'm happy shipping the fork, it's a minor change, and I've already gotta absorb any changes they make anyway.  Ultimately, it's worth it to keep supporting .net 3.5

PR #537, opened 2016-05-19
Add ContainMultiline tests

Jaykul:


Jaykul:
I am confused as to how ContainMultiline got in without tests, but here they are...

PR #546, opened 2016-05-29
New function to create test file for existing function

benny-gold:


dlwyatt:
As was mentioned in the Gitter chat, this probably isn't needed since New-Fixture will already skip files that exist.

PR #562, opened 2016-06-13
Updates to better support Pester on Nano

JamesWTruher:
Update Coverage tests to use a different resource. The WindowsFeature resource
doesn't exist on Nano, but the Environment resource does
Update the TestResult tests to only validate against schema when the Validate
method is available on the system. Nano, for example, doesn't have this method
available because of the differences between Core and Full CLR

### DIFF
--- a/Functions/Coverage.Tests.ps1
+++ b/Functions/Coverage.Tests.ps1
@@ -256,11 +256,14 @@ InModuleScope Pester {
 
                 configuration MyTestConfig   # does NOT trigger breakpoint
                 {
+                    # add this line to avoid a warning during test
+                    Import-DscResource -ModuleName PSDesiredStateConfiguration # Triggers breakpoint
                     Node localhost    # Triggers breakpoint
                     {
-                        WindowsFeature XPSViewer   # Triggers breakpoint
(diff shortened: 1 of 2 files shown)


JamesWTruher:
fixed the trailing space

dlwyatt:
Looks like the tests didn't get run again when you rebased.  Can you send another commit to kick them off?  Want to make sure the DSC stuff passes in PSv4; that had failed the last time.

dlwyatt:
Actually, maybe I can kick off the tests manually; will try now.

Functions/Coverage.Tests.ps1
```diff
@@ -282,21 +285,21 @@ InModuleScope Pester {
             Enter-CoverageAnalysis -CodeCoverage "$root\TestScriptWithConfiguration.ps1" -PesterState $testState
 
             It 'Has the proper number of breakpoints defined' {
-                $testState.CommandCoverage.Count | Should Be 7
+                $testState.CommandCoverage.Count | Should Be 8
```
vors: Judging from comments in the scripts, this should stay `7`: new resource parameter `# does NOT trigger breakpoint`

JamesWTruher:
closing stale PR

PR #579, opened 2016-07-14
Make Should Throw case insensitive

### LINKED ISSUE #577: Request for throw to be a case insensitive match
The throw condition is a case sensitive match on any string you pass to it. For example:

```
{ throw "Luke Skywalker" } | Should throw "luke skywalker"
```

Fails. Whereas:

```
{ throw "Luke Skywalker" } | Should throw "Luke Skywalker"
```

Passes.

When I'm writing text to catch for throw's I just write in lowercase rather than having to make sure the case is exactly correct. The point would be to catch the error text rather than worrying about what case it is and it would be more convenient if Pester matched the text in the exception message regardless of case.

The problem seems to be in the Functions\Assertions\PesterThrow.ps1 file:

```
function Get-DoMessagesMatch($value, $expected) {
    if ($expected -eq "") { return $false }
    return $value.Contains($expected)
}
```

Which uses the case sensitive .Contains() which could be changed to -contains whcih is case insensitive:

```
function Get-DoMessagesMatch($value, $expected) {
    if ($expected -eq "") { return $false }
    return $value -contains $expected
}
```

I've verified that this works as I expected.

nohwnd:
Fixes #577

### DIFF
--- a/Functions/Assertions/PesterThrow.ps1
+++ b/Functions/Assertions/PesterThrow.ps1
@@ -51,7 +51,7 @@ function PesterThrow([scriptblock] $ActualValue, $ExpectedMessage, $ErrorId, [sw
 
 function Get-DoMessagesMatch($ActualValue, $ExpectedMessage, $ExpectedErrorId) {
     if ($ExpectedMessage -eq "") { return $false }
-    return $ActualValue.Contains($ExpectedMessage)
+    return $ActualValue -like "*$ExpectedMessage*"
 }
 
 function Get-ExceptionLineInfo($info) {
--- a/Functions/Assertions/PesterThrow.Tests.ps1
+++ b/Functions/Assertions/PesterThrow.Tests.ps1
@@ -18,6 +18,13 @@ InModuleScope Pester {
             { throw $expectedErrorMessage } | Should -Throw $expectedErrorMessage
         }
 
+        It "returns true if the statement throws an exception and the actual error text matches the expected error text (case insensitive)" {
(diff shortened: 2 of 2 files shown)


Functions/Assertions/PesterThrow.ps1
```diff
@@ -51,7 +51,7 @@ function PesterThrow([scriptblock] $ActualValue, $ExpectedMessage, $ErrorId, [sw
 
 function Get-DoMessagesMatch($ActualValue, $ExpectedMessage, $ExpectedErrorId) {
     if ($ExpectedMessage -eq "") { return $false }
-    return $ActualValue.Contains($ExpectedMessage)
+    return $ActualValue -like "*$ExpectedMessage*"
```
dlwyatt: Might be some potential for bugs here if `$ExpectedMessage` contains special characters.  Maybe this would be safer:

```posh
return $ActualValue -match [regex]::Escape($ExpectedMessage)
```

dlwyatt: Or just use .NET:

```posh
return [string]::Compare([string]$ActualValue, [string]$ExpectedMessage, $true) -eq 0
```

dlwyatt: Or `return [string]::Equals([string]$ActualValue, [string]$ExpectedMessage, [System.StringComparison]::InvariantCultureIgnoreCase)`.  Lots of ways to skin that particular cat.  :)

nohwnd:
Will be implemented in another PR.

PR #591, opened 2016-08-15
Added InconclusiveCount to Invoke-Pester output

timker:
InconclusiveCount count was missing with -passthru


dlwyatt:
Thanks!  :)

Issue #584, opened 2016-07-22
AfterAll scope is incorrect, if BeforeAll throws


vors:
Consider this example:

``` powershell
Describe 'Problem with exceptions in BeforeAll' {
    Context 'foo' {
        BeforeAll {
            Write-Host "foo:BeforeAll"
            throw "before"
        }

        AfterAll {
            Write-Host "foo:AfterAll"
            throw "after"
        }

        It 'something' {
            1 | Should Be 1
        }
    }

    Context 'bar' {
        BeforeAll {
            Write-Host "bar:BeforeAll"
        }

        AfterAll {
            Write-Host "bar:AfterAll"
        }

        It 'something' {
            1 | Should Be 1
        }
    }
}
```

I expect that first Context will fail, but the second one would be executed just fine.

Instead, here is what I see in Pester v.3.4.1
Note that `foo:AfterAll` is incorrectly displayed twice, so for me it looks like some registration of BeforeAll/AfterAll doesn't work properly, if these guys are throwing.

```
Describing Problem with exceptions in BeforeAll
   Context foo
foo:BeforeAll
    [-] Error occurred in Context block 29ms
      RuntimeException: before
      at <ScriptBlock>, F:\dev\PowerShell2\src\repro.Tests.ps1: line 5
      at Invoke-Blocks, C:\Users\sevoroby\Documents\WindowsPowerShell\Modules\pester\3.4.1\Functions\SetupTeardown.ps1: line 134
      at Invoke-TestGroupSetupBlocks, C:\Users\sevoroby\Documents\WindowsPowerShell\Modules\pester\3.4.1\Functions\SetupTeardown.ps1: line 113
foo:AfterAll
    [-] Error occurred in Describe block 14ms
      RuntimeException: after
      at <ScriptBlock>, F:\dev\PowerShell2\src\repro.Tests.ps1: line 10
      at Invoke-Blocks, C:\Users\sevoroby\Documents\WindowsPowerShell\Modules\pester\3.4.1\Functions\SetupTeardown.ps1: line 134
      at Invoke-TestGroupTeardownBlocks, C:\Users\sevoroby\Documents\WindowsPowerShell\Modules\pester\3.4.1\Functions\SetupTeardown.ps1: line 124
foo:AfterAll
```

it-praktyk:
What I see in Pester 4.0.6-RC

```
PS >Invoke-Pester -Script <FOLDER_PATH>\test.ps1
[...]

PR #610, opened 2016-09-22
an error in teardown blocks can put pester in a broken state

JamesWTruher:
this ensures that pester will remove testdrive and exit describe properly in the case the AfterAll throws

I _think_ the right behavior is that if tear down throws an error should be reported, but it's certainly a point for discussion.


vors:
This is not exactly, but very similar problem to https://github.com/pester/Pester/issues/584

JamesWTruher:
closing for now, I'll provide an update soon

PR #616, opened 2016-09-29
Forward merge 'master' to 'DevelopmentV4'

nohwnd:


nohwnd:
Fails on posh2 because of out-host. Need to investigate.

Issue #625, opened 2016-10-24
Null reference exception when upgrading from Pester 3.3.14 to 3.4.3


splatteredbits:
PowerShell 4/Windows 2012 R2

I just upgraded to 3.4.3 from 3.3.14. When I run one of my test fixtures, I get this exception:

> RuntimeException: You cannot call a method on a null-valued expression.
> at Test-IsClosure, D:\Jenkins_Home\jobs\WhsInit\workspace\Arc\Pester\Functions\Mock.ps1: line 1440
> at Mock, D:\Jenkins_Home\jobs\WhsInit\workspace\Arc\Pester\Functions\Mock.ps1: line 180
> at New-Mock, D:\Jenkins_Home\jobs\WhsInit\workspace\Tests\Initialize-Computer.Tests.ps1: line 76
> at <ScriptBlock>, D:\Jenkins_Home\jobs\WhsInit\workspace\Tests\Initialize-Computer.Tests.ps1: line 831

I'm not sure what the cause is, but I've made a fix and will create a pull request for it.

dlwyatt:
Can you give a repro of this error?  Get-ScriptBlockScope should never be returning `$null`, unless MS makes some changes to PowerShell internals (and the versions of Windows / PowerShell / Pester that you listed have already been tested.)

splatteredbits:
I spent a couple hours trying to distill the problem down to a reasonable, my-company-agnostic test case, and couldn't do it. I'm willing to send you a zip of what I got, but unfortunately, I can't post it here. I'd have to send it to you via another mechanism.

dlwyatt:
No problem.  My email address is here on github; can send me a link to however you want to share it (onedrive, etc)

splatteredbits:
Wrong ticket. It thought this was #627. I was able to distill this down to:

```
Describe 'Mock when creating a scriptblock from a string' {
    {
        Mock -CommandName 'Get-Content' -Verifiable -MockWith ([scriptblock]::Create(""))
    } | Should Not Throw
}
```

It looks like Pester doesn't like script blocks created from strings.

Is it OK if I change the title of this to "Mock fails when MockWith script block is created from a string"? Not sure why I didn't do this work yesterday.

dlwyatt:
That's perfect, thanks!  I need to look through other bits of Pester code to see what that assumption (Get-ScriptBlockScope never null) might break.

PR #626, opened 2016-10-24
Fixing issue #625: Test-IsClosure throws a null-valued expresesion ex…

splatteredbits:
Here's my fix for issue #625.


dlwyatt:
Fixed in next v4 release

PR #634, opened 2016-11-09
Created New-MockObject function

adbertram:


dlwyatt:
I'm excited about this one, but it does need some sort of test, even if it's just a basic one to make sure it works on all supported PS versions.  :)

adbertram:
I tested it on PSv2 but just didn't include the test. I'll do that.

Functions/New-MockObject.ps1
```diff
@@ -0,0 +1,51 @@
...
+    <#
+    .SYNOPSIS
+        This function instantiates a .NET object from a type. The assembly for the particular type must be
+        loaded.
+
+    .PARAMETER Type
+        The .NET type to create an object from.
+
+    .EXAMPLE
+        New-MockObject -Type 'System.Runtime.Serialization.Formatterservices'
```
nohwnd: The example fails with: Cannot create abstract class. The next example on line 13 and 14 is imho unnecessary. Wrapping the call to GetUninitializedObject is the goal of your function, so why expose it's internals. I suggest you make the example `System.Diagnostics.Process`. And show it's output, if you have to: 
```powershell
New-MockObject -Type 'System.Diagnostics.Process'

Handles  NPM(K)    PM(K)      WS(K) VM(M)   CPU(s)     Id  SI ProcessName                                                                                                                                                                            
-------  ------    -----      ----- -----   ------     --  -- -----------                                                                                                                                                                            
              0        0          0     0                                    
```

PR #635, opened 2016-11-09
Added new-mockobject function

adbertram:


Functions/New-MockObject.Tests.ps1
```diff
@@ -0,0 +1,10 @@
+describe 'New-MockObject' {
+
+    . .\New-MockObject.ps1
```
dlwyatt: There's no need for this.  Pester files are already loaded when we're running our own tests, and for some reason this line is failing anyway.

adbertram:
Is there a PR request guideline somewhere that I missed? It'd be nice for a first time PRer like myself.

nohwnd:
Not really, if you run all the tests locally (I usually do .\bin\pester.bat to make sure I am running in a clean powershell instance.) then it should highlight most of the errors. In this case it's the problem that Dave commented on. (the function is in the Functions folder and all of those are loaded automatically). Another problem is that line 4 in that same file has some whitespace on it, but there should be just a new line and nothing else.

adbertram:
I was not aware of a built in test suite to run. What are the recommended steps not assuming any prior knowledge to submitting Github PRs?  I can create a doc for others.

nohwnd:
You write a failing test for the thing you are fixing. You make a change, you run all the tests that pester has. You test as much as you can on all the supported PowerShell versions (on their native OSs). When everything passes and you are happy with your code you make a PR. Ideally you also rebase on the latest upstream master and squash all your commits into one.


Jakub Jareš

PowerShell & Pester
t @nohwnd

10. 11. 2016 v 18:23, Adam Bertram <notifications@github.com<mailto:notifications@github.com>>:

I was not aware of a built in test suite to run. What are the recommended steps not assuming any prior knowledge to submitting Github PRs? I can create a doc for others.

nohwnd:
@adbertram Is there an actual use case where you used this way of creating new objects? I briefly looked at populating the actual properties and fields, and how to dumb it down to be useful for the average user who does no like to go knees deep in assemblies. So are there any objects that gave you hard time?

dlwyatt:
This helps us in the most tricky mocking scenarios: commands that have strongly typed parameters, where those types are tricky or impossible to create cleanly in a test. (No public constructor, or a constructor that does annoying things, etc.)

adbertram:
The original intention was to not have to get into the assemblies. I simply needed a way to create an object that could be passed to functions that had a parameter of that type. I didn't need to populate any properties. I used this extensively for various types from the TFS assemblies.

adbertram:
Yes, what Dave said. It was commands that had strongly typed parameters for types that had no public constructor. This was a method I found that instantiated objects from classes where the constructor did not matter.

nohwnd:
Yeah yeah I get that, I remember there was an issue with some Sharepoint or AD related types here. What I am trying to understand is how injecting an empty object is helpful. Are those just dummy objects that enable you to run the function but you don't use them to carry out the actual test??

dlwyatt:
Exactly. Gets you through parameter validation so the mock can run. Mocking both Get-Thingy and Set-Thingy, when [Thingy] is an annoying class to create in the test.

adbertram:
Here's an example.

https://gist.github.com/adbertram/dcb6b86c817473f9b62b878c91c54c41

Get-TfsPendingChange outputs an object of type `Microsoft.TeamFoundation.VersionControl.Client.PendingChange`.

Undo-TfsPendingChange has a PendingChange parameter that is strongly typed to that object. I can't easily instantiate that type. Using this method, I can just run New-MockObject inside of that mock to replicate the type that Get-TfsPendingChange returns.

adbertram:
This way, I don't have to think about it. I just figure out the type the function should output, pass that type to New-MockObject and the function runs as I'd expect.

adbertram:
I'm not sure what's going on with the test now. It looks good to me.

nohwnd:
Okay now it makes sense. Both the Get-* and Set-* functions already exist as some external dependencies, and we mock them. We test another function that is in the middle between those two or some other logic.

adbertram:
Yep. This method has saved me a ton of time by not having to go digging into constructors, figuring out arguments, etc just to get a simple object.

nohwnd:
I just realised we told you that the function will be loaded automatically which is true, but we forgot to tell you to publish it from the module. And that is likely why it fails. So add it to the manifest.

adbertram:
I give up on the tests.

nohwnd:
You added that to the manifest, but not to the .psm1. If you'd run the tests locally you'd see it still fails.

adbertram:
How do I run the tests locally?

nohwnd:
You go to the location where you cloned Pester. Unload the Pester version that loads from your normal module location. Load pester from where you are and then invoke it. 
```powershell
Get-Module Pester | Remove-Module 
Import-Module .\Pester.psd1
Invoke-Pester
```

adbertram:
Should I have added the function to the module all along instead of creating a separate PS1 in Functions?

nohwnd:
no that's not what I meant, you have to specify the command to export in the bottom of the psm1 file where the `Export-ModuleMember` is. You should keep it in separate file.
image

adbertram:
OK, I'm adding all of this into a Contributing to Pester page on the Wiki.

nohwnd:
Another way to run the tests is `.\bin\pester.bat`. That has the advantage of running in clean powershell session. And there are two tests that the CI server ignores by default, so those can and likely will fail in your local run. You'd only need to care about them when releasing a new version. Which you won't do as an external contributor anyway.

image

adbertram:
Feel free to correct whatever I've got wrong on this Wiki page. https://github.com/pester/Pester/wiki/Contributing-to-Pester

nohwnd:
Let's move the discussion about the contribution guide to https://github.com/pester/Pester/issues/636 I will hopefully add more info to the wiki article over the weekend.

adbertram:
I'm really not sure what's going on with the tests now. I've ran the tests locally and it works fine.

adbertram:
Doh! Nevermind. I forgot to check in my changes.

nohwnd:
Good, now it's passing. So now you would ideally rebase on upstream master, squash your commits to one and force push. But I will do that for you when merging.

nohwnd:
Squashed and merged as b81acba294d09d775a3b6f5fd12a82ca90b9109a

PR #643, opened 2016-11-17
Added Get-OperatingSystem and Get-TempDirectory funcs

adbertram:
Fix for TestDrive on PSCore and ability to better differentiate between OSes

### DIFF
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -3,7 +3,7 @@ Set-StrictMode -Version Latest
 if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows)
 {
     $tempPath = $env:TEMP
-} 
+}
 else
 {
     $tempPath = '/tmp'
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -1,6 +1,13 @@
 Set-StrictMode -Version Latest
 
-$tempPath = Get-TempDirectory
+if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows)
+{
+    $tempPath = $env:TEMP
+} 
+else
+{
+    $tempPath = '/tmp'
+}
 
 Describe "Setup" {
     It "returns a location that is in a temp area" {
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -1,6 +1,6 @@
 Set-StrictMode -Version Latest
 
-$tempPath = (Get-Item $env:temp).FullName
+$tempPath = Get-TempDirectory
 
 Describe "Setup" {
     It "returns a location that is in a temp area" {
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -666,7 +666,7 @@ function Get-OperatingSystem
 
     if (Get-Command -Name 'uname' -ErrorAction Ignore)
     {
-        if ((uname) -eq 'Darwin')
+        if ((Invoke-Uname) -eq 'Darwin')
         {
             'MacOS'
         }
@@ -679,6 +679,11 @@ function Get-OperatingSystem
     {
         'Windows'
     }
+}
+
+function Invoke-Uname {
+    param()
+    uname
 }
 
 function Get-TempDiretory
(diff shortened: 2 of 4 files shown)


Pester.psm1
```diff
@@ -659,6 +659,48 @@ function Get-ScriptBlockScope
     [scriptblock].GetProperty('SessionStateInternal', $flags).GetValue($ScriptBlock, $null)
 }
 
+function Get-OperatingSystem
```
adbertram: I got that updated. Do you know why the tests are failing?

nohwnd: There is a lot of tests failing. Are the tests working locally?

Pester.psm1
```diff
@@ -659,6 +659,40 @@ function Get-ScriptBlockScope
...
+    {
+        'OSX'
+    }
+    elseif ($IsLinux)
+    {
+        'Linux'
+    }
+}
+
+function Get-TempDiretory
```
dlwyatt: Typo here, Diretory instead of Directory.  That's the cause of most of the failing tests.

Pester.psm1
```diff
@@ -659,6 +659,40 @@ function Get-ScriptBlockScope
     [scriptblock].GetProperty('SessionStateInternal', $flags).GetValue($ScriptBlock, $null)
 }
 
+function Get-OperatingSystem
+{
+    [CmdletBinding()]
+    param()
+
+    if ($PSVersionTable.PSVersion.Major -lt 6)
```
dlwyatt: This should be `if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows)`

adbertram:
Yay! Tests pass! Wadda ya to merging this now? I'm trying to overcome some of the issues plaguing PS v6 so I can run unit tests on my Mac.

nohwnd:
As with the previous one we have two options, you squash your commits yourself or I'll do it for you. :) In this case the second option is probably better, because there are already reviews for your commits and I think that GitHub hangs on those commits even after you force push, and you end up with dangling commits.

So hang on, I will merge it.

Functions/TestDrive.Tests.ps1
```diff
@@ -1,6 +1,13 @@
 Set-StrictMode -Version Latest
 
-$tempPath = (Get-Item $env:temp).FullName
+if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows)
```
nohwnd: You are repeating the implementation of Get-TempDirectory here. Use the function directly instead.

Pester.psm1
nohwnd: This function uses magic strings, and we don't know yet that we will need to differentiate among multiple operating systems. At the moment this function is only used to distinguish Windows from non Windows systems so I suggest you make it `Test-Windows` function that will return Boolean (better name? IsWindows?).

Pester.psm1
```diff
@@ -659,6 +659,40 @@ function Get-ScriptBlockScope
     [scriptblock].GetProperty('SessionStateInternal', $flags).GetValue($ScriptBlock, $null)
 }
 
+function Get-OperatingSystem
+{
+    [CmdletBinding()]
+    param()
+
+    if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows)
```
nohwnd: A comment explaining why we do the check this way would be warranted here. If you don't know that PowerShell was windows only up until version 6 it might be a bit confusing.

Pester.psm1
adbertram: I chose to create a Get-OperatingSystem function to lay the groundwork for other OSes. With a single Get-OperatingSystem function, you can add and remove OSes at will and bundle up all the logic inside there instead of creating different Test-OS functions.

Also, PS core works on MacOS and Linux and it's quite possible that the temp location (that this function was created for) will change amongst MacOS or different Linux distros.

oivindoh:
Has this PR been forgotten?

Looks like it should now have be 👍 to go!

nohwnd:
Thx for the reminder. I'll try to meege and release new version today.

nohwnd:
Sqashed and merged as https://github.com/pester/Pester/commit/770dec4b73adc6a4620a99be23f5d48cdfa81379

PR #652, opened 2016-12-02
Tests for comment based help added

it-praktyk:
The related issue #648

Issue #648, opened 2016-11-20
Should we add tests that validate help?


nohwnd:
At the moment I am torn between liking the idea of being sure the help is there, and getting test failures for not having help. At the moment we have a group of tests for these non essential requirements. The whitespace tests. 

When these are failing it has no impact on how Pester is working in PowerShell. It's only stylistic rules that should be kept in the repository. And the help falls in pretty much the same category, a non-functional requirement. It would be much better to have it as a some kind of warning that is raised on demand.

You can see the situation with the versioning tests, those should be excluded from the test base because they will fail for anyone who is not currently releasing the Pester version. 

Imho we need something like test traits (XUnit) where we can specify a category of tests (or more generally a trait a test has) and build scenarios around that. (or you know just do it in place with IFs :D )

```powershell
it "Should run during version release" -Traits @{Category = 'Version release'} {} 
it "Should validate help" -Traits @{Category='Version release'; Environment = 'Production'} {}
it "should find correct temp on linux" -Traits @{ Os = "Linux", Version = "6+" } {}
```
@it-praktyk let's discuss this here.

it-praktyk:
The summary of current state of helps - for functions exported by module

The functions without 'Description'
- amount: 11
- functions: Context, Describe, Get-MockDynamicParameters, Get-TestDriveItem, Invoke-Mock, New-MockObject, New-PesterOption, SafeGetCommand, Set-DynamicParameterVariables, Set-TestInconclusive, Setup, 

The functions without described parameters (at least one missed)
- amount: 7 functions, 26 missed parameters
- function: Get-MockDynamicParameters (5), Get-TestDriveItem (1), Invoke-Mock (8), Invoke-Pester (3), Set-DynamicParameterVariables (3), Set-TestInconclusive (1), Setup (5)

The functions without any example 
- amount: 14
- function: AfterAll, AfterEach, BeforeAll, BeforeEach, Get-MockDynamicParameters,  Get-TestDriveItem, In, Invoke-Mock, New-PesterOption, SafeGetCommand, Set-DynamicParameterVariables, Set-TestInconclusive, Setup, Should

Results calculated based on the tests what you can find [here](https://github.com/it-praktyk/Pester/blob/pr/647/Help.Tests.ps1) - code equal with the pull request #647 .

dlwyatt:
[...]

PR #667, opened 2016-12-24
Tests for comment based help added

it-praktyk:
The related issue #648
Replaces the pull request #652

### DIFF
--- a/Pester.Tests.ps1
+++ b/Pester.Tests.ps1
@@ -285,3 +285,119 @@ InModuleScope Pester {
         }
     }
 }
+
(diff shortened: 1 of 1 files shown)


it-praktyk:
Please wait with approve for that pull request until all functions will not have updated help. I'll try do it soon.

nohwnd:
I will, I don't want to merge this without having a good strategy to allow people to post code without help. Not everybody is help-writing kind, and running tens of non-functional tests locally on every change is not fun. Probably those tests should become part of some pre-release workflow, but for that we need a way to do default exclude, or at least some guide on how to run our tests. But thanks for the PR and food for thought :)

it-praktyk:
I close the current pull request. I'll replace it with another one soon.

PR #678, opened 2017-01-13
Add changelog to Release Notes for discoverability

juneb:


nohwnd:
I wanted to add this, when you mentioned it on Twitter, but the description explicitly says that it should be:
>Release notes for this particular version of the module

Is this going to print the whole file? I assume it will. Could you test please and maybe post some screens? To meet the requirements of the module manifest we need to do some post processing on the changelog and manifest, and need tests for that. But it should not be difficult, it's just splitting on second  `##`, and cleaning up a bit.

nohwnd:
Moving this to 4.1, the release notes should be different from changelog (as discussed with alx9r), and we need to come up with ways of publishing them in an unified way.

PR #685, opened 2017-01-19
Added basic help for Should.

it-praktyk:


nohwnd:
Thanks, but unless the help is dramatically different it's okay if you just PR against master, I am forward merging the changes to the V4 anyway.

it-praktyk:
I pushed pull requests twice because since yesterday I didn't know when v. 4 branch will be published. 

I hope that Pester v.4 will be release in Debian style: "[As always, Debian 9 "Stretch" will be released "when it's ready".](https://release.debian.org/) so before that maybe the next v.3.x will be released too.

nohwnd:
Gotcha, but that is not how it works. All the stuff that goes into Pester v3 is merged to v4 as soon as possible, when you add help to Should in v3, I will eventually merge that into v4 as well. You don't have to duplicate the PRs. Unless you specifically target only v4, making PR for v3 is enough.

it-praktyk:
OK, I got it now.

PR #715, opened 2017-01-30
Update Gherkin for Pester 4 Changes

### LINKED ISSUE #714: [Pester 4.0.2] Gherkin scenario counts are broken
The changes to how Describe and Context are handled in Pester 4 broke the output for Gherkin by flipping the values of "Context" and "Describe" in the `$Pester.TestResult` ...

I need to investigate further -- just wanted to log this here so we can be sure to fix it before release.

Any thoughts about how to include tests that fail so that I can test the result of failing tests ... without ... failing tests? (Wow, that's confusing).

Jaykul:
This should fail -- right now. 
Once it's fixed, it will fix #714 and then you can merge it.

### DIFF
--- a/Examples/Gherkin/Gherkin-Scope.feature
+++ b/Examples/Gherkin/Gherkin-Scope.feature
@@ -1,0 +1,31 @@
+Feature: Pester's Gherkin implementation test scope
+    As a PowerShell test author
+    I want variables to work across step implementations
+    So that writing tests will be simple
+
+    Scenario: Scope Should Be Easy Part 0
+        Given I initialize the variable One to "Uno"
+        When I set the variable One to "Hello World"
+        Then the variable One should be "Hello World"
+
+    Scenario: Scope Should Still Be Easy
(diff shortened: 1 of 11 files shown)


Jaykul:
@nohwnd can you explain what this failure is? Did I actually break PowerShell 2, or is the TeamCity script broken? It's claiming there's a syntax error with an = missing, but that means nothing to me. If I actually broke the build on PS2, I can find a VM to test on and fix it, but ...

nohwnd:
@jaykul sorry I missed your comment, will have a look soon.

nohwnd:
@Jaykul In PowerShell v2 you had to use `=$true` in parameter attributes. So I guess [this line](https://github.com/pester/Pester/pull/715/commits/7e25ba3283801b3847edf8753bc4277df71b2f75#diff-d95f2f48299562a597fc7ad45650b50dR59) is the cause.

Jaykul:
I'll fix these then.

Jaykul:
This is done, @nohwnd -- but needs to be followed with #726

PR #735, opened 2017-02-27
Enable PesterThrow to check FullyQualifiedErrorId before checking exc…

JamesWTruher:
…eption message

FullyQualifiedErrorId is not localized which enables `should throw` tests to be more
portable when working with multiple languages. Exception message is still checked, but
only after FullyQualifiedErrorId is validated.

### DIFF
--- a/Functions/Assertions/PesterThrow.Tests.ps1
+++ b/Functions/Assertions/PesterThrow.Tests.ps1
@@ -15,6 +15,11 @@ InModuleScope Pester {
             Test-PositiveAssertion (PesterThrow { throw $expectedErrorMessage } $expectedErrorMessage)
         }
 
+        It "returns true if the statement throws an exception and the FullyQualifiedErrorId matches the expected error text" {
+            $expectedErrorMessage = "PathNotFound,Microsoft.PowerShell.Commands.GetItemCommand"
+            Test-PositiveAssertion (PesterThrow { get-item -ea stop "TESTDRIVE:/filedoesnotexist.txt" } $expectedErrorMessage)
+        }
+
         It "returns false if the statement throws an exception and the actual error does not match the expected error text" {
             $unexpectedErrorMessage = "unexpected error message"
             $expectedErrorMessage = "some expected error message"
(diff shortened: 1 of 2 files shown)


JamesWTruher:
comments? anyone?

nohwnd:
I'd love to merge it but I can't spend any time on Pester these days unfortunately :/

JamesWTruher:
is there anyone else that can merge this?

dlwyatt:
Should we put this into the v4 branch instead?  That way we can separate FQID checking from Message checking, instead of having both pieces of logic in the same parameter.

JamesWTruher:
I'm happy to alter it - I had thought of a couple of different implementations, e.g., as a new assertion `ThrowFQID` but I thought this was still in keeping with what the intent of the `throw` assertion, just doing an extra, more rigorous, check first. Let me know what you would like.  

Another implementation thought was as a second *optional* parameter, but I didn't like this at all as none of the assertions require the use of a parameter name.

dlwyatt:
In pester v3, none of the assertions had the _possibility_ of using a parameter name; `Should` worked on automatic positional `$args`.  In the v4 code, we have a lot more flexibility through dynamic parameter sets.  I sort of lean toward preferring `Should -Throw -FQID 'Whatever'`, personally.

nohwnd:
In my opinion we should implement the common use cases as the parameters to should throw (I already have the implementation in a branch somewhere), and we should also return the exception from the should (it will be discarded by default because all the output goes to null), but you can also capture it and add as many assertions as you want. The same concept is useful for Be as well, and it would be even more useful if we added, script callback that would allow us to point the assertion to just one property of the object, while still passing the whole object on. Or we could validate the whole object against another object to avoid assertion roulette, which is something I am implementing in my Assertion module :)
I will post more examples soon :)

Jakub Jareš

.NET, Pester, PowerShell
t @nohwnd

9. 3. 2017 v 2:20, Dave Wyatt <notifications@github.com<mailto:notifications@github.com>>:


In pester v3, none of the assertions had the possibility of using a parameter name; Should worked on automatic positional $args. In the v4 code, we have a lot more flexibility through dynamic parameter sets. I sort of lean toward preferring Should -Throw -FQID 'Whatever', personally.

JamesWTruher:
ok - I'll create a PR for v4 along those lines. 
Is there any possibility for a v3 solution? Would a new assertion be reasonable?
My desire is to use this in https://github.com/PowerShell/PowerShell and while we have our own fork (we need cross-platform), I would rather reduce divergence rather than widen it.

nohwnd:
@JamesWTruher  This functionality seems to be in already (not sure if it was there since 7 months ago as the files say, or if Dave just merged some olde commit of his ), but you can do: 
```powershell
Describe "d" { 
    It "it will fail because errorId constraint is not met" { 
        {  Get-Command csdf -ErrorAction stop } | Should -Throw -Errorid SomeOtherException 
    }

    It "it will pass because errorId constraint is met" {  
        {  Get-Command csdf -ErrorAction stop } | Should -Throw -Errorid CommandNotFoundException 
    }
}
```

nohwnd:
@JamesWTruher At the moment I am working on making Pester work on PS6 (I am using the microsoft/powershell docker image). 

I am closing this PR, please open an issue if you'd like to discuss more. :)

PR #743, opened 2017-03-19
Fix #697

### LINKED ISSUE #697: Issue with Pester v4.0.2 Mocking a Dummy Function with a Pipeline in two consequitive Context Blocks
Hi Guys,

It appears that Pester 4.0.2 has broken our unit tests in the [xStorage DSC Resource module](https://github.com/PowerShell/xStorage).

Basically, what is happening is that we are creating a [dummy function of Get-Volume](https://github.com/PowerShell/xStorage/blob/dev/Tests/Unit/MSFT_xDiskAssignPath.tests.ps1#L333) that accepts a pipeline parameter (this is so that it can accept a mocked partition CIM object that isn't a real CIM class - as is standard for this type of test).

The [first context that Mocks](https://github.com/PowerShell/xStorage/blob/dev/Tests/Unit/MSFT_xDiskAssignPath.tests.ps1#L398) this function [works perfectly](https://ci.appveyor.com/project/PlagueHO/xstorage/build/2.9.68.0#L293).

The [second context that Mocks](https://github.com/PowerShell/xStorage/blob/dev/Tests/Unit/MSFT_xDiskAssignPath.tests.ps1#L462) this function [fails to mock correctly](https://ci.appveyor.com/project/PlagueHO/xstorage/build/2.9.68.0#L310) : the real Get-Volume is being called - generating the error because we're passing in a PSObject via the pipeline, not a real CIM Partition object.

If I use Pester 3.4.6 everything works perfectly.

Any help would be greatly appreciated.

dlwyatt:
In the past we've been renaming functions back and forth when mocking an existing function.  Apparently we've been lucky that this worked at all, because when we rename the function back, we can wind up modifying its scope.

This PR changes the approach:  we never modify the original command, even if it's a function.  Instead, we create an alias with the same name (which takes precedence anyway), and point that alias to our mock's bootstrap function that can exist under any name.  (Using GUIDs to ensure uniqueness.)  When we delete the alias, the original unmocked version of the command goes back to being in effect without any tampering on Pester's part.


dlwyatt:
Set the PR to point to master by mistake, re-opening for DevelopmentV4

PR #750, opened 2017-03-21
Use OS independent path for resolve test

nohwnd:
Used path of powershell instead of hardcoded path to notepad.exe to make it compatible with Linux.


nohwnd:
The broken check is from Appveyor that I tried to setup for Linux. The TC build passed.

Issue #708, opened 2017-01-26
[Pester 4.0.2] Localization error


megamorf:
Hey guys,

we use Pester on German Windows Servers (2012R2)

```powershell
PS ctx-tests:\> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.14409.1005
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14409.1005
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

When we run our tests each Pester invocation is followed by a Pester error:

```
Import-LocalizedData : Die Windows PowerShell-Datendatei "RSpec.psd1" kann nicht im Verzeichnis "C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.2\de-DE\" oder einem übergeordneten 
Verzeichnis gefunden werden.
In C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.2\Pester.psm1:674 Zeichen:9
+         Import-LocalizedData -BindingVariable Script:ReportStrings -B ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Program File...e-DE\RSpec.psd1:String) [Import-LocalizedData], PSInvalidOperationException
    + FullyQualifiedErrorId : ImportLocalizedData,Microsoft.PowerShell.Commands.ImportLocalizedData
```

Which basically means that it cannot find 'RSpec.psd1' in 'C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.2\de-DE\' or the directory above. I guess users of other non-English OS' are going to run into the same error message sooner or later.

it-praktyk:
I added fallback to en-US.
The corrected code is available in my branch
https://github.com/it-praktyk/pester/tree/DevV4_Import-LocalizedData_Fallback

Could you test?

megamorf:
yup, that solves the issue 👍

megamorf:
Can this be merged into master please and be released. As it is now all our machines that pull the latest version of pester are affected by this bug and throw the `RSpec.psd1`-error.

nohwnd:
@megamorf fixed and pushed in 4.0.3.

PR #755, opened 2017-03-22
Swap Describe and Context and order correctly with \ as a separator

### LINKED ISSUE #751: Content of Describe and Context elements in PassThru object are swapped
The values for the Context and Describe NoteProperties of the TestResult objects are swapped.

For Example:
```powershell
PS> 'Describe "This should be in the Describe NoteProperty" { context "This should be in the Context NoteProperty" { it "whatever" { $true | Should Be $true } } }' | Out-File -FilePath .\test.ps1
PS> $result = Invoke-Pester -Script .\test.ps1 -PassThru
PS> $result.TestResult

ErrorRecord            :
ParameterizedSuiteName :
Describe               : This should be in the Context NoteProperty
Parameters             : {}
Passed                 : True
Show                   : All
FailureMessage         :
Time                   : 00:00:00.0643275
Name                   : whatever
Result                 : Passed
Context                : This should be in the Describe NoteProperty
StackTrace             :
````

(Found in 4.0.2)

### PR DESCRIPTION
**Pull request by nohwnd**
Fixed the placement, changed the separator to be `\` which seemed more natural. 

```powershell
'Describe "top" { describe "middle" { context "bottom" { it "whatever" { $true | Should Be $true } } } }' | Out-File -FilePath .\test.ps1
$result = Invoke-Pester -Script .\test.ps1 -PassThru
$result.TestResult
```
```
ErrorRecord            : 
ParameterizedSuiteName : 
Describe               : top
Parameters             : {}
Passed                 : True
Show                   : All
FailureMessage         : 
Time                   : 00:00:00.0974222
Name                   : whatever
Result                 : Passed
Context                : middle\bottom
StackTrace             : 
```

Fixes #751

### DIFF
--- a/Functions/PesterState.ps1
+++ b/Functions/PesterState.ps1
@@ -182,7 +182,11 @@ function New-PesterState
             $describe = ''
             $contexts = [System.Collections.ArrayList]@()
 
-            foreach ($group in $script:TestGroupStack.GetEnumerator())
+            # make a copy of the stack and reverse it
+            $reversedStack = $script:TestGroupStack.ToArray()
+            [array]::Reverse($reversedStack)
+
+            foreach ($group in $reversedStack)
             {
                 if ($group.Hint -eq 'Root' -or $group.Hint -eq 'Script') { continue }
                 if ($describe -eq '')
@@ -195,7 +199,7 @@ function New-PesterState
                 }
             }
 
-            $context = $contexts -join '/'
+            $context = $contexts -join '\'
 
             $script:TestResult += & $SafeCommands['New-Object'] psobject -Property @{
                 Describe               = $describe


nohwnd:
@dlwyatt pls review if that is how you meant it to work.

nohwnd:
It's missing tests, but only because I'd like to release 4.0.3 asap to push out the culture fallback fix that blocks megamorf in #708

dlwyatt:
The bug was that I assumed that the loop would enumerate from the root down to the It block, but since it's a stack, it goes the other way.  The easiest fix would be to use ToArray() instead of GetEnumerator() (and to call Array::Reverse if needed.)

The intention was to have existing scripts get the same output they always had, and to support _something_ similar for newer test scripts with deeper nesting (when the calling code still accesses TestResult instead of the newer TestActions property); the best way to do that seemed to be to just append to Context.

nohwnd:
I also considered doing array reverse, but the code I have also works even though it might be less optimal (not that it mattered in this case). 

Are you fine with having the path in Describe or should I redo it to have the path in Context?

dlwyatt:
I think having the extras in Context makes more sense, since then it follows the same path down as the test scripts.

nohwnd:
Ok, will change that.

PR #764, opened 2017-03-29
[V3Maintenance] Backport from DevV4 - mostly Join-Path added

it-praktyk:
A purpose of the current commit is easier maintenance between branches - most of changes where introduced by 5dae41c14c1bff5bae08b8e39446d437fc0c1747


it-praktyk:
I assume that due to expected Pester 3.4.x deprecation the current pull request can be closed.

PR #765, opened 2017-03-30
Make Add-AssertionOperator a public function

cdhunt:
Allow other modules to extend Pester by making `Add-AssertionOperator` public.

### DIFF
--- a/Pester.psd1
+++ b/Pester.psd1
@@ -50,6 +50,7 @@ FunctionsToExport = @(
     'SafeGetCommand'
     'New-PesterOption'
 	'New-MockObject'
+    'Add-AssertionOperator'
 
     # Gherkin Support:
     'Invoke-Gherkin'
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -140,7 +140,7 @@ function Assert-ValidAssertionAlias
     }
 }
 
-#<#
+<#
 .SYNOPSIS
     Register an Assertion Operator with Pester
 .DESCRIPTION
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -140,6 +140,51 @@ function Assert-ValidAssertionAlias
     }
 }
 
+#<#
+.SYNOPSIS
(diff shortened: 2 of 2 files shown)


dlwyatt:
Please target the `master` branch from now on; I merged DevelopmentV4 in about a week ago.

Also, we'll need to write some documentation on this before making it public in a release.  (Planning to do that this weekend anyway.)

nohwnd:
So do we need to keep the development v4 branch around or it can be deleted?

dlwyatt:
Thought I'd already done that.  My mistake, it's gone now.  :)

cdhunt:
Should the function be moved out of the PSM1 and into a separate file under `Functions\`?

dlwyatt:
No; it needs to exist before the other ps1 files get run.

cdhunt:
I'm having problems actually utilizing the function. I am passing in an Object as `$ActualValue` and pulling out and testing a property. It seems like `$ActualValue` is still being passed to my Message functions instead of the property value.

https://gist.github.com/cdhunt/018281748ca316803bae67671a69fd72
```
Expected SizeRemaining {MSFT_Volume (ObjectId = "{1}\\USHOLCHU1-L\root/Microsoft/Windows...)} to be greater than {52428800}
```

Also, it appears just the value of $failureMessage is being passed to `$testResult` instead of an object with `Succeeded` and `FailureMessage` properties. I can't figure that one out because when I step through the code it never seems to enter my `ArchitestVolumeHaveSizeRemaining` function.

cdhunt:
Ok, all of the above mentioned failures were just due to crisscrossed function names from bad copy & paste.

it-praktyk:
A comment based help doesn't exist for Add-AssertionOperator, wiki is not updated too.

cdhunt:
Created a wiki page to start.

https://github.com/pester/Pester/wiki/Add-AssertionOperator

it-praktyk:
@cdhunt, thank you!

PR #707, opened 2017-01-25
[WIP] Clean of code - PSScriptAnalzyer rules: PSAvoidUsingCmdletAliases, PSAvoidDefaultValueForMandatoryParameter

it-praktyk:
The related issue #700

Issue #700, opened 2017-01-22
Add checks of basic PSScriptAnalyzer rules


nohwnd:
Add PSScriptAnalyzer analyzer rules to be used as a part of pre-release checks:
- PSAvoidUsingCmdletAliases
- PSUseDeclaredVarsMoreThanAssignments
- PSUseOutputTypeCorrectly
- PSAvoidDefaultValueForMandatoryParameter
- PSUseCmdletCorrectly

reported by @it-praktyk

it-praktyk:
You can assign it to me.

it-praktyk:
@nohwnd , please change the title of the issue to: "Add checks of basic PSScriptAnalyzer rules"

The current results of added tests
```powershell
Describing PSScriptAnalyzer rules
   [-] Pester source code doesn't violate the PSScriptAnalzyer rule PSAvoidUsingCmdletAliases 1.44s
     The following 52 lines violate PSScriptAnalyzer rule PSAvoidUsingCmdletAliases:

     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\build.psake.ps1, Line: 26
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\build.psake.ps1, Line: 27
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Pester.psm1, Line: 175
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Pester.psm1, Line: 190
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Assertions\Should.Tests.ps1, Line: 112

     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Gherkin.ps1, Line: 182
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Gherkin.ps1, Line: 182
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Gherkin.ps1, Line: 182
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Gherkin.ps1, Line: 183
     File: C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Functions\Gherkin.ps1, Line: 186
[...]

PR #701, opened 2017-01-22
Tests for selected PSScriptAnalyzer rules added

it-praktyk:
Rules covered by added tests
- PSAvoidUsingCmdletAliases
- PSUseDeclaredVarsMoreThanAssignments
- PSUseOutputTypeCorrectly
- PSAvoidDefaultValueForMandatoryParameter
- PSUseCmdletCorrectly

The related issue #700

PR #766, opened 2017-03-30
Cleaning of code - PSScriptAnalyzer rules

it-praktyk:
Violations reported by PSScriptAnalyzer v.1.11.0

The related issue #700 
Replace the pull request #707

### DIFF
--- a/Functions/Assertions/Should.Tests.ps1
+++ b/Functions/Assertions/Should.Tests.ps1
@@ -109,7 +109,7 @@ InModuleScope Pester {
         It "ensures all assertion functions provide failure messages" {
             $assertionFunctions = @("PesterBe", "PesterThrow", "PesterBeNullOrEmpty", "PesterExist",
                 "PesterMatch", "PesterContain")
-            $assertionFunctions | % {
+            $assertionFunctions | ForEach-Object -Process {
                 "function:$($_)FailureMessage" | Should Exist
                 "function:Not$($_)FailureMessage" | Should Exist
                 "function:$($_)FailureMessage" | Should -Exist
--- a/Functions/Assertions/Should.Tests.ps1
+++ b/Functions/Assertions/Should.Tests.ps1
@@ -109,7 +109,7 @@ InModuleScope Pester {
         It "ensures all assertion functions provide failure messages" {
             $assertionFunctions = @("PesterBe", "PesterThrow", "PesterBeNullOrEmpty", "PesterExist",
                 "PesterMatch", "PesterContain")
-            $assertionFunctions | ForEach-Object -Process {
+            $assertionFunctions | ForEach-Object {
                 "function:$($_)FailureMessage" | Should Exist
                 "function:Not$($_)FailureMessage" | Should Exist
                 "function:$($_)FailureMessage" | Should -Exist
--- a/Functions/GherkinHook.Tests.ps1
+++ b/Functions/GherkinHook.Tests.ps1
@@ -103,11 +103,11 @@ Describe 'Testing Gerkin Hook' {
         & ( Get-Module Pester ) {
             BeforeEachScenario "I Click" { }
             $GherkinHooks["BeforeEachScenario"].Tags
-        } | Select -Last 1 | Should Be "I Click"
+        } | Select-Object -Last 1 | Should Be "I Click"
 
         & ( Get-Module Pester ) {
             AfterEachFeature "I Click" { }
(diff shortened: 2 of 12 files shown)


dlwyatt:
I really don't care what script analyzer thinks. I prefer the aliases "select", "where", etc. However, I do think we should get rid of the instances of % and ?

I'm also fine with positional parameters when the meaning is obvious, regardless of the whiny linting tool. :)

it-praktyk:
I hope that your personal preferences will not be block of merge that PR ;-)

By mistake I left proposed - under #701 - tests and the current results you can see on the CI results page [here](https://build.powershell.org/viewLog.html?buildId=10199&buildTypeId=Pester_TestPester&guest=1).

What about these?

W. alias Whiny ;-)

dlwyatt:
I won't be merging this as-is.  `ForEach-Object -Process` just makes me want to scream.  :P

it-praktyk:
Under verifying other rules used by PSScriptAnalyzer I found real bug :-/

2017-03-31 21_43_51-pester tests ps1 head - pester tests ps1 - pester - visual studio code

it-praktyk:
@Jaykul , could you check ? The partial result of test below
```
Pester source code doesn't violate the PSScriptAnalzyer rule PSUseDeclaredVarsMoreThanAssignments
      File: C:\Users\<USERNAME>\Pester\Functions\Gherkin.ps1, Line: 4
      File: C:\Users\<USERNAME>\Pester\Functions\Gherkin.ps1, Line: 339
```
Thank you.

it-praktyk:
The related issue #768

it-praktyk:
Guys, I did that work because I think that cleaner code is a better code.

As you can observer I closed my previous pull request #701 because the PSScriptAnalyzer rules can't be used automaticaly for the code of Pester project.

I hope that you will merge it even if you don't agree with some principles of cleaner coding.

W. alias Whiny ;-)

it-praktyk:
The CI validation for the current branch failed due to the bug #768 - please consider merge or rebase a code after reviewing and correcting the mentioned bug.

dlwyatt:
> I think that cleaner code is a better code.

I totally agree with that.  However, I don't think that the script analyzer rules produce cleaner code in all cases:

- The `*-Object` cmdlets are mostly silly names, likely created only because they had agreed to name everything with Verb-Noun (which is still silly, because `Where` and `ForEach` aren't verbs anyway.)

- Parameter names are, in some cases, just noise.  `-Process` on `ForEach-Object` is one example.  When you're only using that block (which is probably 95% of the uses), it's perfectly clear without the parameter name.  If you are passing in multiple blocks, then parameter names are appropriate.  Another common case is where the first positional parameter has a name that exactly matches the noun of the cmdlet:  `Test-Path -Path $somePath` is silly, when `Test-Path $somePath` is still just as clear.  In the case of `Join-Path`, I go even farther and use both arguments by position:  `Join-Path $parent $child` is already easy to read.

- I also use positional parameters when calling `New-Object`, mostly because it just feels right after typing `new TypeName(constructorParameters)` in C-Style languages for so long.  Right or wrong, that's just my preference, and you'll see the same thing all over the place (even in Microsoft's own code.)  I prefer the look of `New-Object TypeName($params)`, even though the syntax is kind of abusing the PowerShell parser a bit.

I'm fine with making some of these changes to clean up the code, but not all of them, and I don't expect Pester's code to ever match the stock script analyzer rules 100%.

Jaykul:
I'll check the line mentioned, that looks bad.
But I agree about parameters and aliases on the -Object commands, those "rules" are silly.

it-praktyk:
I just removed removed named parameters.

dlwyatt:
Getting there :)  I'll strip out the -Object names and merge it later tonight.

it-praktyk:
So, generally will you revert big part of the commit 89022f6b8db263206fecbe553af5a37dccdafe5f ? Even if I removed names of parameters introduced by it?

it-praktyk:
@Jaykul , did you check https://github.com/pester/Pester/pull/766#issuecomment-290829140 ?

Jaykul:
@it-praktyk I don't remember. I'll try to look tonight...

Jaykul:
@it-praktyk finally checked. Both those lined can just be removed. See #801

it-praktyk:
@Jaykul, thank you :-)

it-praktyk:
Thank you

PR #779, opened 2017-05-17
Renamed common function name to prevent conflict with local scope

adbertram:
I confirmed that if the user has a module with a function name of Get-OperatingSystem, the Environment script will attempt to use that one. Renamed functions in this script to prevent conflict in the future.


it-praktyk:
@adbertram could you rename the pull request to something more descriptive? Thx.

nohwnd:
@adbertram  I cannot replicate this, could you provide steps to replicate the issue please?

This is what I tried: 
```powershell
# remove all currently imported Pester modules
Get-Module Pester | Remove-Module -Force

# import pester from repo 
cd C:\projects\pester_nohwnd
Import-Module .\Pester.psd1 

# create a module that exports the conflicting function
# and import the module
New-Module -ScriptBlock { 
        function Get-OperatingSystem { throw "asdb" } 
    } -Name osUtils | Import-Module

# Get-OperatingSystem # <- this throws if enabled

Invoke-Pester # <- this works fine, and does not pick-up the function
```

If the conflict exists, then a better solution might be using module qualified function name, but that depends on the cases where the conflict occurs.

dlwyatt:
No harm in merging this, it's not a public function anyway.

PR #786, opened 2017-06-04
Fix Assert-VerifiableMocks silently does not report anything if 'Verifiable' is spelled wrong on one mock

### LINKED ISSUE #781: Assert-VerifiableMocks silently does not report anything if 'Verifiable' is spelled wrong on one mock
This code block reports the correct error.

```
Describe 'Tests' {
    Mock Get-Service -MockWith { return 'wuausrv' } -Verifiable
    Mock Get-Process -MockWith { return 'setup' } -Verifiable

    Context 'When something happens' {
        It 'Should return correct values' {
            Get-Service | Should Be 'wuausrv'
            #Get-Process | Should Be 'setup'
        }
    }

    Assert-VerifiableMocks
}
```

Assert-VerifiableMocks throw the correct expected error

```
Describing Tests

  Context When something happens
    [+] Should return correct values 124ms
  [-] Error occurred in Describe block 27ms

     Expected Get-Process to be called with $True
    At C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.3\Functions\Mock.ps1:428 char:9
```

But if we make a typo in the parameter -Verifiable for the mock `Get-Process`

```
Describe 'Tests' {
    Mock Get-Service -MockWith { return 'wuausrv' } -Verifiable
    Mock Get-Process -MockWith { return 'setup' } -Verfiable

    Context 'When something happens' {
        It 'Should return correct values' {
            Get-Service | Should Be 'wuausrv'
            #Get-Process | Should Be 'setup'
        }
    }

    Assert-VerifiableMocks
}
```

Assert-VerifiableMocks reports success

```
Describing Tests

  Context When something happens
    [+] Should return correct values 76ms
```

I don't know if there is something you can do to fix this. But wanted to report it just in case. 😄

alx9r:
This PR resolves #781 by doing the following:

* add tests that assert that all non-deprecated, non-internal, public Pester commands use CmdletBinding
* add `[CmdletBinding()]` to the signature of those commands that were missing it
* fix the comment-based help synopsis for `Setup` to make identifying public vs. Pester-internal more consistent

### DIFF
--- a/Functions/Assertions/Set-TestInconclusive.ps1
+++ b/Functions/Assertions/Set-TestInconclusive.ps1
@@ -51,6 +51,7 @@ function Set-TestInconclusive {
     .LINK
     https://github.com/pester/Pester/wiki/Set%E2%80%90TestInconclusive
 #>
+    [CmdletBinding()]
     param (
         [string] $Message
     )
--- a/Functions/In.ps1
+++ b/Functions/In.ps1
@@ -16,11 +16,11 @@ The path that the execute block will be executed in.
 The script to be executed in the path provided.
 
 #>
-
-param(
-    $path,
-    [ScriptBlock] $execute
-)
+    [CmdletBinding()]
+    param(
+        $path,
+        [ScriptBlock] $execute
+    )
     Assert-DescribeInProgress -CommandName In
 
     $old_pwd = $pwd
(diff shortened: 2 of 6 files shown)


alx9r:
Thanks @nohwnd.

PR #798, opened 2017-07-11
Changed all URLs in PrivateData to HTTPS

ChrisUbben:
Using insecure network communications increases the risk for attacks such as Man in the Middle attacks.
As part of good coding practice, we should default to using https in all URLs unless http is specifically needed.

### DIFF
--- a/Pester.psd1
+++ b/Pester.psd1
@@ -111,13 +111,13 @@ PrivateData = @{
         Tags = @('powershell','unit_testing','bdd','tdd','mocking')
 
         # The web address of an icon which can be used in galleries to represent this module
-        IconUri = "http://pesterbdd.com/images/Pester.png"
+        IconUri = "https://pesterbdd.com/images/Pester.png"
 
         # The web address of this module's project or support homepage.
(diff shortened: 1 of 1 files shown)


it-praktyk:
Did you try open https://pesterbdd.com/images/Pester.png? I did.
selection_001

ChrisUbben:
Yes I did, it appears to be a certificate issue from the Pester website.
Are there plans to fix this issue? or should connections to pesterbdd.com only use http?

Pester.psd1
```diff
@@ -111,13 +111,13 @@ PrivateData = @{
         Tags = @('powershell','unit_testing','bdd','tdd','mocking')
 
         # The web address of an icon which can be used in galleries to represent this module
-        IconUri = "http://pesterbdd.com/images/Pester.png"
+        IconUri = "https://pesterbdd.com/images/Pester.png"
```
it-praktyk: @ChrisUbben, please update your pull request - the pesterbdd.com will not be available via https soon.

Issue #121, opened 2014-03-14
Should Contains for evaluating arrays


tribou:
It would be nice if the following were possible:

``` powershell
$array | Should Contain $item
```

Instead of:

``` powershell
$array -contains $item | Should Be $true
```

The former would help if it responded with a more specific failure message for better troubleshooting. (Instead of the latter's "Expected: {True} but was {False}")

Jaykul:
To wake up this thread ... that sounds about right.

pcgeek86:
+1 to OP

PR #804, opened 2017-07-27
Add array and hastable assertions

stuartleeks:
Based on a tweet from @it-praktyk (https://twitter.com/ITpraktyk/status/889585074760552453), I'm submitting a PR with the [array](https://github.com/stuartleeks/pestermatcharray) and [hashtable](https://github.com/stuartleeks/pestermatchhashtable) assertions from my pester extensions. 

I used them in my [posh-HumpCompletion](https://github.com/stuartleeks/posh-HumpCompletion) tests here: https://github.com/stuartleeks/posh-HumpCompletion/blob/master/posh-HumpCompletion.tests.ps1#L64

Let me know what you think, and whether there are any changes you'd recommend :-)


nohwnd:
Hi Stuart,
Thanks for the PR. I like the idea, but I will mark it for 4.1 release so I might take a while till it's reviewed properly and merged. I only had a very quick look and I think we need to brainstorm about the naming a bit till we commit to using Match in the assertion name. This term is normally associated with Regular expressions.
 
It's also a bit limiting that you can provide only a single param to the assertion, but that is how it works for the moment.

nohwnd:
Please also review why it failed to build on the build server. 

https://build.powershell.org/viewLog.html?buildId=10874&tab=buildResultsDiv&buildTypeId=Pester_TestPesterOnPowerShellV2

stuartleeks:
Thanks @nohwnd - I look forward to discussing this more with you when the 4.0 release is out :-)

qq - is the build server triggered automatically? I've updated the code to hopefully fix the error on the build server and pushed it but it doesn't look like it rebuilt

stuartleeks:
@nohwnd  - I've just updated the PR and it looks like the build is passing now: https://build.powershell.org/viewLog.html?buildId=10884&tab=buildResultsDiv&buildTypeId=Pester_TestPesterOnPowerShellV2

it-praktyk:
#121 - is it related?

stuartleeks:
From scanning the thread on that issue, it looks like the OP wants to test for the existence of an element in an array. This PR is intended to allow comparison of the full array or hashtable, so I think that they are slightly different cases.

it-praktyk:
Thx, I just found it in a middle of the night so wasn't sure.

dlwyatt:
In v4, `Should Be` already compares arrays, so we may not need a separate operator for that.

DarkLite1:
Thanks @dlwyatt , I didn't know this. This feature, comparing of `Arrays` is not in the `Help` for the CmdLet `Should` in version `4.0.8`:

```
PS C:\> Get-Help Should

NAME
    Should
    
SYNOPSIS
    Should is a keyword what is used to define an assertion inside It block.
    
    
SYNTAX
    Should [[-LegacyArg1] <Object>] [[-LegacyArg2] <Object>] [[-LegacyArg3] <Object>] [-ActualValue <Object>] [<Common
    Parameters>]
    
    
DESCRIPTION
    Should is a keyword what is used to define an assertion inside the It block.
    Should provides assertion methods for verify assertion e.g. comparing objects.
    If assertion is not met the test fails and an exception is throwed up.
    
    Should can be used more than once in the It block if more than one assertion
    need to be verified. Each Should keywords need to be located in a new line.
    Test will be passed only when all assertion will be met (logical conjuction).
    

RELATED LINKS
    about_Should
    about_Pester 

REMARKS
    To see the examples, type: "get-help Should -examples".
    For more information, type: "get-help Should -detailed".
    For technical information, type: "get-help Should -full".
    For online help, type: "get-help Should -online"



PS C:\> Get-Help Should -Examples

NAME
    Should
    
SYNOPSIS
    Should is a keyword what is used to define an assertion inside It block.
```

it-praktyk:
@DarkLite1, I assume that it was described in the updated version of about_Should. Can you verify?

Help for Should itself is very limited intentionally.

it-praktyk:
The label 'Documentation' added - to remember to add information about `Should -Be` and arrays to 'about_Should_help'

nohwnd:
@stuartleeks Stuart thanks for your contribution, but I won't merge this. I plan to move Assert into Pester and use the functionality developed there to give Pester collection, hashtable, and dictionary comparing assertions. Hope this will give us more options such as ordered, non-ordered, each and any.

Issue #785, opened 2017-06-04
[Pester 4.0.3] Call depth overflow on $xmlObject | Should be $xmlObject


alx9r:
## Repro

```PowerShell
Describe 'overflow' {
    It '[xml] object is itself.' {
        $doc = [xml]''
        $doc | Should be $doc
    }
}
```

## Expected Result

I expected this test to pass.  It passes using Pester 3.4.3.

## Actual Result

```
Describing overflow
  [-] [xml] object is itself. 45.77s
    The script failed due to call depth overflow.
    at <ScriptBlock>, : line 4
    4:         $doc | Should be $doc
```

## Environment

```
PS C:\> $PSVersionTable

Name                           Value                      
----                           -----                      
PSVersion                      5.1.14409.1005             
PSEdition                      Desktop                    
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}    
BuildVersion                   10.0.14409.1005            
CLRVersion                     4.0.30319.42000            
WSManStackVersion              3.0                        
PSRemotingProtocolVersion      2.3                        
SerializationVersion           1.1.0.1                    

PS C:\> Get-Module Pester

ModuleType Version    Name                                
---------- -------    ----                                
Script     4.0.3      Pester                              
```

alx9r:
The problem seems to manifest in [this recursive call in `ArraysAreEqual`](https://github.com/pester/Pester/blob/2541ef9075ad36b842fc598fc10cb6a258ee971d/Functions/Assertions/Be.ps1#L201).  Here is a repro directly calling `ArraysAreEqual`:

```PowerShell
InModuleScope Pester {
    Describe ArraysAreEqual {
        It '[xml] object is itself.' {
            $doc = [xml]''
            ArraysAreEqual -First $doc -Second $doc
        }
    }
}
```

alx9r:
I think the original repro does not fail with Pester 3.4.3 because 3.4.3 predates array comparison.  As of 2541ef9075ad36b842fc598fc10cb6a258ee971d Pester seems to treat most IEnumerables the same as arrays for comparison.  

I think that might not be a great strategy.  It looks to me like `ArraysAreEqual` uses the subscript operator to iterate into and compare the "Expected" and "Actual" arrays passed to `Should`.  I'm not aware of there being much of a relationship between an object being `IEnumerable` and what, exactly, the subscript operator returns.

alx9r:
[...]

PR #824, opened 2017-08-16
make ArraysAreEqual safer

alx9r:
self-limit recursion depth to cause fast(er) failure in case cyclic object graphs are encountered
rename IsCollection to IsArray


alx9r:
@nohwnd I think I've implemented the remainder of [your recommendations](https://github.com/pester/Pester/issues/785#issuecomment-322821470).  Or at least my attempt.  Would you like me to do anything else in this PR? Or can you take it from here?

nohwnd:
I can, but tomorrow, it's already nigh here :)

Issue #833, opened 2017-08-25
Parameterized Tests don't get marked as Inconclusive when empty


Remco-74:
When using an empty parameterized test, the test does not get marked as 'Inconclusive'. It does get marked as 'Passed'.

    Describe 'Defect: Parameterized Tests do not get marked as Inconclusive when empty.' {

        BeforeAll {
            $testCases = @{ Name = 'a' }
        }

        It 'will mark an empty test method as Inconclusive' {

        }

        It 'will mark an empty parameterized test method as Inconclusive' -TestCases $testCases {
            param($Name)

        }
    }

Results in:

    Executing script .\Empty.tests.ps1

      Describing Defect: Parameterized Tests do not get marked as Inconclusive when empty.
        [?] will mark an empty test method as Inconclusive 31ms
        [+] will mark an empty parameterized test method as Inconclusive 11ms
    Tests completed in 42ms
    Tests Passed: 1, Failed: 0, Skipped: 0, Pending: 1, Inconclusive: 0

It would be expected that a parameterized test case is evaluated in the same way as a normal test case.

it-praktyk:
@Remco-74, IMHO, it's strictly related to #826. I think that a one of these should not be processed. 

@nohwnd, are you agree?

nohwnd:
We should keep both. 
- #826 describes what should happen when no test cases are provided. 
- This describes what happens if you provide test cases but keep the _body_ of the test "empty" (just having the parameter block and nothing else).

it-praktyk:
OK, thx for a clarification.

Remco-74:
Correct, these are two diffrent issues. 

I could not locate the code that marks testcases inconclusive when empty, otherwise i could try to submit a pr. Can anyone point me in the right direction?

nohwnd:
@Remco-74 Sure [here it is](https://github.com/pester/Pester/blob/23eaf10c354eb4fbebc9d44f960bec89b24b5976/Functions/It.ps1#L161).

Remco-74:
I've changed the test to:

    #mark empty Its as Pending
    #[String]::IsNullOrWhitespace is not available in .NET version used with PowerShell 2
    **$scriptIsEmpty = (($null -EQ $script.Ast.BeginBlock) -and ($null -EQ $script.Ast.ProcessBlock) -and ($null -EQ $script.Ast.EndBlock))
    if ($PSCmdlet.ParameterSetName -eq 'Normal' -and $scriptIsEmpty)**
    {
[...]

PR #835, opened 2017-08-26
Fixed detection of empty tests

Remco-74:
This is a fix for issue https://github.com/pester/Pester/issues/833


nohwnd:
Thanks for the PR, but you have to support PowerShell v2, please review the failure log on our build server to learn more about why it fails on PowerShell v2. (Imho it's the Ast, and you should rather depend on regex - at least for powershell 2).

Functions/It.ps1
```diff
@@ -158,8 +158,12 @@ function ItImpl
 
     #mark empty Its as Pending
     #[String]::IsNullOrWhitespace is not available in .NET version used with PowerShell 2
```
it-praktyk: @Remco-74, exactly in this line is an answer why the sollution proposed by you doesn't work with PowerShell 2.

Can you change your code to use the old detection method when it's run with PowerShell 2.0 and the new one for the newer version of PowerShell?

PR #837, opened 2017-08-27
Write better readme

nohwnd:


nohwnd:
@it-praktyk I think that too, but let's keep that for 4.1 when we add some more documentation. (Mainly I am trying to avoid getting more work assigned to v4 again :) )

Issue #808, opened 2017-07-31
Gherkin related functions - partial lack of help


it-praktyk:
One of the biggest features of Pester v4 is the implementation of Gherking-style (is it a correct description?).

I checked a status of help for Gherkin related functions - I used for this https://github.com/it-praktyk/Pester/tree/DevV4_Addd_Tests_For_Help_2 - and help is not completed - results are listed below.

I think that it should be added before releasing of final Pester v4 version.

What do you think?

Guys - especially: @Jaykul, @adbertram, @KevinMarquette - help is needed.

```
    Context The function [Find-GherkinStep] - Help
      [-] Synopsis for the function is filled up 167ms
        Expected: value was {Find-GherkinStep [[}, but should not have been the same
        at <ScriptBlock>, C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Help.Tests.ps1: line 43
        43:                     $HelpSynopsisBegin | Should Not Be "$Function [["
      [-] Description for the function is filled up 66ms
        Expected: value to not be empty
        at <ScriptBlock>, C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Help.Tests.ps1: line 59
        59:                     $FunctionDescription | Should not BeNullOrEmpty
      [-] The parameter [BasePath] contains description 95ms
        Expected: value to not be empty
        at <ScriptBlock>, C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Help.Tests.ps1: line 81
        81:                         $ParameterDescription | Should not BeNullOrEmpty
      [-] The parameter [Step] contains description 62ms
        Expected: value to not be empty
        at <ScriptBlock>, C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Help.Tests.ps1: line 81
        81:                         $ParameterDescription | Should not BeNullOrEmpty
      [-] Example - At least one example exist 73ms
        Expected {0} to be greater than {0}
        at <ScriptBlock>, C:\Users\Wojtek\Documents\Scripts\2 - GitHub Extenal\Pester\Help.Tests.ps1: line 95
        95:                     $ExamplesCount | Should BeGreaterthan 0

    Context The function [Get-MockDynamicParameters] - Help
      [+] Synopsis for the function is filled up 176ms
      [-] Description for the function is filled up 58ms
        Expected: value to not be empty
[...]

PR #839, opened 2017-08-28
Invoke-Gherkin - reorganization and update of help

it-praktyk:
@KevinMarquette, I've added links to your blog post.

@Jaykul, I did this change to align organization of a comment based help to way as it's in other functions of  Pester.

The related issue: #808.


Jaykul:
🤷‍♂️ I don't like having the parameter help so far from the parameters, because I don't trust people (me) to update it, but I'm not going to fight against the module standard 😁

PR #865, opened 2017-09-06
Ship multiple gherkin assemblies

Jaykul:
Shipping gherkin.dll (and Newtonsoft.Json) for both .NET35 (PS2) and Core

Note that the .NET 3.5 build (in the "legacy" folder) works fine on .NET 4.x so I didn't bother shipping the .NET 4.x specific version -- although we could, if we want to go for the XXL module size ;-)

Also, if you want to build your own .NET 3.5 version of gherkin.dll, you need [my fork](https://github.com/jaykul/gherkin-dotnet/tree/Pester-PowerShell2-Net35)...


it-praktyk:
@Jaykul, could you post also SHA256 sums of dll files, please? 

After merging to https://github.com/it-praktyk/pester/tree/DevV4-1_Linux_compatibility_3 your commits I observe issue with loading - but it's caused probably by damaged dll files in my copy. 

Thank you.

PR #871, opened 2017-09-10
Fix manifest

### LINKED ISSUE #802: Compare version in changelog to branchtag does not compare pre-release
Got me this morning, I've put `4.0.5-rc` in the changelog, and then tagged the branch with `4.0.5` which was stupid, because the package become a release version on choco. The release scripts allowed me to release without a problem which should not be the case. 

The changelogversion and the tag should be compared as a whole string, then any of them should be compared to the module version from psd1. That way we are sure the changelog does not say RC and releasing is not done as stable, and hopefully avoid stupid mistakes.

nohwnd:
Fix #798
Fix #678
Fix #802


Pester.Tests.ps1
```diff
@@ -21,33 +24,9 @@ Describe -Tags 'VersionChecks' "Pester manifest and changelog" {
...
-        $script:changelogVersion -as [Version]  | Should Not BeNullOrEmpty
-    }
-
-    It "changelog and manifest versions are the same" {
-        $script:changelogVersion -as [Version] | Should be ( $script:manifest.Version -as [Version] )
-    }
-
-    if (Get-Command git.exe -ErrorAction SilentlyContinue)
-    {
+    if (Get-Command git.exe -ErrorAction SilentlyContinue) {
```
it-praktyk: Yeah, the next one command to fix to achieve compatibility with OSes other than Windows :-/

Issue #875, opened 2017-09-11
Review of examples for Should need to be done due to changed syntax


it-praktyk:
The quote from incoming Pester v. 4.0.7 RTM

> New assertion syntax with assertions as parameters to Should, that allows richer assertion vocabulary. Instead of Should Not Be 10 you should write Should -Not -Be 10

A documentation of project needs to be reviewed and aligned to address that preferred syntax.

it-praktyk:
The related pull request #903

it-praktyk:
I've updated the wiki pages about Should
- [Pester v4.x](https://github.com/pester/Pester/wiki/Should)
- [Pester v3.x](https://github.com/pester/Pester/wiki/Should-v3)
and cross-linked them so I'm closing the issue now.

PR #876, opened 2017-09-13
the about_Should updated to the new syntax

it-praktyk:
The related issue #875 - a partial fix


it-praktyk:
The changes were included in #903

PR #879, opened 2017-09-15
Merge Pester v. 4.0.7 to Dev4_1_PSCore_compatiblity

it-praktyk:


nohwnd:
What is happening here? Seems that you just need to merge master into the pscore compatibility branch. :)

it-praktyk:
Yeap. But I need to find when I did mistake on a test what throws an exception - before or under resolving a merge conflict.

it-praktyk:
The issue `The variable '$IsCoreCLR' cannot be retrieved because it has not been set.` is known issue and will be resolved soon.

PR #884, opened 2017-09-16
Build on Linux on Travis CI

nohwnd:


nohwnd:
It is setup to only build when `.travis.yml` is present, so we should see the checks from travis only on Linux compatible builds.

nohwnd:
@it-praktyk The PR is fine, but the tag tests are not excluded (I will probably wrap them in block and add environment variable to the batch file I am using to validate everything is fine, and to the TFS release build), and there are few other problems with the compatibility, I suggest this to be merged so we start building on linux.

PR #893, opened 2017-09-21
Allow assertion operators to be re-registered if they are identical

### LINKED ISSUE #891: Custom assertion operators can't easily be reused
I have a few custom assertion operators which I'd like to use in multiple test suites, but Pester only allows an assertion to be defined once.

### Assertions.ps1
```powershell
function IsAwesome($ActualValue, $ExpectedValue, [switch]$Negate) {
  # ... snip implementation ...
}
Add-AssertionOperator -Name IsAwesome -Test $function:IsAwesome
```

### Get-SomethingInteresting.Tests.ps1
```powershell
. "$PSScriptRoot\Assertions.ps1"
Describe 'Get-SomethingInteresting' {
  # ... snip tests ...
}
```

### Set-SomethingInteresting.Tests.ps1
```powershell
. "$PSScriptRoot\Assertions.ps1"
Describe 'Set-SomethingInteresting' {
  # ... snip ...
}
```
----

## Actual result
```
PS> Invoke-Pester
Executing script Get-SomethingInteresting.Tests.ps1
  (...snip output...)
Executing script Set-SomethingInteresting.Tests.ps1
  [-] Error occurred in test script 'Set-SomethingInteresting.Tests.ps1' 59ms
    at Assert-AssertionOperatorNameIsUnique, C:\path\to\Pester\4.0.8\Pester.psm1: line 243
    Assertion operator name 'IsAwesome' has been added multiple times.
```
## Expected result
I can reuse `Assertions.ps1` in multiple test files, somehow.

As a workaround for this I could create a single `TestEntryPoint.Tests.ps1` which dot-sources `Assertions.ps1` plus all of my test files, then run my tests by calling `Invoke-Pester -Script TestEntryPoint.Tests.ps1`, but I like simply running `Invoke-Pester` and having everything work.

brantb:
Fixes #891.

I feel it's a bit redundant to have both `Test-AssertionOperatorIsDuplicate` *and* `Assert-AssertionOperatorNameIsUnique` but I didn't want to tear up and refactor _too_ much. Let me know what you think.

### DIFF
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -203,19 +203,24 @@ function Add-AssertionOperator
         [switch] $SupportsArrayInput
     )
 
-    $namesToCheck = @(
-        $Name
-        $Alias
-    )
-
-    Assert-AssertionOperatorNameIsUnique -Name $namesToCheck
-
     $entry = New-Object psobject -Property @{
         Test                       = $Test
         SupportsArrayInput         = [bool]$SupportsArrayInput
         Name                       = $Name
         Alias                      = $Alias
     }
+    if (Test-AssertionOperatorIsDuplicate -Operator $entry)
+    {
+        # This is an exact duplicate of an existing assertion operator.
+        return
+    }
+
+    $namesToCheck = @(
+        $Name
+        $Alias
+    )
+
+    Assert-AssertionOperatorNameIsUnique -Name $namesToCheck
 
     $script:AssertionOperators[$Name] = $entry
 
@@ -227,7 +232,29 @@ function Add-AssertionOperator
 
     Add-AssertionDynamicParameterSet -AssertionEntry $entry
 }
+function Test-AssertionOperatorIsDuplicate
+{
+    param (
+        [psobject] $Operator
+    )
 
+    $existing = $script:AssertionOperators[$Operator.Name]
+
+    # Test SupportsArrayObject and scriptblock (as string) are identical
+    $isDuplicate = $existing -and
+        $Operator.SupportsArrayInput -eq $existing.SupportsArrayInput -and
+        $Operator.Test.ToString() -eq $existing.Test.ToString()
+
+    # Test Alias
+    foreach ($string in $Operator.Alias | Where { -not (Test-NullOrWhiteSpace $_)})
+    {
+        if (-not $existing.Alias -contains $string) {
+            $isDuplicate = $false
+        }
+    }
+
+    return $isDuplicate
+}
 function Assert-AssertionOperatorNameIsUnique
 {
     param (
--- a/Pester.psm1
+++ b/Pester.psm1
@@ -245,11 +245,13 @@ function Test-AssertionOperatorIsDuplicate
         $Operator.SupportsArrayInput -eq $existing.SupportsArrayInput -and
         $Operator.Test.ToString() -eq $existing.Test.ToString()
 
-    # Test Alias
-    foreach ($string in $Operator.Alias | Where { -not (Test-NullOrWhiteSpace $_)})
-    {
-        if (-not $existing.Alias -contains $string) {
-            $isDuplicate = $false
+    # Test Aliases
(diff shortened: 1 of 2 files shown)


brantb:
I'm unsure what is causing the CI tests to fail here -- Could anyone help me out with this? :)

Pester.psm1
```diff
@@ -227,7 +232,31 @@ function Add-AssertionOperator
...
+
+    $existing = $script:AssertionOperators[$Operator.Name]
+
+    # Test SupportsArrayObject and scriptblock (as string) are identical
+    $isDuplicate = $existing -and
+        $Operator.SupportsArrayInput -eq $existing.SupportsArrayInput -and
+        $Operator.Test.ToString() -eq $existing.Test.ToString()
 
+    # Test Aliases
+    if ($Operator.Aliases) {
```
dlwyatt: The problem is here.  Should be `$Operator.Alias`.  Having the wrong property name is causing strict mode failures.

brantb: 🤦‍♂️  Oops. Good catch.

dlwyatt:
Interesting... same error even with the new property name:  `Property 'Alias' cannot be found on this object. Make sure that it exists.`

Try changing `[psobject] $Operator` in the param block to just `[object] $Operator`.  PSObject has a bit of a different meaning that may be causing some trouble in PSv2.

Pester.psm1
```diff
@@ -227,7 +232,31 @@ function Add-AssertionOperator
...
+    # Test SupportsArrayObject and scriptblock (as string) are identical
+    $isDuplicate = $existing -and
+        $Operator.SupportsArrayInput -eq $existing.SupportsArrayInput -and
+        $Operator.Test.ToString() -eq $existing.Test.ToString()
 
+    # Test Aliases
+    if ($Operator.Alias) {
+        foreach ($string in $Operator.Alias | Where { -not (Test-NullOrWhiteSpace $_)})
+        {
+            if ($existing.Alias -and (-not $existing.Alias -contains $string)) {
```
brantb: Running the tests locally in PSv2, it seems to be throwing an error on this line. I've never had to write PSv2-compatible code before so I'm not sure what's wrong with this.

dlwyatt:
Same error.  It's a bit awkward to troubleshoot because the line number / content isn't showing up in the log, but I suspect this is coming from a different line now:

```posh
    # Test Aliases
    if ($Operator.Alias) {
        foreach ($string in $Operator.Alias | Where { -not (Test-NullOrWhiteSpace $_)})
        {
            if ($existing.Alias -and (-not $existing.Alias -contains $string)) {
                $isDuplicate = $false
            }
        }
    }
```

We know `$Operator` is not null, and Alias is always a property on these.  However, at this point in the code, `$existing` may be null, and I suspect that PSv2 behaves differently when we try to access $existing.Alias .  I'd suggest refactoring the function a bit to account for that:

```posh
function Test-AssertionOperatorIsDuplicate
{
    param (
        [object] $Operator
    )

    $existing = $script:AssertionOperators[$Operator.Name]
    if (-not $existing) { return $false }

    return $Operator.SupportsArrayInput -eq $existing.SupportsArrayInput -and
           $Operator.Test.ToString() -eq $existing.Test.ToString() -and
           -not (Compare-Object $Operator.Alias $existing.Alias)
}
```

brantb:
That did the trick, with an additional null check for `$Operator.Aliases`. Let's see what TeamCity thinks!

dlwyatt:
You shouldn't have to null check that, we have total control over that property.  Adding a default value to the `$Alias` parameter in `Add-AssertionOperator` should do the trick:

```
        [ValidateNotNullOrEmpty()]
        [string[]] $Alias = @(),
```

dlwyatt:
Might also be a good idea to add `[AllowEmptyCollection()]` there, just in case we ever wind up calling it with splatting or something.

Issue #904, opened 2017-10-11
Failed Gherkin tests do not report the correct error.


RandomNoun7:
When a test in a *.steps.ps1 file fails, the error reported in some cases is not the error the caused the test to fail, but an internal error handling the ErrorRecord object. To illustrate, create a feature file and a test file as specified below and run Invoke-Gherkin. The error you receive will not be expected error of $true <> $false, but ```ConvertTo-FailureLines : Cannot bind argument to parameter 'ErrorRecord' because it is null.```

bugrepro.feature
```cucumber
Feature: Reproduce Gherkin Bug
If a Pester Test fails in the Gherkin run mode, Gherkin.ps1 does not handle the error record properly and causes an internal failure, rather than reporting the real error that occurred.

    Scenario: Gherkin Bug
        Given A test to run
        Then Pester reports an internal error
```

bugrepro.steps.ps1
```powershell
Given "A test to run" {
}

Then "Pester reports an internal error" {
    $true | Should Be $false 
}
```
Bug Report Info
```
Pester version     : 4.0.8 C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.8\Pester.psm1
PowerShell version : 5.1.15063.483
OS version         : Microsoft Windows NT 10.0.15063.0

```

myriad90:
Seeing the same issue with same version of Pester. Basic assertion tests fail exactly as described in this bug

ConvertTo-FailureLines : Cannot bind argument to parameter 'ErrorRecord' because it is null.
At C:\Program Files\WindowsPowerShell\Modules\Pester\4.0.8\Functions\Output.ps1:267 char:25
+                         ConvertTo-FailureLines |
+                         ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [ConvertTo-FailureLines], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,ConvertTo-FailureLines

PR #905, opened 2017-10-11
Fix an issue returning errors from Gherkin tests.

RandomNoun7:
In reference to Issue #904

### DIFF
--- a/Functions/Gherkin.ps1
+++ b/Functions/Gherkin.ps1
@@ -698,6 +698,10 @@ Pester state object. For internal use only
         }
         ${Pester Result} = ConvertTo-PesterResult -ErrorRecord $PesterException
 
+        if ($PesterException.ErrorRecord) {
+            $PesterException = $PesterException.ErrorRecord
+        }
+
         # For Gherkin, we want to show the step, but not pretend to be a StackTrace
         if(${Pester Result}.Result -eq 'Inconclusive') {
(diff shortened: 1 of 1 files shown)


RandomNoun7:
Those changes were more comprehensive than I had in mind, but I don't have a problem with them.

Jaykul:
@it-praktyk I believe your commit resolves the exception/ErrorRecord problem incorrectly. The whole problem is that the `$PesterException` variable frequently contains an `ErrorRecord` instead of an `Exception` -- if that's the case, you should pass it, not throw it away.

it-praktyk:
The commit comment mentioned above is not part of any current pull request to Pester.

Jaykul:
I have to say, I had forgotten all about this pull request (which I'd actually asked @RandomNoun7 to make, I think), and I included another fix for this bug in #931 when I saw the bug earlier today. (I was saying to myself: I was sure I had fixed this ... but since I didn't have a branch with the fix in it waiting to be pulled, I made one)...

@nohwnd @dlwyatt @it-praktyk who actually has merge permissions?

To whomever merges -- it would be good to merge this PR to give @RandomNoun7 credit. I'll be happy to rebase #931 to just do the `&$SafeCommands` stuff 😉

it-praktyk:
> @nohwnd @dlwyatt @it-praktyk who actually has merge permissions?
Technically all of us. But based on a previous agreement I don't merge to master.

Issue #906, opened 2017-10-11
Extend Pester Gherkin Support to Write Test Stubs from Feature Files


RandomNoun7:
My employer has a use case where we want people to be able to write business scenarios as Gherkin files, and then automatically turn those files into test stubs that others can implement.

I found that Pester already has a Gherkin parsing dll that I leveraged to make writing the stubs possible. Since I am using Pester's dll for this, I think it would be nice if this functionality was within Pester itself so I am not forced to use the dll from outside of Pester, which would be very messy. 

The new function is called ```Import-FeatureFile```

It takes a path, recurses through the folder structure, and writes stubs for any Gherkin files that it finds. 

If someone then updates a Gherkin file with more tests, just invoke ```Import-FeatureFile``` again, and it will add the new stubs to the existing test file.

I am very much open to any criticism of how this code looks, functions, or any other suggestions required to get this code ready for inclusion in the project.

Thanks,
Bill

it-praktyk:
> My employer has a use case where we want people to be able to write business scenarios as Gherkin files, 
> and then automatically turn those files into test stubs that others can implement.

Thank you for your contributions to Pester project. I hope that we will review it soon.

Can I ask you what resources do you use to learn employees write Gherkin tests? I'm asking because Gherkin is a relatively new feature of Pester and we still didn't complete documentation to it.

RandomNoun7:
@it-praktyk 
Right now it's just me learning it. This project is in the very earliest stages, so it's up to me to create the process and then train others. 

My main resources are just Cucumber and Gherkin documentation and [this excellent blog post](https://kevinmarquette.github.io/2017-04-30-Powershell-Gherkin-advanced-features/#table-support) by Kevin Marquette.

it-praktyk:
Feel free to contribute to the documentation too. Thank you in advance.

Jaykul:
This is a great idea...

nohwnd:
Pester v4 is maintenance only, and Pester v5 does not contain Gherkin anymore and won't add it in the future. It is up to the community around Gherkin to pick this up in a new module, or a fork of v4.

PR #907, opened 2017-10-11
Add a function to write test stubs for feature files.

RandomNoun7:
This is in regards to issue #906

### DIFF
--- a/Examples/Gherkin/Gherkin-Import.feature
+++ b/Examples/Gherkin/Gherkin-Import.feature
@@ -1,0 +1,12 @@
+Feature: Import a Feature File
+Pester should be able to ingest a feature file and write test stubs.
+
+    Scenario Outline: Import a parameterize scenario
+        Given An '<entity>' to test
+        When An '<entity>' exists
+        Then '<entity>' should have the correct '<value>'
+
+        Examples:
+            | entity  | value   |
+            | Entity1 | Value 1 |
+            | Entity2 | Value 2 |


RandomNoun7:
It looks like I forgot to add one of the required files for testing to the commit. I also just realized during use that I need to support the possibility of the user passing in a table parameter.

it-praktyk:
@RandomNoun7, please review results of tests run under a continues integration process. Thank you.

it-praktyk:
@Jaykul, can I ask you about review this pull request?

Thank you.

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
...
+                        
+                        Add-Stubs -stubFile $stubFile -parsedInput $parsed   
+                    }
+                }
+                catch {
+                    throw $_
+                }
+            }   
+        }
+        else {
```
Jaykul: You're duplicating code here. You need:
* One function that parses a file (it already exists, it's called `Import-GherkinFeature` and you're re-doing it here)
* Another function that enumerates all the feature files and calls the `Import-GherkinFeature` function for each.

If we need to, we can _refactor_ `Import-GherkinFeature` to return an object or a hashtable instead of an array so that you can re-use it here, but we don't want to duplicate the whole parser/parse/enumerate thing...

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
...
+    }
+
+    end {
+
+    }
+
+
+}
+
+function Add-Stubs {
```
Jaykul: You would need to check and determines whether the stub already exists --in any .Steps file. There's a function `Find-GherkinStep` you could use. You shouldn't re-generate steps that already have a definition, that could cause ambiguity errors.

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
...
+        [Parameter(Mandatory = $true)]
+        [Gherkin.Ast.Step]
+        $step
+    )
+    
+    $outString = $step.keyword + "`""
+    
+    $description = $step.text
+    
+    $regmatch = [regex]::new("<\w*>").matches($description)
```
Jaykul: I don't understand the assumption you're making here about having angle brackets in your step text...

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
...
+    if ($regmatch -ne $NULL) {
+        $outString += "`tparam("
+        foreach ($match in $regmatch) {
+            $paramName = $match.value.replace('<', '$').replace('>', '') + ','
+            $outString += $paramname
+        }
+        $outString = $outString.trimEnd(',') + ")`r`n"
+    }
+    
+    $outString += "`t`$true | Should Be `$false `r`n}`r`n"
```
Jaykul: The default implementation should probably call `Set-TestInconclusive` or something, rather than throwing meaningless errors

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
...
+        $stubFile,
+        # Parsed Gherkin Data
+        [Parameter(Mandatory = $true)]
+        [Gherkin.Ast.GherkinDocument]
+        $parsedInput
+    )
+    
+    $content = Get-Content -Path $stubFile -raw 
+
+    foreach($step in $parsedInput.Feature.Children.steps) {
```
Jaykul: Needs to work down to PowerShell 2, so we can't use unrolling

Functions/Gherkin.Tests.ps1
```diff
@@ -60,3 +60,57 @@ Describe 'Invoke-Gherkin' {
...
+    
+    $props = @{
+        Results              = invoke-gherkin $featureFilePath -PassThru -Show None
+        EntityParamOnlyCount = (Select-String -Path $testsFilePath -SimpleMatch 'param($entity)').count
+        DualParamCount       = (Select-String -Path $testsFilePath -SimpleMatch 'param($entity,$value)').count
+    }
+    
+    $oldContent = Get-Content $featureFilePath
+
+    $newContent = "`r`n`r`n`tScenario: newTeset`r`n`t`tGiven A New Test"
```
it-praktyk: Using the chars: "\`r\`n" will cause that the code will not be compatible with Linux and macOS. Please use ```$([System.Environment]::NewLine)``` instead.

Jaykul: Technically, ``` `r`n``` here won't hurt anything -- it's just going into a .feature file, and neither PowerShell nor Gherkin parsers care about line endings (regardless of OS).

I still think it should be changed to a multi-line string with the line breaks _within_ the string, for readability, but whatever you use needs to be consistent throughout (there are ``` `r`n``` all through this test code).

RandomNoun7:
Especially given the requested changes to incorporate Functions like ```Import-GherkinFeature``` it looks like I need to just delete this pull request and start over. I'll get started on that within the next couple weeks I think.

Examples/Gherkin/Gherkin-Import.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Import a Feature File
```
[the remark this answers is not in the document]
RandomNoun7: Should the new function be called something like New-GherkinFixture?

Functions/Gherkin.ps1
```diff
@@ -787,3 +787,150 @@ function ConvertTo-HashTableArray {
     }
 }
 
+function Import-FeatureFile {
```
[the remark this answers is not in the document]
RandomNoun7: Should the new function be called something like New-GherkinFixture?

Functions/Gherkin.ps1
RandomNoun7: If the user then writes two feature files that really do have scenarios by the same name, are we saying that we shouldn't allow that? It seems like that would be dictating to the user what their workflow should be.

Functions/Gherkin.ps1
RandomNoun7: The documentation I'v seen so far says that if you have variables in the step text that correspond to something like an examples table, that the word representing the variable name should be enclosed in brackets. [Cucumber Documentation](https://github.com/cucumber/cucumber/wiki/Scenario-Outlines)
That regex just seemed an easy way to detect when someone was attempting that and to figure out what the variable name should be. If that's not the case I'm happy to find some other way to detect that.

Functions/Gherkin.ps1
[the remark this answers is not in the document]
Jaykul: Maybe. I've been thinking about this off and on since I saw the request, I feel like probably the best workflow is that you're going to want to run this repeatedly as you add test cases, so I wanted to focus on the idea of _updating_ and generating _missing_ steps, but I can't settle on a verb. Perhaps Update-GherkinFixture?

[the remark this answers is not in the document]
RandomNoun7: I remember now that I settled on Import-FeatureFile because I didn't want to use the New or the Update verbs. 

The reason is I had in mind a script developer (specifically myself) being able to use a single Cmdlet in a CICD script that would both create new stub files if new feature files were created, but also create new test stubs in existing files if an analyst added new scenarios to an existing feature file. In that case it didn't seem like either the New or the Update verb was quite right. 

The idea though is that the business analysts writing scenarios, and the PowerShell developers writing the body of the tests inside the stubs, aren't the same people. So what I want is a repo where analysts contribute to the development of feature files, and then when the repo is pushed up, the CICD process automatically creates or updates the stubs as needed, and if new tests are present, they will automatically either fail or go inconclusive, and those new tests showing up in the dashboard (implemented elsewhere of course), would be one of the triggers to developers that they need to go in and fill out the tests.

So given that we want to avoid the name collision with Import, and I also think New or Update are perhaps not quite right, maybe we could use Publish instead? The alternative might be to use the New verb, but also make Update public, and New would detect if a stub already exists and just call Update in the background for you. That would make both of them available, but in reality a CICD script author could just use New all the time and update would be only for interactive use if the feature file and the gherkin test authors are the same person.

What I don't want is for the person writing the CICD pipeline script to have to go to the trouble of figuring out which files in a repo should have to have New called on them, and which ones already have some stubs written and therefore need Update called instead.

Functions/Gherkin.ps1
Jaykul: This is simply a namespace/scope issue. 

In a single application (or module), where you have all your specs (.feature files) together in one folder, you don't want `When I create a user` to have different implementations depending on which file it's in -- that would be confusing. 

That is: when you have two *steps* with the same text, they should match the _exact_ same implementation across all your features. In fact, I frequently create a single "Steps" file, regardless of the number of feature files, so I don't have to keep track of which steps I've already implemented in another file.

Functions/Gherkin.ps1
Jaykul: Oh, yes, `Scenario Outlines` ... that's right. Of course, it's not quite simple as that, there are also doc strings and data tables, and once people start customizing the [step definition](https://cucumber.io/docs/reference#step-definitions), anything can have parameters... which is why you need to use `Find-GherkinStep`

RandomNoun7: I'll definitely need to do that. I discovered not long after this PR that my parameterization code wasn't working properly in cases of things like doc strings or passing in JSON strings and such.

Functions/Gherkin.ps1
RandomNoun7: I agree that it would be confusing to have a single scenario with multiple implementations, but my question would be, how much of a single view of how to organize one's work should be enforced by Pester? 

I personally had envisioned a very limited scope of behavior for this feature, which is that for each feature file the code finds, it should create a corresponding steps file with the scenario stubs implemented, such that the behavior is as predictable or as understandable after the fact as possible by reading the documentation for Cucumber, not for Pester.

With that in mind, I agree that multiple implementations could be confusing, but that confusion would be the responsibility of the developer to either resolve, or decide that the duplication is intentional. On the other hand, if that duplication is intentional, for whatever reason, and Pester refuses to implement the test stub, that confusion is now our fault.

So when we talk about tracking which stubs we have implemented and avoiding duplication, I think of some of the following scenarios:
1. Two different features files in the same directory specify the same scenario  
  Should we create only a single steps file for all features files in the directory? If that's the case then given that the test runs are matched based on regex then I think we would indeed only implement the stub once. But I honestly don't prefer to work this way. I prefer lots of small features and steps files to keep them short and to the point. Perhaps we could support a switch parameter option that would force all steps files in a directory to combined into a single file to support both of our work flows.  
2. Two different features files in different directories specify the same scenario.  
  Should we make any attempt to prevent duplication across directories? Perhaps if we do want to support that, instead of a switch param to combine in a single directory, we could support a ```-stepsScope``` param that would offer a choice of ```file```, ```directory```, ```all``` and if all is selected, a ```-stepsFileOutputPath``` param that would specify where the master steps file should be written, that would default to ```$pwd```

I bring up scenario number two because in the project I was asked to implement this feature for, we did indeed have the scenarios files spread out in the repository according to the place where it made the most sense to find them. Scenarios for server side code were in one folder, and scenarios that dictated how the client should behave were in a different folder that corresponded to the client.

Jaykul: It's not about enforcing a single way of doing things. In particular, we cannot enforce any relationship between feature files and step files. That is, we don't load <FOO>.feature and look for <FOO>.steps.ps1, we load *.steps.ps1 recursively, and then we start running through features.

Part of the reason I like Gherkin is that I automatically get to re-use _step definitions_ in many of scenarios, which usually ends up meaning you write a lot less code than you typically do in Pester's RSpec-based syntax... 

Normally you keep all your .feature files **together**, because together they are the specification for your product. Obviously if you have one "project" with multiple products (i.e. multiple clients and a server), you can (and probably should) create separate specs folders...  Also, you can keep your .steps files in the same folder with the .feature files, or in a subfolder -- currently Pester's `Invoke-Gherkin` merely requires them to be next to or in the -Path you run Gherkin against.  Obviously in that case, you can _already_ run Invoke-Gherkin multiple times, against different folders, but then you don't get _one_ result count, and you don't get as much re-use.

## Step definitions are just functions. 

Pester's Gherkin (like all the others that I know of) finds *all* the steps at once, and basically sources them into Pester. If you define the same name in two different files, you're **overwriting** the definition, just as you would with functions in modules...

Imagine a user-provisioning module:

```gherkin
Feature: Creating AD User

Scenario: Simple
  Given we are an administrator
  When we create a user
  Then Get-ADUser should work
```

Imagine I've written a `ADUser.Steps.ps1` file and gotten everything working correctly.

Now it's time to add a new feature, so I add a feature file:

```gherkin
Feature: Creating O365 User

Scenario: Simple
  Given we are an administrator
  When we create a user
  Then Get-User should work
```

Then we generate stubs ... 

If you generate a new stub for `Given we are an administrator` and `When we create a user` in a new file `O365.Steps.ps1`, you _**will break** the existing tests_.  I don't see a way around that which doesn't force you to rewrite every step definition in every feature (which is certainly out of the question).

The bottom line is that we should avoid generating stubs that will break things. If you're running in a folder and the other step is defined in a different folder, you'll have no problems. If the other step is in the same folder, I'd suggest putting a comment in the generated file pointing out where the step is defined.

RandomNoun7: Maybe then only generating a single file per folder would make it easier, both to ensure I don't create two stubs for the same function, and to drive that point home to the user.

Jaykul: Might be! Because of the lack of tooling, it's _definitely_ easier to do that way right now. 

When I work with, say, SpecFlow in C#, they have good tooling in Visual Studio where I get intellisense while writing the .feature files to auto-complete step statements based on steps that I've already written. In that world, I don't care as much where my steps are, and I'm much more likely to break them up ... I believe they also have an indicator in the editor, while you're working on the feature file, to tell you if a step doesn't have an implementation ... and their tool for generating steps is also integrated to the editor.

In my ideal world, I'd love to write a Pester+Gherkin extension for VS Code, so I could run a _specific scenario_ from code, have the results show as red/green/black markers in the sidebar, like what https://wallabyjs.com/ does.  Such a plugin would be able to show you _as you write_ your feature file what the steps are that are already available, and would be able to generate new step placeholders ... all by calling the functions from Pester...

RandomNoun7: As I read your description, an extension for VSCode is exactly where I was thinking that belonged too. I think I have a path forward now for this feature though at least. 

I'm switching jobs this week, so things are going to be chaotic for me for a little, but I'll start work as soon as I can.

nohwnd:
@RandomNoun7 so can I close this? :)

PR #920, opened 2017-10-19
Added detailed code coverage. 

jonkeda:
Hi mr Wyatt,

I added a switch -DetailedCodeCoverage to the Invoke-Pester command. If this switch is on then the JaCoCo codecoverage file will have additional data. Per sourcefile a line counter and each line covered or missed numbers are added to the export of de JaCoCo codecoverage file. Furthermore the lines will not be printed to the console.

The JaCoCo reader can be used to view the covered and missed percentages and show the sourcefiles with red and green line coloring.

This tool can be found here: https://visualstudiogallery.msdn.microsoft.com/14da4aa2-e833-4f1d-8765-0d3d79fd08a3?redir=0

One unittest was added to check the creation of the right data.

As an example run the following:

1. Invoke-Pester ** -CodeCoverage Pester.psm1 -detailedcodecoverage -CodeCoverageOutputFile pestercc.xml
2. And view the pestercc.xml with the JaCoCo Reader.

Next step is to add the code coverage reader functionality to the PoshTools.

Let me also thank you for Pester.

Kind regards,
Daan Jonkers


it-praktyk:
@jonkeda,
thank you for your contribution.

Please check the [result of tests](https://build.powershell.org/viewLog.html?buildId=13058&buildTypeId=Pester_TestPester&guest=1) - your code is not compliant with the style rules defined for Pester.

Pester.psm1
```diff
@@ -761,6 +761,8 @@ New-PesterOption
         [ValidateSet('JaCoCo')]
         [String]$CodeCoverageOutputFileFormat = "JaCoCo",
 
+		[Switch]$DetailedCodeCoverage = $false,
```
it-praktyk: The newly added switch is not described in the comment based help.

jonkeda:
Fixed the style errors and added a comment on the DetailedCodeCoverage switch parameter.

Currently it fails localy on the release version tag. 

Also added a -CodeCoverage parameter to the Describe function. If the DetailedCodeCoverage switch is on then the CodeCoverage files are collected by reading the parameter from the describe function.


For instance add the following to the Coverage.Tests.ps1 file:

-CodeCoverage $PSScriptRoot\Coverage.ps1

and it will cover this file also.

scriptkiddie123:
@jonkeda it is a bit difficult to test with test runner, as you have added an additional flag.
The way I see it you don't need a flag for extra details. The only reason the current implementation is so brief, is because I could not be bothered, to make it more detailed as TFS/VSTS did not have build in support for more. In other words, Pesters code coverage reports should always be detailed.

I have just tried to setup VSTS to build Pester with Pester Test Runner, but a lot of the test failed, so I think it will take me a couple of hours to figure out how to get a "beta" version of Pester into a VSTS build.

I have a few other projects on my todo list, but I will return this soon as it seems quite cool :-)

Functions/Coverage.ps1
```diff
@@ -621,6 +629,80 @@ function Get-JaCoCoReportXml {
     $jaCoCoReportXml.report.name = "Pester ($now)"
     $jaCoCoReportXml.report.sessioninfo.start=$startTime.ToString()
     $jaCoCoReportXml.report.sessioninfo.dump=$endTime.ToString()
+
+    if ($DetailedCodeCoverage)
+    {
+
+        $fileName = ""
```
it-praktyk: The variable assigned but not used. Can you check why?

ChrisLGardner:
Any reason that DetailedCodeCoverage isn't just the default output when CodeCoverageOutputFile is specified? It seems like an unnecessary switch. 

I'd also suggest still outputting to the screen as it's quite handy to see at a glance what has been missed if you're not using the JaCoCo reader.

it-praktyk:
@ChrisLGardner,
- because is not merged yet ;-)
- because it's a breaking change and IMHO shouldn't be introduced without changing a major version number.

nohwnd:
@it-praktyk Is it really a breaking change? I haven't tested it yet. But to me the switch also seems unnecessary. @ChrisLGardner have you tried it? I don't have experience with the jacoco format? Can we make the detailed format the default, or would it break existing builds?

nohwnd:
@ChrisLGardner would be great if you could test it a bit more 👍

nohwnd:
@jonkeda Thanks I merged it as is, if we figure out the change is not breaking I will just make it the default without the `-DetailedCodeCoverage` option. 😊

nitinkshirsagar:
After generating coverage report XML file for pester tests, How to publish jacoco report on jenkins ?

nohwnd:
@nitinkshirsagar I haven't worked with Jenkins in few years, but I would say you need a plugin, such as [this one](https://wiki.jenkins.io/display/JENKINS/JaCoCo+Plugin), and give it path or wildcarded path to your code coverage report.

PR #924, opened 2017-10-29
Incorrect name of the variable used

it-praktyk:
CC: @Jaykul


Jaykul:
Oh, this is another one I included this in my other PR(s).

Issue #636, opened 2016-11-11
Create pester contribution guide


nohwnd:
The process of proposing a new feature to Pester and implementing it through PR needs to be better described to avoid handing out partial information in every PR that proposes a new feature. @adbertram already summed up some of the process in this wiki article https://github.com/pester/Pester/wiki/Contributing-to-Pester and I plan to continue expanding on that. Just from the top of my head I'd say we need: Reporting a bug, Proposing a new feature, Implementing a bug fix, implementing new feature, publishing functions from Pester, making sure tests pass, and then describe how the code is tested on TC and that some of the tag tests may fail. Anything else?

it-praktyk:
Maybe creating a template for issue - like described [here](https://github.com/blog/2111-issue-and-pull-request-templates) is a good idea?

A proposed template content
- information about what data is necessary to report bug
- link to [Contributing to Pester](https://github.com/pester/Pester/wiki/Contributing-to-Pester)

nohwnd:
The pull request template would be also nice, the contribution guide seems comprehensive enough as is.

it-praktyk:
The pull request template is currently on place - https://raw.githubusercontent.com/pester/Pester/master/.github/PULL_REQUEST_TEMPLATE.md

nohwnd:
Ok, closing :)

PR #929, opened 2017-11-02
Created Github issue template

BrandonLundt:
To assist with Issue #636 I would like to propose the following template.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,11 +1,12 @@
 ## 1. Provide a general summary of the issue in the Title above
-<!--
+
 Please review https://github.com/pester/Pester/wiki/Contributing-to-Pester prior to submitting an issue.
--->
+
 ## 2. Describe Your Environment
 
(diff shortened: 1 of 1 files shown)


it-praktyk:
I've asked @BrandonLundt to prepare the issue template for Pester when I've found his contribution in the [PSGitLab](https://github.com/ngetchell/PSGitLab/tree/master/.github) repository useful.

Thank you @BrandonLundt!

nohwnd:
Why are there the HTML comments in the file? Do they render as hints or something? I don't see them used in the linked issue template, and did not find anything mentioned in github documentation.

johlju:
The comments are probably used for examples and text that shouldn’t be shown if the user creates an issue and just add information (and not remove the help/guide text). If the comments would not be there. If the user did not remove the help/guide text then it would show in the issue. If the help/guide text would contain and issue reference example (issue number) then a lot of issues would point to the same issue. Not sure if that’s the case here. 
But personally I rather like seeing the help/guide text in comments.

PR #925, opened 2017-10-29
Update Pester to work on PowerShell Core at Windows, Linux, macOS

it-praktyk:
The related issue: 
- #639 
- #838 
- #864
- #870

PR #931, opened 2017-11-05
Fix #904 output proper failure messages

### LINKED ISSUE #904: Failed Gherkin tests do not report the correct error.

Jaykul:
This fixes #904 the "Convert-ErrorLines" error

Also finally converts all the gherkin code to use the `&$SafeCommands` syntax the rest of Pester uses.
Also splits the Assemblies so there's one for legacy (down to PowerShell 2, with .Net 3.5) and one for core


Jaykul:
P.S. If #925 gets merged before this one, I will be happy to rebase

Jaykul:
Note, the #932 PR is based on this one. Because of that (and how git works), #932 includes these two commits -- so if you merge that one, you don't need this one.

Issue #933, opened 2017-11-09
InModuleScope gives error "Module 'MyModule' is not a Script module." in "Manifest"-type modules


mikeclayton:
Hi,

I've got a PowerShell *.psd1 Module which references some *.ps1 files in NestedModules and I'm trying to mock the functions defined in those *.ps1 files, but I'm having some problems. I've seen similar issues that recommend using InModuleScope, but when I do that I get the following error:

```
[-] Error occurred in Describe block 516ms
      RuntimeException: Module 'MyModule' is not a Script module.  Detected modules of the following types: 'Manifest'
      at Get-ScriptModule, C:\src\MyProject\packages\Pester\4.0.8\Functions\InModuleScope.ps1: line 128
```

I'll post my module's code in a follow-up comment, but I think the problem is because Get-ScriptModule (in Functions\InModuleScope.ps1) is filtering on modules that are of type "Script" (i.e. *.psm1) files, but I've got *.ps1 files referenced in a *.psd1 "Manifest" module.

```
$scriptModules = @($modules | & $SafeCommands['Where-Object'] { $_.ModuleType -eq 'Script' })
```

I can **seem** to fix the issue by adding "Manifest" module types to the check, as follows:

```
$scriptModules = @($modules | & $SafeCommands['Where-Object'] { $_.ModuleType -in @( 'Script', 'Manifest' ) } )
```

I don't really understand the whole Pester codebase well enough to know whether this might break any other scenarios, but I'll happily send a Pull Request if this looks like a valid fix.

Cheers,

Mike

mikeclayton:
Code to reproduce issue:

**MyModule.psd1**
```
@{
ModuleVersion     = '1.0'
NestedModules     = @( "Invoke-PrivateMethod.ps1", "Invoke-PublicMethod.ps1" )
FunctionsToExport = @( "Invoke-PublicMethod" )
}
```

**Invoke-PrivateMethod.ps1**
```
function Invoke-PrivateMethod
{
    write-host "private method called!";
}
```

**Invoke-PublicMethod.ps1**
```
function Invoke-PublicMethod
{
    Invoke-PrivateMethod;
}
````

**Invoke-PublicMethod.Tests.ps1**
```
$ErrorActionPrefernce = "Stop";
Set-StrictMode -Version Latest;

Describe "Invoke-PublicMethod" {

    InModuleScope "MyModule" {

        Mock -CommandName "Invoke-PrivateMethod" `
[...]

Issue #812, opened 2017-08-03
Summary of scope-related differences for mocks between Pester 3 and 4


alx9r:
I've encountered a few scope-related edge cases for mocks that behave differently between Pester 3 and 4.  Here are suites of mock-scoping tests for [Pester 3](https://github.com/alx9r/ToolFoundations/blob/a7ad54074594a9cdf519fc253e8193675b170f4e/EnvironmentTests/pester3InModuleScopeTests.ps1) and [Pester 4](https://github.com/alx9r/ToolFoundations/blob/a7ad54074594a9cdf519fc253e8193675b170f4e/EnvironmentTests/pester4InModuleScopeTests.ps1).  Most of the scope-related mock behavior is identical between Pester 3 and 4, however, there are a handful of notable differences which are consolidated in [this gist](https://gist.github.com/alx9r/56bbc9077b39e5693b08821183d89c21).  Vaguely-speaking there are scope-related differences for mocks between Pester 3 and 4 in the following circumstances:

1. Indirectly invoking mocked commands without using InModuleScope at all when using F5 in ISE.
2. Indirectly invoking mocked commands from a module that doesn't import the module where the mocked command resides.

## Implications

Each of these are definitely edge cases but have the potential to be confusing (especially to new users).  

(1) is a new circumstance where Pester behaves differently when pressing F5 versus using `Invoke-Pester`.  
(2) is a new circumstance wherein a call to `Mock` succeeds, but the resulting mock is not invoked.

I don't think these warrant any action other than awareness.  I am creating this issue mainly as a reference for future users who also run into these edge cases.

nohwnd:
Linked to this in migration guide, closing so we can close the milestone.

Issue #387, opened 2015-07-15
Module '' is not a Script module.  Detected modules of the following types: 'Manifest'


alextc:
Hi,
I am getting this error message:
Module '' is not a Script module.  Detected modules of the following types: 'Manifest'
when attempting to use Mock -ModuleName when the module is a nested module.

Here is my test:

Import-Module LocalAdminPasswordReset

Describe "Assert-LocalAdminPasswordAgeHasNotExceededResetThreshold.psm1" {
	Context "When Threshold reached" {
		Mock -ModuleName LocalAdminPasswordReset Get-LocalAdminPasswordAgeInHours { 100 }

		It 	"Should return false when password age is greater than the threshold" {
			# Default threshold is 24 hours
			Assert-LocalAdminPasswordAgeHasNotExceededResetThreshold | Should be $false
		}	
	}
}

LocalAdminPasswordReset module (which is on PSModulePath) is a nested module. The psd1 file for the module contains multiple nested modules:
GetLocalAdminPasswordAgeInHours.psm1 where the LocalAdminPasswordAgeInHours CmdLet is 
defined.

AssertLocalAdminPasswordAgeHasNotExceededResetThreshold.psm1 where the Assert-LocalAdminPasswordAgeHasNotExceededResetThreshold is defined.

Is this a supported scenario? 
Are there workarounds to make this scenario work?

Thanks
Alex.

dlwyatt:
If I remember correctly, it should work if you specify the name of the nested module in your calls to Mock and/or Assert-MockCalled.

PR #934, opened 2017-11-09
Support "Manifest"-type modules in Get-ScriptModule

mikeclayton:
A fix for https://github.com/pester/Pester/issues/933 to support "Manifest"-type modules (i.e. no RootModule specified in a *.psd1 file).

### DIFF
--- a/Functions/InModuleScope.ps1
+++ b/Functions/InModuleScope.ps1
@@ -125,7 +125,7 @@ function Get-ScriptModule
 
         $actualTypes = $actualTypes -join ', '
 
-        throw "Module '$ModuleName' is not a Script module.  Detected modules of the following types: '$actualTypes'"
+        throw "Module '$ModuleName' is not a Script module.  Detected modules of the following types: '$actualTypes'.    See https://github.com/pester/Pester/wiki/Testing-different-module-types for more information."
     }
 
     return $scriptModules[0]


Functions/InModuleScope.ps1
```diff
@@ -108,7 +108,7 @@ function Get-ScriptModule
         throw "No module named '$ModuleName' is currently loaded."
     }
 
-    $scriptModules = @($modules | & $SafeCommands['Where-Object'] { $_.ModuleType -eq 'Script' })
+    $scriptModules = @($modules | & $SafeCommands['Where-Object'] { $_.ModuleType -in @( 'Script', 'Manifest' ) })
```
it-praktyk: Does the 'ScriptModules' have to be renamed due to changed usage?

Functions/InModuleScope.Tests.ps1
```diff
@@ -18,27 +18,70 @@ Describe "Module scope separation" {
...
 
     Remove-Module TestModule -Force
+
+}
+
+Describe "Executing test code inside a 'Manifest' module" {
+
+    New-Item -Type "Directory" -Path "TestDrive:\MyModule";
+
+    Set-Content -Path "TestDrive:\MyModule\Invoke-PublicMethod.ps1"  -Value 'function Invoke-PublicMethod { return Invoke-PrivateMethod; }';
```
it-praktyk: We adjust Pester to work on PowerShell Core (Windows/Linux/macOS) - please check the pull request #925.

Paths used by you will not be compatible with non-Windows systems :-/

Please check how I did it [here](https://github.com/pester/Pester/pull/925/commits/d999adfb2860aa781233df673e899457ad6b736b).

Functions/InModuleScope.ps1
it-praktyk: But generally, your pull request looks reasonable. Please rebase it after merging #925 to master.

Thank you for your contribution!

mikeclayton:
@it-praktyk - I've fixed the path issue. I can rename ```Get-ScriptModule``` if you like - there's only 3 references to it throughout the whole of Pester so it shouldn't be too hard, but I'm not sure what a good name might be... maybe ```Get-ScriptOrManifestModule```, or ```Get-MockedModule```?

Let me know what you think about the name and I'll roll that into the PR as well.

Cheers,

M

alx9r:
I wonder what, exactly, a Manifest-type module is and why it would be used instead of a Script-type module?

PowerShell scope has proven to behave inconsistently for different module types.  That inconsistency affects mock behavior.  [Here is just one such example wherein the behavior of a dynamic module is different from a file-backed module.](https://github.com/PowerShell/PowerShell/issues/4713)  That sort of variation results in bugs that are a lot of work to understand and overcome.  It might be worth reviewing the tests in #812 to get a sense of just how complex the way scope affects Pester mocks already is.  It seems like enabling "Manifest" type modules has the potential to compound that complexity.  Pester's own suite of tests does not cover the full breadth or depth of this complexity so it's not currently a straightforward task to confirm consistency of mock and scope behavior amongst the module types (or from one release of Pester to another, for that matter).

It seems to me that, prior to enabling Manifest-type modules, the benefit to users of supporting Manifest-type modules should be weighed against the burden of increased work for Pester maintainers to make Pester work correctly with the larger resultant matrix of cases.

mikeclayton:
@alx9r - yeah, i was hesitant to send a pull request as the workaround is pretty simple (i.e. just add an empty *.psm1 to the module under test), but I figured I'd send it anyway and let you decide if it was something you thought had value or not.

I'm not really sure what the extent of the differences is either, but I've worked with modules using both the "Manifest" approach and the "Script" approach (my preference is the Manifest approach, btw) and never really experienced any differences between them, although there could be extreme edge-case behaviours like you've mentioned which could come with a maintenance overhead.

Given that there's only one other similar issue - https://github.com/pester/Pester/issues/387 - and that was two years ago, I won't be offended if you reject this PR as "low value"!

mikeclayton:
Going back to article you linked to, the "NestedModules" key in the *.psd1 has this description:

```In addition to other modules, you may also load script (.ps1) files here. These files will execute in the context of the root module. (This is equivalent to dot sourcing the script in your root module.)```

It says "equivalent" rather than "identical", but this is basically how I normally import functions into the module scope, rather than dot-sourcing from a *.psm1. I guess it's a case of coding style, and it sounds like I'm in the minority given there's very few other issues opened when doing it this way without a RootModule :-).

**Comment by alx9r** [comment]
> Given that there's only one other similar issue - #387 - and that was two years ago, I won't be offended if you reject this PR as "low value"!

I think we're on the same page.  Given that this comes up from time-to-time I think it's worth trying to understand what, exactly, a manifest module is and make it easier for future users who encounter what you did to work around it.  Perhaps a better error message is the right solution here.

**Comment by mikeclayton** [comment]
Sure thing. I'm happy to have a go at a change which adds (ii) below:

(i) looks for a Script module with the given name (as per the current codebase) and uses it if found
(ii) otherwise looks for a Manifest module with the given name and throws a specific error if it finds one, (maybe providing a url linking to a wiki page with more info about adding an empty *.psm1 RootModule)
(iii) otherwise just throws the same error as it does in the current implementation

Does that sound like a reasonable approach for now? Or would you prefer to investigate Manifest modules a bit more first?

**Comment by alx9r** [comment]
>Or would you prefer to investigate Manifest modules a bit more first?

I'd like to at least find out what we can about manifest modules from the PowerShell team and go from there.

**Comment by mikeclayton** [comment]
Ok, cool. I'll hold off in that case and keep an eye out for an update.

M

**Comment by nohwnd** [comment]
@alx9r what is the plan with this PR? Shoule I close it and create an issue that requests a better error message (as mentioned in the conversation)? Or did you have a look and we can fix this now? If the workaround is simple, and the issue only pops up once in few years I would be totally for using the workaround to have less things to support.

**Comment by alx9r** [comment]
>what is the plan with this PR?

@nohwnd The nature of manifest-type modules is still a bit mysterious to me and I have found little demystifying documentation.  Currently I am hoping for a response to PowerShell/PowerShell#5399 from the PowerShell team.

>If the workaround is simple, and the issue only pops up once in few years I would be totally for using the workaround to have less things to support.

I support this.  Although without understanding more about manifest-type modules, it's hard to know whether the workaround will always be as simple as it was for #933.

**Comment by dlwyatt** [comment]
Sorry I haven't chimed in on this yet.  The reason Pester was looking specifically for Script modules came from this old issue:  https://github.com/pester/Pester/issues/275 .

**Comment by alx9r** [comment]
Thanks for the link to that issue @dlwyatt.

FWIW, one of the [conclusions that were drawn in that thread](https://github.com/pester/Pester/issues/275#issuecomment-72766380) was

>Pester's Mock and InModuleScope functionality seems pretty worthless for any module type other than Script anyway. If you bind InModuleScope (or Mock) to the binary module, they don't really work. For Manifest modules, it's even worse; the SessionState property for those is null, and Pester winds up throwing a different error.

This doesn't seem to be the case for manifest-type modules all of whose nested modules are `.ps1` files.  In that case `$ExecutionContext.SessionState` exists, and [allowing `Manifest` in `InModuleScope`](https://github.com/pester/Pester/pull/934/files#diff-8e7314c4915d39294c79736b9fd65560L111) seems to work.

AFAICT #275 involved a mix of binary and script implemenations which is probably where the difference originates.  I wonder if mixing binary and script implementations is the reason manifest-type modules exists in the first place.

**Comment by dlwyatt** [comment]
As long as you check for non-null SessionState, allowing Manifest modules should be okay, but I'd recommend prioritizing Script if both types are found (in case you have a manifest module with a nested script module of the same name.)

**Comment by mikeclayton** [comment]
I've tried to create some pathological test cases for Get-Modules returning multiple modules as the result of a single Import-Module to reproduce https://github.com/pester/Pester/issues/275 but I think I'm missing something (or maybe the behaviour of PowerShell has changed there in the last 3 years!!!).

Things I've tried:

**Manifest Module**
```
@{
ModuleVersion     = '1.0'
NestedModules     = @( "Invoke-PrivateManifestMethod.ps1", "Invoke-PublicManifestMethod.ps1" )
FunctionsToExport = @( "Invoke-PublicManifestMethod" )
}
```
```Get-Module mymodule``` returns one module of type "Manifest"

**Script Module**
```
@{
ModuleVersion     = '1.0'
RootModule        = "MyRootModule.psm1"
NestedModules     = @( "Invoke-PrivateScriptMethod.ps1", "Invoke-PublicScriptMethod.ps1" )
FunctionsToExport = @( "Invoke-PublicScriptMethod" )
}
```
```Get-Module mymodule``` returns one module of type "Script"

**Binary Module**
```
@{
ModuleVersion     = '1.0'
RootModule        = "MyBinaryModule\MyBinaryModule.dll"
NestedModules     = @( "Invoke-PrivateScriptMethod.ps1", "Invoke-PublicScriptMethod.ps1" )
FunctionsToExport = @( "Invoke-PublicScriptMethod" )
}
```
```Get-Module mymodule``` returns one module of type "Binary"

In all three cases the imported module has a non-$null SessionState, and on the surface of it the Manifest module seems to work ok, but the Binary module doesn't really play nicely with Pester mocking (which is already expected), although I'm not sure if any of this actually helps the discussion or not...

The pragmatic way forward seems to be to just improve the error message for the specific case of Manifest modules to suggest the RootModule workaround, and maybe revisit if the PowerShell team reply to https://github.com/PowerShell/PowerShell/issues/5399.

**Comment by nohwnd** [comment]
Sounds like a plan. Could you suggest wording for the error message please?

**Comment by mikeclayton** [comment]
Maybe just change the existing error message in ```Get-ScriptModule``` to:

```
"Module '$ModuleName' is not a Script module.  Detected modules of the following types: '$actualTypes'".  See https://github.com/pester/Pester/wiki/Pester/ModulesTypes for more information."
```

(I don't mind having a first stab at a "ModuleTypes" page for the wiki)

----
Incidentally, I found another small glitch in ```Get-ScriptModule```. It looks like the behaviour of the built-in ```Get-Module``` cmdlet has changed to return ```$null``` rather than throw if there are no modules loaded with the specified name, so the "try ... catch" doesn't "catch", and you get this error from ```Get-ScriptModule```:

```
 "Module 'MyNonExistentModule' is not a Script module.  Detected modules of the following types: ''"
```
instead of
```
No module named 'MyNonExistentModule' is currently loaded.
```
A simple fix is to add this after the existing "try ... catch":

```
    if ($modules.Length -eq 0)
    {
        throw "No module named '$ModuleName' is currently loaded."
    }
```

Should I raise a separate issue for this?

nohwnd:
@mikeclayton Yes, that would be great. 
I think a better title for the page would be [Testing different module types](https://github.com/pester/Pester/wiki/Testing-different-module-types).

mikeclayton:
I've added a first draft at https://github.com/pester/Pester/wiki/Testing-different-module-types. (I wasn't sure how to submit it for peer review first, so yell if I did it wrong!).

Feel free to edit, sanitise, butcher or delete it depending on how far off the mark it is.

nohwnd:
@mikeclayton looks great. Did some [very minor rewording, and formatting](https://github.com/pester/Pester/wiki/Testing-different-module-types/_compare/ef6280289f5a6452ff69b8f49f107d701b6cfd5f...bd544293ba18bf79a730615dc7d417f7b4e26743). 👍

mikeclayton:
@nohwnd - I've made a couple more [grammatical tweaks](https://github.com/pester/Pester/wiki/Testing-different-module-types/_compare/bd544293ba18bf79a730615dc7d417f7b4e26743...d4358e561d6f100292c5338c2185c20139ed8f9a), and I've fixed (and tested) the Dynamic module code sample so it actually works now!

All that's left is to reference the article in the error message in InModuleScope.ps1. What's the best way to do that? Close this PR and open a new on? Or do you have something else in mind?

nohwnd:
@mikeclayton yes please. Create a new one, and I will mention it here and close this one.

alx9r:
@mikeclayton That wiki page looks pretty great.  :+1:

Issue #846, opened 2017-08-31
Calls to Write-Progress add significant performance overhead


Remco-74:
Write-Progress is called twice for every 'It' (before and after execution). The calls to Write-Progress add significant performance overhead to every testcase. 

**Results with Write-Progress as default (executed 4 times):**

    Tests completed in 76.82s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ....
    Tests completed in 64.22s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 63.29s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 64.63s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0

**Results with Write-Progress disabled by setting $ProgressPreference (executed 4 times):**

    Tests completed in 52.55s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 53.02s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 53.19s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 52.04s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0

**Results without Write-Progress (commented out)  (executed 4 times):**

    Tests completed in 52.95s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 52.01s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 52.03s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0
    ...
    Tests completed in 52.58s
    Tests Passed: 871, Failed: 0, Skipped: 0, Pending: 15, Inconclusive: 0

These timings are based on the exact same testset. I used 

> Get-Module | Remove-Module -Force

Before the first call to be sure that the correct code is loaded and to prevent caching issues.

**Results:** 
[...]

PR #938, opened 2017-11-12
Remove progress to make execution faster and fix linux formatting errors

nohwnd:
workaround for #846 

As requested by #321 we added progress reporting, but it adds a lot of overhead and breaks the output on Linux. I am removing it for now, so we can come up with a better solution that would also work for build servers.


nohwnd:
fixed #846
moves to #939

PR #941, opened 2017-11-12
Update New-Fixture help and add stripping .ps1 from the Name parameter

### LINKED ISSUE #369: New-Fixture creates code that can cause a StackOverflowException because it uses case sensitive matching
I noticed something interesting with the code created by New-Fixture. It adds a line like this:

``` powershell
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path).Replace(".Tests.", ".")
```

This creates a problem when you use Invoke-Pester, if you type the test name using ".tests" instead of ".Tests" (note the case) it ends up continually re-running $sut as it doesn't replace ".tests" as it would if it was typed as ".Tests". This eventually causes the script to fail with a SOE.

The $sut code created by New-Fixture should replace the $sut code with a method that does a replace in a case-insensitive manner (unfortunately string.Replace doesn't have an overload that supports StringComparison so Regex.Replace may be the best option here).

it-praktyk:
Fix #369 

Additionally, the extension 'ps1' will be stripped from a provided name to don't duplicate it

### DIFF
--- a/Functions/New-Fixture.Tests.ps1
+++ b/Functions/New-Fixture.Tests.ps1
@@ -69,6 +69,18 @@ Describe "New-Fixture" {
             Join-Path -Path "$path" -ChildPath "$name.ps1" | Should -Exist
             Join-Path -Path "$path" -ChildPath "$name.Tests.ps1" | Should -Exist
         }
+        It "Creates fixture if Path is set to '(pwd)' and Name contains the 'ps1' extension:" {
+            $name = "Relative4-Fixture.ps1"
+            $nameWithoutExtension = $name.Substring(0,$($name.Length)-4)
+            $path = "TestDrive:\"
+
+            Push-Location -Path $path
+            New-Fixture -Name $name -Path (Get-Location) | Out-Null
+            Pop-Location
+
+            Join-Path -Path "$path" -ChildPath "$nameWithoutExtension.ps1" | Should -Exist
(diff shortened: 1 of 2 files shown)


nohwnd:
@it-praktyk the current code uses `-replace` which already is case insensitive. The dots in the `.Tests.` already are escaped, so the only real change is removing `.ps1` from name. So what issue is this PR solving?

it-praktyk:
Haha,
I know how I "confirmed" that bug still exists.

```
PS >Get-ChildItem


    Directory: <FOLDER_PATH>\test


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       13.11.2017     22:14             26 after.ps1
-a----       13.11.2017     22:14            275 after.Tests.ps1
-a----       13.11.2017     22:06             27 before.ps1
-a----       13.11.2017     22:06            260 before.tests.ps1


PS >Get-ChildItem | foreach { ( Split-Path -Leaf $_).Replace(".Tests.",".")}
after.ps1
after.ps1
before.ps1
before.tests.ps1
PS >
```

The part of code ```.Replace(".Tests.",".")}``` - by mistake - I took from an example.

I also played with some other files with names containing 'tests.tests.ps1'. Yeap, I was too tired.
I'll update an example and revert the real code soon.

it-praktyk:
I changed labels. Now it's not a bug but a documented feature ;-)

Issue #937, opened 2017-11-12
Tests from Gherkin.Tests.ps1 fail on Linux


it-praktyk:
I enabled executing test from the file Functions\Gherkin.Tests.ps1 (I commented the [line](https://github.com/pester/Pester/blob/2c1f24b855014ad1e6468e19009fbba930cfc1cc/Functions/Gherkin.Tests.ps1#L3)) and after that run tests for Gherkin - all tests fail.

```
PS > Invoke-Pester -Script ./Functions/Gherkin.Tests.ps1
Executing all tests in './Functions/Gherkin.Tests.ps1'

Executing script ./Functions/Gherkin.Tests.ps1

  Describing Invoke-Gherkin
WARNING: OS Information retrieval is not possible, reports will contain only partial
 system data
    [-] Works on the Validator example 3.31s                                               Expected: {13}                                                                       But was:  {0}                                                                        31:         $gherkin.Results.PassedCount | Should -Be $gherkin.Results.TotalCount
      at Invoke-Assertion, <FOLDER_PATH>/Pester/Functions/Assertions/Should.ps1: line 209
      at <ScriptBlock>, <FOLDER_PATH>/Pester/Functions/Gherkin.Tests.ps1: line 31
    [-] Supports testing only scenarios with certain tags 24ms
      Expected: {3}
      But was:  {0}
      35:         $gherkin.Mockery.PassedCount | Should -Be $gherkin.Mockery.TotalCount
      at Invoke-Assertion, <FOLDER_PATH>/Pester/Functions/Assertions/Should.ps1: line 209
      at <ScriptBlock>, <FOLDER_PATH>/Pester/Functions/Gherkin.Tests.ps1: line 35
    [-] Supports tagging examples 32ms
      Expected: {4}
      But was:  {0}
      40:         $gherkin.Example1.PassedCount | Should -Be $gherkin.Example1.TotalCount
      at Invoke-Assertion, <FOLDER_PATH>/Pester/Functions/Assertions/Should.ps1: line 209
      at <ScriptBlock>, <FOLDER_PATH>/Pester/Functions/Gherkin.Tests.ps1: line 40
    [-] Supports excluding scenarios by tag 24ms
      Expected: {10}
      But was:  {0}
      50:         $gherkin.NotMockery.PassedCount | Should -Be 10
      at Invoke-Assertion, <FOLDER_PATH>/Pester/Functions/Assertions/Should.ps1: line 209
[...]

PR #942, opened 2017-11-13
Fix Gherkin for Linux (#937) and PS2

Jaykul:
Fix the linux case-sensitive problem causing #937

### DIFF
--- a/Examples/Gherkin/Gherkin-Scope.feature
+++ b/Examples/Gherkin/Gherkin-Scope.feature
@@ -27,5 +27,5 @@ Feature: Pester's Gherkin implementation test scope
         And the variable One should be "Uno"
 
     Scenario: Scope Should Not Bleed
-        Then the variable Script:Two should be "Dos"
+        Then the variable Script:Two should not exist
         And the variable One should not exist
--- a/Examples/Gherkin/Gherkin-Scope.feature
+++ b/Examples/Gherkin/Gherkin-Scope.feature
@@ -13,7 +13,6 @@ Feature: Pester's Gherkin implementation test scope
         When I set the variable Script:Two to "Hello World"
         Then the variable Script:Two should be "Hello World"
 
-
     Scenario: Scope Should Be Easy Part 1
         Given I initialize variables One and Script:Two to "Uno" and "Dos"
         When I set the variable One to "Hello World"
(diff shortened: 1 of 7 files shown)


Jaykul:
Incidentally, I didn't find a solution for testing PS2, and I don't see one on the CI server, so I'm not sure if we have any back-compat problems?

dlwyatt:
V2 testing happens here: https://build.powershell.org/viewType.html?buildTypeId=Pester_TestPesterOnPowerShellV2

CI server is happily licensed again.

it-praktyk:
I've tested PowerShell v.2 compatibility using Vagrant box made by @ferventcoder what you can download from https://app.vagrantup.com/ferventcoder/boxes/win2008r2-x64-nocm.

it-praktyk:
@dlwyatt, currently to check build results login to the TeamCity server at https://build.powershell.org is needed. Is it by design?

nohwnd:
@it-praktyk it aleays has been like that. Just click login as a guest.

nohwnd:
@Jaykul It still fails on 8 errors. :/

Jaykul:
The tests for Gherkin.Hook are legitimately failing just because the *tests* don't work in PS2

I had a look at those last night, rewrote them all -- the current tests are using reflection on the output of Get-Command and they say the "IsMandatory" property doesn't exist in PS2.  I just threw out the reflection and wrote some additional tests to validate the functionality instead. It was basically covered by other tests, but this way we'll capture exactly what's going on.

The other three failures are only failing intermittently, and I can't tell why from the logs. Hopefully they'll pass when I push this fix for the other tests, but either way, I will set up a VM to try to reproduce that failure. 

One possibility I can check (and/or put in a flag for) is the .NET Framework version -- the Gherkin assembly is compiled against .Net 3.5 (it's a *mandatory* upgrade for support, and was pre-installed on the only OSes left in support, as far as I know), but hypothetically you could be testing against something older.  Pester has never specified it's .NET Framework version -- I just made the assumption that we could require the OS be in support ;-)

Jaykul:
I have to give up on this for now. I'll get back to it in 8 hours or so, after work.

The remaining failures appear to be because you made the tests use StrictMode in PS2. 
They pass in PS3+, including on Linux.

Jaykul:
😅 😌 Huzzah! 🎉 🍾 🎆

nohwnd:
@Jaykul I totally see someone will post a PR that will say `Clarify meaning of BAF in the code` :D 

And thanks, will release soon.

PR #948, opened 2017-11-19
Don't export the function Invoke-GherkinStep

it-praktyk:
Based on the @Jaykul [comment](https://github.com/pester/Pester/issues/808#issuecomment-338449227) the function Invoke-GherkinStep don't need to be exported from Pester.

Update
In the initial version, I proposed to don't export the function Find-GherkinStep what was my mistake


it-praktyk:
The related issue #648

it-praktyk:
I've realized that in my initial commit I make mistake I proposed to don't export the wrong function. I correct my fault now.

PR #954, opened 2017-12-09
Improve Should -Throw

### LINKED ISSUE #279: Should Throw - Needs more flexibility/power
Asserting on the expected message in the exception is nice:

``` Powershell
{ throw 'foo' } | Should Throw 'foo'
```

But what if I wanted to examine additional properties of the exception, or even the exception type? I'm forced to do this:

``` Powershell
$caughtException = $null
try {
    throw (new-object system.notimplementedexception)
} catch {
    $caughtException = $_
}

$caughtException | Should Not Be $null
$caughtException.Exception.GetType() | Should Be ([System.NotImplementedException])
$caughtException.Exception.Message | Should ...
$caughtException.Exception.CustomProperty | Should ...

```

Proposed Syntax

``` Powershell
$action = { throw (new-object system.notimplementedexception) }
$actualException = $action | Should Throw ([System.NotImplementedException]) "My Message" -PassThru
$actualException.CustomProperty | Should ...
```

### LINKED ISSUE #953: Should Not Throw filters suppresses exceptions when they don't match filter
Specifying filters on `Should Not Throw` suppresses exception when it does not match the message. This does not make sense. The prerequisite for `Should Not Throw` is that no exception was thrown, the same way the pre requisite for `Should Throw` is that an exception was thrown. 

```powershell
get-module pester | Remove-Module ; Import-Module pester # -RequiredVersion 3.4.6

describe "a" { it "i" {
    try {
        { throw "asdf" } | Should Not Throw "hello"
        Write-Host "this should not be reached"
    } catch { Write-Host "this should throw from the assertion ($_)" }

    try {
        { throw "asdf" } | Should Not Throw
        Write-Host "this should not be reached"
    } catch { Write-Host "this should throw from the assertion ($_)" }
} }

# output  
# this should not be reached
# this should throw from the assertion (Expected: Expected: the expression not to throw an exception...
```

nohwnd:
Add `ExceptionType` filter
Work in progress, just want to see if it builds

- Add type to the error messages
- Improve the assertion messages, now they don't state what is actuall different

- Fix the way Axioms are loaded
Fix #279 
Fix #953


.travis.yml
```diff
@@ -22,4 +22,4 @@ matrix:
             - powershell
 
 script:
-  - pwsh -c 'Import-Module ./Pester.psd1; Invoke-Pester -ExcludeTag VersionChecks -EnableExit'
+  - pwsh -c 'Import-Module ./Pester.psd1; Invoke-Pester -ExcludeTag VersionChecks, StyleRules -EnableExit'
```
it-praktyk: This will have effect only for builds on Linux and macOS.

nohwnd: I know.

nohwnd:
Related #514

PR #945, opened 2017-11-16
Update Get-ScriptModule error message with link to wiki

mikeclayton:
A documentation-based fix to issue https://github.com/pester/Pester/issues/933 instead of the code-based fix in pull request https://github.com/pester/Pester/pull/934.

Adds a link to wiki documentation at https://github.com/pester/Pester/wiki/Testing-different-module-types for the error message thrown when trying to use Mock and InModuleScope commands on non-Script modules.

Note that merging this PR will mean that the wiki page will be referred to in Pester error messages, so the url will need to kept alive as long as the relevant versions of Pester are supported.

Issue #720, opened 2017-01-31
Move help to PlatyPS and generate pages of wiki based on it too


nohwnd:
https://github.com/PowerShell/platyPS

it-praktyk:
The related issue #648 

The current state of help - after merging currently opened pull requests.

![pesterv4-help-20170131_-_2017-01-31_18 25 34](https://cloud.githubusercontent.com/assets/4910261/22476521/074f7f94-e7e3-11e6-9de5-7077f44b00ba.png)

Jaykul:
So ... if we think we want to do this, I think we should ...

## Just Do It!

```posh
New-MarkdownHelp -Module Pester -OutputFolder Docs -WithModulePage
```

And then ask people to start helping us fix them.

vors:
Would be great to see pester help in platyPS. Let me know if you find anything blocking transition.

Jaykul:
Next step, clearly -- get people to help write help ;-)

it-praktyk:
The related issue #555

it-praktyk:
The related issue: [Would you like to include API documentation to your build?]( https://github.com/pester/Pester/issues/407).

it-praktyk:
The related series of blog posts [Automated Documentation in a CI/CD Pipeline for PowerShell Modules with PlatyPS, psake, AppVeyor, GitHub and ReadTheDocs](https://get-powershellblog.blogspot.com/2017/03/write-faq-n-manual-part1.html) by @markekraus

nohwnd:
Anyone interested in implementing this or should we keep it as is and close it?

it-praktyk:
I'll try to do that.

The related issue
https://github.com/PowerShell/platyPS/issues/371

Wojciech Sciesinski

________________________________

nohwnd:
halbarad (discord) APP [1:04 PM]
@PrzemyslawKlys @nohwnd, this is how I add it in:

```powershell
#.ExternalHelp VMConfig-help.xml
Function Start-VmcLab {
```

There's another way using an attribute on the function but I always forget what it is.
This is how I generate that xml from the md files.

```powershell
New-ExternalHelp -Path $(Build.SourcesDirectory)\Docs -OutputPath ` 
    $(Build.SourcesDirectory)\Output\en
```

PrzemyslawKlys:
I actually generate Docs folder for all my projects:

```
[...]

PR #955, opened 2017-12-11
Tests for comment based help

it-praktyk:
The related issues
- #648 
- #720
- #808

Replaces the pull request #667.

We can exclude running this test by CI by adding an exclusion in .travis.yml. I don't do it now to avoid [a conflict with #945](https://github.com/pester/Pester/pull/954/commits#diff-354f30a63fb0907d4ad57269548329e3R25).


nohwnd:
@it-praktyk most of the setup needs to be done on the server, if we need to skip tests during normal build. You can add it to travis.yml, I can solve conflicts :D

it-praktyk:
As you can see in detailed tests results for the commit https://github.com/pester/Pester/pull/955/commits/982d8901bbf9ec54df071ae9b9871c4096f7a213 only tests for really missed help elements failed - I hope that it's good reason to merge the current pull request.
The next commit https://github.com/pester/Pester/pull/955/commits/25bceddbe443f985559d0876815e836ff278dbd0 excludes those tests to run under non-releases CI pipelines.

@Jaykul, I kindly remind you about missed examples for Find-GherkinStep and GherkinStep functions. Thank you!

PR #964, opened 2017-12-20
Correct Write-PesterStart - Fix #935

### LINKED ISSUE #935: Test 'uses Format-PesterPath with the provided path' fails when called with -Show Failed
Invoking Pester tests (on macos) by ` invoke-pester -show failed, summary` makes the test fail. Most likely because we are depending on global Pester state and that says that we should not show that info, and so the formatting function is not called. 

To reproduce run the the tests with the given command, and then run with the full outupt, there should be one error less in the full output. 

```
[-] uses Format-PesterPath with the provided path 42ms                                                                                     Expected Format-PesterPath in module Pester to be called at least 1 times but was called 0 times                                         180:             Assert-MockCalled Format-PesterPath -ParameterFilter {$Path -eq $expected}                                              at <ScriptBlock>, /Users/nohwnd/projects/pester_main/Functions/Output.Tests.ps1: line 180
```

it-praktyk:


it-praktyk:
I've based my "fix" on https://github.com/it-praktyk/Pester/blob/67f5b658aa906cd1fe18169834c4af25854d60ae/Functions/Output.ps1#L248

nohwnd:
Then I am probably doing it wrong as well :) Will have a look. Ideally we would only use a single one in the condition.

Issue #974, opened 2018-01-18
AfterAll synopsis is incorrectly stating it runs at end of every It


bgelens:
## 1. Provide a general summary of the issue in the Title above

Current AfterAll synopsis

```
    Defines a series of steps to perform at the end of every It block within
    the current Context or Describe block.
```

Should be

```
    Defines a series of steps to perform at the end of the current Context
    or Describe block.
```

## 2. Describe Your Environment

Operating System, Pester version and PowerShell version:

```powershell
Pester version     : 4.1.0 C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.0\Pester.psm1
PowerShell version : 5.1.16299.98
OS version         : Microsoft Windows NT 10.0.16299.0
```

Fixed in #975

PR #975, opened 2018-01-18
Fix AfterAll synopsis for issue #974

bgelens:
Current AfterAll synopsis

```
    Defines a series of steps to perform at the end of every It block within
    the current Context or Describe block.
```

Should be

```
    Defines a series of steps to perform at the end of the current Context
    or Describe block.
```


it-praktyk:
LGTM

PR #976, opened 2018-01-19
Make getting started link point to wiki

niroyb:


nohwnd:
Not sure if I should merge this, I really want to write that article, I just don't have the time :/

PR #994, opened 2018-02-19
#808 add GherkinStep examples

KevinMarquette:


it-praktyk:
Thank you, I really appreciate that.

PR #1001, opened 2018-02-25
Fix GitHub issue template by using the -ListAvailable option on Get-Module when Pester is not already loaded

bergmeister:
PowerShell does lazy loading, i.e. if `Invoke-Pester` has not been executed yet, then `Get-Module Pester` does not return anything. Therefore the `-ListAvailable` option has to be used when executing the command in a new shell.


it-praktyk:
"Failed" tests are probably caused by changes described in the blog post [Weak cryptographic standards removal notice](https://githubengineering.com/crypto-removal-notice/).

PR #1013, opened 2018-03-25
JaCoCo report always mentions only 1 file covered instead of the actual number

Thilas:
When generating a JaCoCo report along with code coverage, I noticed that the CLASS counter was not correct: if more than 1 file is actually covered, it always show `1`. The only exception is when there are no covered files, then it properly shows 0.

The issue comes from the way `$hitFiles` is determined using `Measure-Object`. Instead, I suggest a slightly different way similar to how other counter are computed. I adapted an existing test to highlight this issue.

### DIFF
--- a/Functions/Coverage.Tests.ps1
+++ b/Functions/Coverage.Tests.ps1
@@ -33,29 +33,37 @@ InModuleScope Pester {
 
 '@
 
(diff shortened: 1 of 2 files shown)


it-praktyk:
@Thilas, you can't. It's internal error of the TeamCity instance.

@dlwyatt, can you handle it? Thank you.

PR #1019, opened 2018-04-08
Add space in DescribeImpl function.

sethvs:


it-praktyk:
@nohwnd, can we open an issue to remember about reviewing tests for the Describe function?

Why wasn't it discovered previously?

nohwnd:
Not sure if I follow, this is purely formatting issue, even though it does not look like one. In other words the code is syntactically correct and works. 

```powershell
$Pester = @{ TestNameFilter = $true}
$Pester.TestNameFilter-and -not $false

# returns 
# $true
```

nohwnd:
@it-praktyk you need to join the powershell one-liner contest next year :)

PR #1026, opened 2018-04-13
Skipping test from inside the It block

### LINKED ISSUE #1022: Skip a test from the inside of the It script block

My usecase is for environmental validation, not unit tests as such. I have been recently working on dbachecks project trying to find an efficient way of structuring code while making sure we can unit test environmental tests. One of the remaining challanges is that to keep code simple and any logic testable I don't want to depend on the `-Skip` parameter of `It`. 

Is there a way, or do you see it possible, to skip the test from inside the `It` script block similarily to how you can now mark the test as inconclusive? 

## 3. Expected Behavior

```
@($testCases).ForEach{
  It "My Test" {
     if ($psitem.MyProperty1 -eq $true) { Set-TestSkipped } else {
        $psitem.MyProperty2 | Should -Be 42 -Bdcause "it is the reason behind everything"
     }
  }
}
```

or similarily using test cases 
```
It "My Test" -TestCases $testCases {
   param($myProperty1, $myProperty2)
   if ($myProperty1 -eq $true) { Set-TestSkipped } else {
      $myProperty2 | Should -Be 42 -Bdcause "it is the reason behind everything"
   }
}
```

I'd rather not use the Inconlusive output, as to me it is very different status. Inconclusive is a test which cannot be asserted at a time for some reason, skipped is something that shouldn't even be asserted.

michalporeba:
Resolves #1022 

image

at the moment all but 5 unit tests pass, all five in "Describing Pester manifest and changelog". What do I need to do to make them pass?


michalporeba:
OK, so the tests that failed locally, those relating to changlog and versioning passed in travis, but publish to GitHub failed in TeamCity. I cannot see why (no access to TC) so I will need some help. 

Few things remaining potentially for discussion, that could be included in this PR quite easily are 
1. Changing `Set-TestInconclusive` for it to internally use the new `Set-ItResult`
2. Adding a deprecation message to `Set-TestInconclusive`
3. Implementing `Set-ItResult -Pending` to explicitly set It result to pending from inside the block. I don't know if there is much value for it, there isn't for me, but I thought I'll bring it up as it would be easy to do it, and may make sense for 'completeness' of the solution

Any thought? Any recommended changes to the current implementation? I will welcome any reviews. Can anybody from the Pester crew comment on how do you see it? How likely is this PR to make it into Pester soonish? @nohwnd, anybody else? I'm working on big changes to dbachecks which I want to push out of the door this month, and if only possible, I'd like to use this new functionality already. 

Thanks your your time and help (and I do appreciate there is PSConfEU coming up very soon, so many people are rather busy).

nohwnd:
> no access to TC
Just click login as guest, the build is public. It's not very obvious sorry.

1, 2. yeah that would be nice, Set-TestInconclusive should be deprecated in that case, and we need to offer a strategy for replacement.
3. Yes please, it only makes sense to add the function if it supports the complete set of states.

I am sorry but I cannot review your code at the moment, I am also preparing for the psconfeu and need to finish my talks that before comitting to someting else. In general the changes in this repo are merged rather slowly, especially if they concern public apis. 

 Are you attending the conf?:)

michalporeba:
OK, I'll work something in for 1,2 and 3.
I realise a lot of people are on the conf, but not me. I've just got back
home. And anyway, I'm new to the whole powershell thing, so I only have
just learnt about the conference 2 weeks ago ;)
I'll see, if I learn some powershell this year, perhaps I'll come next year
;)

michalporeba:
The build error in TC appears to be to do with build configuration so somebody with the full access will have to investigate it. 
image

michalporeba:
OK, so now I guess I will need some help with that v2. `\Functions\Assertions\Set-TestInconclusive.Tests.ps1` is failing, complains the mock function was not called. Any ideas?

michalporeba:
Can anybody help with making that powershell 2 test pass? 
Also if somebody could review the code I'd be eager to make any adjustments over the weekend

michalporeba:
Thank you @it-praktyk. Those errors you see happen locally as, but they are only to do with the release process (CI) and they pass in TravicCI and TeamCity. 

The failing test can be found in the TeamCity if you log in as guest
image
The problem is in only one of unit tests, not a very important one too. It would be easy to simply remove it, but I'd like to understand and fix it. I could probably remove it just to see if the build will pass without it.

michalporeba:
Thank you @nohwnd. I didn't realise you can filter like that on Assert-MockCalled. That's cool. 
Could you explain the changes? You moved the mock inside It and you moved the parameter filtering from the mock to assertion. Why would one (or both) of them be needed to fix it? I'm just trying to understand things better.

nohwnd:
@michalporeba I don't think either was needed to fix it, I was just trying it out so I don't have to spin up a powershell 2 machine. :) 

My reasoning is this: 

- Mock should be defined as close to the usage as possible, the less shared things there are among tests the better.

- Catching all calls and then counting the calls you wanted is easier to comprehend (and safer) than only mocking the calls that you expect and then counting them. 

I don't think skipping the deprecation test is way to go forward. There is something wrong with that test, so I will have to bite the bullet and install v2 somewhere :D

michalporeba:
@nohwnd I don't know what has changed but suddenly both tests show as passing. Can this be merged, please?

nohwnd:
@michalporeba thanks for the PR and sorry about the long silence I was taking a little break. I like the changes, but I don't like adding the explanation to the result 'is skipped without an explanation, consider adding -Because parameter to the Set-ItResult call', that will get old really fast, and sometimes you don't want to provide a reason (maybe because you are skipping the test programatically). The because parameter is optional so it should not be forcing the user to provide the reason.

fourpastmidnight:
> OK, so the tests that failed locally, those relating to changlog and versioning passed in travis, but publish to GitHub failed in TeamCity. I cannot see why (no access to TC) so I will need some help.
> 
> Few things remaining potentially for discussion, that could be included in this PR quite easily are
> 
> 1. Changing `Set-TestInconclusive` for it to internally use the new `Set-ItResult`
> 2. Adding a deprecation message to `Set-TestInconclusive`
> 3. Implementing `Set-ItResult -Pending` to explicitly set It result to pending from inside the block. I don't know if there is much value for it, there isn't for me, but I thought I'll bring it up as it would be easy to do it, and may make sense for 'completeness' of the solution
> 
> Any thought? Any recommended changes to the current implementation? I will welcome any reviews. Can anybody from the Pester crew comment on how do you see it? How likely is this PR to make it into Pester soonish? @nohwnd, anybody else? I'm working on big changes to dbachecks which I want to push out of the door this month, and if only possible, I'd like to use this new functionality already.
> 
> Thanks your your time and help (and I do appreciate there is PSConfEU coming up very soon, so many people are rather busy).

I actually came across this today, and I submitted a PR #1141 to fix this.

PR #1040, opened 2018-05-06
The test for nonexisting module behavior check

### LINKED ISSUE #949: Get-ScriptModule error when no modules with given name imported
```powershell
Pester version     : 4.0.8 C:\src\github\pester\pester\Pester.psm1
PowerShell version : 5.1.15063.726
```

When attempting to Mock a command in a module which is not loaded into scope, Get-ScriptModule throws an unclear error message:

```
 "Module 'MyNonExistentModule' is not a Script module.  Detected modules of the following types: ''".
```

Get-ScriptModule appears to be expecting PowerShell's built-in Get-Module to throw a terminating error if no modules match the name.

It attempts to ```catch``` the error, but it looks like Get-Module actually returns a $null object instead (which might be new behaviour in recent versions of PowerShell) so the ```catch``` block never runs..

One way to fix this is to add the following immediately after the ```try { ... } catch { ... }``` block in Get-ScriptModule:

```powershell
if ($modules.Count -eq 0)
{
    throw "No module named '$ModuleName' is currently loaded."
}
```

Here's some tests to validate current and proposed behaviour:

```powershell
Describe "Mock command" {

    Context "When attempting to mock a command in a non-existent module" {

        It "Current behaviour" {
            {
                Mock -CommandName "Invoke-MyMethod" `
                     -ModuleName  "MyNonExistentModule"
                     -MockWith    { write-host "my mock called!" };
            } | Should Throw "Module 'MyNonExistentModule' is not a Script module.  Detected modules of the following types: ''";
        }

        It "Proposed behaviour" {
            {
                Mock -CommandName "Invoke-MyMethod" `
                     -ModuleName  "MyNonExistentModule"
                     -MockWith    { write-host "my mock called!" };
            } | Should Throw "No module named 'MyNonExistentModule' is currently loaded.";
        }

    }

}
```

I can send a PR if this looks like a valid fix.

it-praktyk:
An additional tests based on @mikeclayton's [proposal](https://github.com/pester/Pester/issues/949#issue-275864940).

Fix #949

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -333,11 +333,12 @@ Describe "When calling Mock on non-existing function" {
 Describe "When calling Mock on non existent module" {
 
     It "Should throw correct error" {
-        {
-            Mock -CommandName "Invoke-MyMethod" `
-                    -ModuleName  "MyNonExistentModule"
-                    -MockWith    { write-host "my mock called!" };
-        } | Should -Throw "No module named 'MyNonExistentModule' is currently loaded.";
+        $params = @{
+            CommandName = 'Invoke-MyMethod'
+            ModuleName = 'MyNonExistentModule'
+        }
+
+        { Mock @params -MockWith { write-host "my mock called!" } } | Should -Throw "No module named 'MyNonExistentModule' is currently loaded."
     }
 
 }
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -330,6 +330,18 @@ Describe "When calling Mock on non-existing function" {
     }
 }
 
+Describe "When calling Mock on non existent module" {
+
+    It "Should throw correct error" {
+        {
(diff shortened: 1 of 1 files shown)


nohwnd:
@it-praktyk it fails with scriptstacktrace not present on an object, this is similar error like in the PR that allows providing scripts in memory not as files. I am guessing the error comes from [here](https://github.com/pester/Pester/blob/master/Pester.psm1#L859) but I did not have chance to replicate it. If you don't have a system with ps2 installed I can fix it for you, because I will be fixing the other one as well.

nohwnd:
@it-praktyk so what was the error that you fixed in your second commit? I am just trying to understand why it failed before and now it does not fail anymore. Does any error that occurs in posh v2 end with the stacktraceerror?

it-praktyk:
screenshot from 2018-05-09 09-02-03

Lack of a backtick caused that test failed under PowerShell 2.0.

Additionally, I splatted some parameters to decrease a line length.

Issue #960, opened 2017-12-17
Documentation update required due to new assertion added by #959


it-praktyk:
In the pull request #959 new assertions were added.

Quotation from the pull request description

> Add -Because parameters to all assertions Fixes #312
> Add -BeLessOrEqual and -BeGreaterOrEqual
> Add -Contain (that operates on arrays) Fixes #121
> Add -BeLikeExactly
> Add -HaveType alias to -BeOfType
> Fix assertion messages in -BeOfType Fixes #729
> Throw argument exception when -BeOfType is given type that is not loaded
> Add -PassThru to -Throw to get the exception when some is thrown and passes the filters
> Add -BeTrue to test for truthy values
> Add -BeFalse to test for falsy values
> Add -HaveCount to count stuff in collections

An appropriate documentation needs to be aligned with the current code.
- [en-US/about_Should.help.txt](https://github.com/pester/Pester/blob/4a04a80b72a3099e6a7b56feccc390668d0a6e83/en-US/about_Should.help.txt)
- [Pester wiki - Should](https://github.com/pester/Pester/wiki/Should)

it-praktyk:
@nohwnd, what do you think about assigning it to the same milestone that code will be released?

nohwnd:
Sure, but don't start documenting it just yet, maybe some of the changes break compatibility, or will be rejected from the final release, so let's not waste time putting changes in that won't be needed in the end.

nohwnd:
The about should part was done in #990

bergmeister:
Why is the milestone for 4.4 when this already shipped in 4.2?

it-praktyk:
IMHO, [the wiki page](https://github.com/pester/Pester/wiki/Should) needs review too.

nohwnd:
For reference, this is how all the assertion operators can be easily listed import-module pester; & (get-module pester) { $AssertionOperators

ClaudioESSilva:
I have updated the Should page with missing parts 😄 
https://github.com/pester/Pester/wiki/Should

it-praktyk:
Yeap,
because I've found - in [your comments](https://github.com/pester/Pester/issues/960#issuecomment-388755764) something what should be added to the wiki.

BTW, can you check if is the syntax correct? 
```powershell
import-module pester; & (get-module pester) { $AssertionOperators
```

nohwnd:
[...]

PR #1046, opened 2018-05-15
Remove "Than" from GreaterOrEqual and LesOrEqual

ClaudioESSilva:
I have updated the Should page with missing parts and found this two typos on this file.
This is part of https://github.com/pester/Pester/issues/960


it-praktyk:
Thx @ClaudioESSilva!

PR #1059, opened 2018-06-08
improvements in TestDrive.ps1

EklipZgit:
Was debugging an issue where if the temp folder backing TestDrive: gets deleted mid-test-run (some bad DSC temp folder cleanup on our end) where

``` powershell
& $SafeCommands['New-PSDrive'] -Name $DriveName -PSProvider FileSystem -Root $directory -Scope Global -Description "Pester test drive"
```
still gets reached because Test-Path TestDrive:\ returns false, and throws because the TestDrive psdrive still being registered. 
During debugging I found the erroraction ignore on Remove-PSDrive and thought that might be swallowing the error leaving testdrive behind. Turned out not to be the case, but is still a bug regardless in my opinion, unless there was some explicit reasoning for ignoring errors. 

Also removed unnecessary hashtable calls for 'out-null' when $null = is more performant and more explicit, and removed the unnecessary regex checking for non-whitespace when a .net call to [string]::IsNullOrWhiteSpace() is more performant.

### DIFF
--- a/Functions/TestDrive.ps1
+++ b/Functions/TestDrive.ps1
@@ -1,6 +1,6 @@
 #
 function New-TestDrive ([Switch]$PassThru, [string] $Path) {
-    if ([string]::IsNullOrWhiteSpace($Path))
+    if ($Path -notmatch '\S')
     {
         $directory = New-RandomTempDirectory
     }
--- a/Functions/TestDrive.ps1
+++ b/Functions/TestDrive.ps1
@@ -1,6 +1,6 @@
 #
 function New-TestDrive ([Switch]$PassThru, [string] $Path) {
-    if ($Path -notmatch '\S')
+    if ([string]::IsNullOrWhiteSpace($Path))
     {
         $directory = New-RandomTempDirectory
     }
@@ -8,7 +8,7 @@ function New-TestDrive ([Switch]$PassThru, [string] $Path) {
     {
         if (-not (& $SafeCommands['Test-Path'] -Path $Path))
         {
-            & $SafeCommands['New-Item'] -ItemType Container -Path $Path | & $SafeCommands['Out-Null']
+            $null = & $SafeCommands['New-Item'] -ItemType Container -Path $Path
         }
 
         $directory = & $SafeCommands['Get-Item'] $Path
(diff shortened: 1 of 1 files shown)


EklipZgit:
@it-praktyk which part makes it unuseable on powershell 2.0?

it-praktyk:
> which part makes it unuseable on powershell 2.0?

With all changes from you commit all tests from the file [TestDrive.Tests.ps1](https://github.com/pester/Pester/blob/master/Functions/TestDrive.Tests.ps1) fail.

When I revert change [in line 3](https://github.com/pester/Pester/pull/1059/commits/8e60f4b97f60a8c3deddb8e520880c013b25d9f1#diff-8c4641e1a8e49f4408c73f38c57ed69dL3) all tests pass.

I test code for PowerShell compatibility using Vagrant box [ferventcoder/win2008r2-x64-nocm](https://app.vagrantup.com/ferventcoder/boxes/win2008r2-x64-nocm).

```
PS > $psversiontable

Name                           Value
----                           -----
CLRVersion                     2.0.50727.5420
BuildVersion                   6.1.7601.17514
PSVersion                      2.0
WSManStackVersion              2.0
PSCompatibleVersions           {1.0, 2.0}
SerializationVersion           1.1.0.1
PSRemotingProtocolVersion      2.1
```

When I check 

```
PS > [String].GetMethod() | Where-Object { $_.name -match 'empty' }

Name                       : IsNullOrEmpty
DeclaringType              : System.String
ReflectedType              : System.String
MemberType                 : Method
MetadataToken              : 100663628
Module                     : CommonLanguageRuntimeLibrary
MethodHandle               : System.RuntimeMethodHandle
Attributes                 : PrivateScope, Public, Static, HideBySig
CallingConvention          : Standard
ReturnType                 : System.Boolean
ReturnTypeCustomAttributes : Boolean
ReturnParameter            : Boolean
IsGenericMethod            : False
IsGenericMethodDefinition  : False
ContainsGenericParameters  : False
IsPublic                   : True
IsPrivate                  : False
IsFamily                   : False
IsAssembly                 : False
IsFamilyAndAssembly        : False
IsFamilyOrAssembly         : False
IsStatic                   : True
IsFinal                    : False
IsVirtual                  : False
IsHideBySig                : True
IsAbstract                 : False
IsSpecialName              : False
IsConstructor              : False
```

When I check a documentation for the [String.IsNullOrWhiteSpace(String) Method](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=netstandard-2.0) I see that it should be available but it's not.

EklipZgit:
I have reverted that line in particular, builds are passing. Please take another look @it-praktyk 

Thanks!

nohwnd:
Thanks for the PR :)

Issue #1060, opened 2018-06-11
Invoke-Pester -PassThru returns pipeline output from test scripts


mikeclayton:
## 1. Provide a general summary of the issue in the Title above

When calling Invoke-Pester with the -PassThru switch, the return value contains all of the pipeline output from the test scripts, as well as the actual test results.

## 2. Describe Your Environment

Operating System, Pester version and PowerShell version:

```
Pester version     : 3.4.0 C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\Pester.psd1
PowerShell version : 5.1.16299.251
OS version         : Microsoft Windows NT 10.0.16299.0```
```

Note: I'm actually importing Pester 4.3.1 from a separate working directory, not the version 3.4.0 installed in the PowerShell modules folder.

## 3. Expected Behavior

I'm not sure if this is actually a bug or not, but my expected behaviour would be for Pester to swallow any pipeline output from the test scripts and just return its test results.

## 4.Current Behavior

Create a line-of-business script which sends values to the pipeline when executed. I try to avoid doing this myself, but I've inherited a script which already does it :-(

For example:

**Invoke-MyFunction.ps1**
```powershell
Import-Module -Name "C:\Windows\System32\WindowsPowerShell\v1.0\Modules\BitsTransfer\BitsTransfer.psd1" -PassThru;

Write-Output "my output";

function Invoke-MyFunction
{
    return $true;
}
```

Create a test script:

**Invoke-MyFunction.Tests.ps1**
```powershell
. .\Invoke-MyFunction.ps1;

Describe "MyFunction Tests" {
    It "Should not throw" {
        Invoke-MyFunction
    }
}
```

And a test runner script:

**Invoke-MyPesterTests.ps1**
```powershell
Import-Module -Name "C:\packages\Pester.4.3.1\tools\Pester.psd1" -ErrorAction "Stop";

$results = Invoke-Pester -Script ".\Invoke-MyFunction.Tests.ps1" -PassThru;

write-host "results :";
write-host ($results.GetType().FullName);
write-host ($results | fl * | out-string);
```

Then, run Invoke-MyPesterTests.ps1
```
c:\src\scratch\pester>powershell c:\src\scratch\pester\Invoke-MyPesterTests.ps1
Executing all tests in '.\Invoke-MyFunction.Tests.ps1'

Executing script .\Invoke-MyFunction.Tests.ps1

  Describing MyFunction Tests
    [+] Should call write-output 510ms
[...]

PR #1063, opened 2018-06-12
Fix for #1060 - Invoke-Pester -PassThru returns pipeline output from test scripts

mikeclayton:
Swallow objects that are written to the output pipeline inside test scripts so that they don't get returned by Invoke-Pester when the -PassThru switch is specified.

Also added some tests to validate the new behaviour.


mikeclayton:
Adding link to issue #1060

Issue #1100, opened 2018-09-10
Windows: Symlinks in TestDrive creates errors on teardown


Brekkjern:
## 1. Provide a general summary of the issue in the Title above

When you create a symlink in the TestDrive while in a Describe block, an error occurs and the test subsequently fails.

## 2. Describe Your Environment

Operating System, Pester version and PowerShell version:

Pester version     : 4.3.1 C:\Users\nkb.GLOBALINTS\Documents\WindowsPowerShell\Modules\Pester\Pester.psm1
PowerShell version : 5.1.17134.228
OS version         : Microsoft Windows NT 10.0.17134.0

## 3. Expected Behavior

TestDrive should be cleared as normal and test should pass.

## 4.Current Behavior

Teardown of TestDrive creates error and test fails.

```powershell
    [-] Error occurred in Context block 16ms
      Win32Exception: There is a mismatch between the tag specified in the request and the tag present in the reparse point
      at Clear-TestDrive, C:\Users\nkb.GLOBALINTS\Documents\WindowsPowerShell\Modules\Pester\Functions\TestDrive.ps1: line 40
      at DescribeImpl, C:\Users\nkb.GLOBALINTS\Documents\WindowsPowerShell\Modules\Pester\Functions\Describe.ps1: line 175
  [-] Error occurred in Describe block 18ms
    Win32Exception: There is a mismatch between the tag specified in the request and the tag present in the reparse point
    at Remove-TestDrive, C:\Users\nkb.GLOBALINTS\Documents\WindowsPowerShell\Modules\Pester\Functions\TestDrive.ps1: line 124
    at DescribeImpl, C:\Users\nkb.GLOBALINTS\Documents\WindowsPowerShell\Modules\Pester\Functions\Describe.ps1: line 171
```

## 5. Possible Solution

Check for symlinks in scoped TestDrive before doing teardowns.

## 6. Context

I am writing a script that will create symlinks and I need to test that it is creating the symlinks correctly. Mocking the method will not necessarily show correct behaviour.

Brekkjern:
I attempted to delete the symlinks myself at the end of each context block, but the errors still appear. As far as I can see the errors are similar.

mikeclayton:
I think this is a bug / feature in PowerShell's Remove-Item command when trying to delete a folder which contains symbolic links.

A simple out-of-band repro is:

```powershell
$root     = "C:\temp\root";
$source   = "$root\source";
$symlink  = "$root\symlink";

[...]

PR #1103, opened 2018-09-20
Added fix for deleting symlinks in TestDrive

mikeclayton:
see https://github.com/pester/Pester/issues/1100

### DIFF
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -182,6 +182,9 @@ InModuleScope Pester {
 
             { Clear-TestDrive } | Should -Not -Throw;
 
+            $count = @(Get-ChildItem -Path $root).Length;
+            $count | Should -Be 0;
+
         }
     }
 }
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -168,3 +168,20 @@ InModuleScope Pester {
         }
     }
 }
+
+InModuleScope Pester {
+    Describe "Clear-TestDrive" {
+        It "deletes symbolic links in TestDrive" {
+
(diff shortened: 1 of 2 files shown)


nohwnd:
@mikeclayton from the test results it seems that the `-Attributes` parameter is present only on PowerShell 5 and latest.

mikeclayton:
@nohwnd - Ah, not so good :-(. I saw the link to the broken build, but didn't think to click "Log in as guest" to see the build log!

I've changed it to use what I think are 2.0-compatible parameters - I'll check if the build works this time.

I also added ```-Recurse``` as it would only remove symlinks from the root of Testdrive:\ otherwise. I guess this could *potentially* cause a performance issue if tests create a lot of files and folders, but ```Clear-TestDrive``` already does this so hopefully that's not a major concern.

nohwnd:
@mikeclayton the previous issue was resolved, but still some remain. Looks like PowerShell 5 is the only one supporting symlinks directly.

mikeclayton:
@nohwnd - backward compatibility is hard :-S.

It looks like it's the setup in the test that's failing now - the old-school way of doing it seems to be to call "cmd /c mklink" according to https://stackoverflow.com/questions/894430/creating-hard-and-soft-links-using-powershell

Would that be an ok way to go, or is it likely to cause more problems that it solves?

(I'm not around for the next week, but I can pick this up again when I'm available again).

Cheers,

M

nohwnd:
@mikeclayton I implemented the test in a way that it works on powershell 2 as well as 5. But the problem is that it needs admin rights to run, so I commented the test for now. Later I will move the tests and set them up to run as admin (which we already do on the build server because your tests have passed few times). No action needed I will merge as is.

PR #1105, opened 2018-09-24
Make an attempt to get environment information for non-Windows platforms

JamesWTruher:
I'm definitely missing environment information when running Pester in our CI on Non-Windows platforms. I would love if the logs had that information. This is an attempt at providing that information with tools that will probably be there. I'm using `uname` and `id` to provide that info. If those tools aren't there, we won't be any worse off, but if they are there, then we can provide the test environment pretty well.

### DIFF
--- a/Functions/TestResults.ps1
+++ b/Functions/TestResults.ps1
@@ -454,7 +454,7 @@ function Get-RunTimeEnvironment() {
             }
             if ( $SafeCommands['id'] -ne $null )
             {
-                $userName = & $SafeCommands['id']
+                $userName = & $SafeCommands['id'] -un
             }
         }
         catch {
--- a/Functions/TestResults.ps1
+++ b/Functions/TestResults.ps1
@@ -429,6 +429,8 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
 }
 function Get-RunTimeEnvironment() {
     # based on what we found during startup, use the appropriate cmdlet
+    $computerName = $env:ComputerName
+    $userName = $env:Username
     if ( $SafeCommands['Get-CimInstance'] -ne $null )
     {
         $osSystemInformation = (& $SafeCommands['Get-CimInstance'] Win32_OperatingSystem)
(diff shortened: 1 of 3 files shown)


JamesWTruher:
@Jaykul @nohwnd - Could you take a look to see if I've taken an appropriate approach? Thanks!

nohwnd:
@JamesWTruher tested it on macOS and the only thing I found that the username is not just username but a whole user info. To print just username it should be `id -un`, which worked on macOS, but I haven't tested it on linux. 
```
Name                           Value
----                           -----
cwd                            /Users/nohwnd/projects/pester_main
user-domain
nunit-version                  2.5.8.0
clr-version                    Unknown
user                           uid=501(nohwnd) gid=20(staff) groups=20(staff...
platform                       Darwin
os-version                     17.7.0
machine-name                   nmbp
```

it-praktyk:
Tested on Linux. 

`id` returns

```
uid=1000(<USER_NAME>) gid=1000(<USER_NAME>) groups=1000(<USER_NAME>),10(wheel),18(dialout),970(vboxusers) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
``

File: Functions/TestResults.ps1 [hunk 1]
```diff
@@ -437,6 +439,28 @@ function Get-RunTimeEnvironment() {
...
+        try {
+            if ( $SafeCommands['uname'] -ne $null )
+            {
+                $osSystemInformation.Version = & $SafeCommands['uname'] -r
+                $osSystemInformation.Name = & $SafeCommands['uname'] -s
+                $computerName = & $SafeCommands['uname'] -n
+            }
+            if ( $SafeCommands['id'] -ne $null )
+            {
+                $userName = & $SafeCommands['id']
```
nohwnd: Please change this to `id -un`, and test it on Linux.

it-praktyk:
```
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<test-results xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="nunit_schema_2.5.xsd" name="Pester" total="982" errors="0" failures="5" not-run="0" inconclusive="0" ignored="0" skipped="0" invalid="0" date="2018-09-30" time="11:07:18">
  <environment machine-name="<COMPUTERNAME>>" os-version="4.18.9-200.fc28.x86_64" nunit-version="2.5.8.0" user="<USER_NAME>" cwd="/home/<USER_NAME>/Scripts/Pester" platform="Linux" user-domain="" clr-version="Unknown" />

OUTPUT_PARTIALLY OMITTED

```

nohwnd:
@it-praktyk what does that mean? That it works? Or that it does not work? More context please.

it-praktyk:
It works like expected now. Means returns only the username, the single word.
________________________________

JamesWTruher:
I was wondering about `id` vs `id -un`. I chose `id` because I thought more info would be better, but if it's really just about the username, then `id -un` matches intent better. 
Thanks for the quick look! 
@it-praktyk @nohwnd

nohwnd:
It’s out.

PR #1109, opened 2018-10-03
Check scope of Set-Variable to resolve #1107

### LINKED ISSUE #1107: Mocks Set-Variable, works only for the first occurrence
Hello,
I want to mock variable assignment using Set-Variable and I'm trying to do something like this

```powershell
function Test-Variable {
    Param()

    Set-Variable -Name 'Variable1' -Value "Test 01"
    if ($null -eq $Variable1 -or $Variable1 -eq [string]::Empty) {
        return "Varaible 1 missing"
    }

    Set-Variable -Name 'Variable2' -Value "Test 02"
    if ($null -eq $Variable2 -or $Variable2 -eq [string]::Empty) {
        return "Varaible 2 missing"
    }
}

Describe 'Test-Variable' {

    Context "Mock Variable1" {
        Mock -CommandName 'Set-Variable' -MockWith {return $null} -Verifiable -ParameterFilter {$Name -and $Name -eq 'Variable1'}

        it 'Mock Variable1' {
            $Result = Test-Variable
            $Result | Should -Contain "Varaible 1 missing"
           
            Assert-MockCalled -CommandName Set-Variable -Times 1 -Exactly -ParameterFilter {$Name -eq 'Variable1'}
        }
    }

    Context "Mock Variable2" {
        Mock -CommandName 'Set-Variable' -MockWith {return $null} -Verifiable -ParameterFilter {$Name -and $Name -eq 'Variable2'}

        it 'Mock Variable2' {
            $Result = Test-Variable
            $Result | Should -Contain "Varaible 2 missing"
           
            Assert-MockCalled -CommandName Set-Variable -Times 1 -Exactly -ParameterFilter {$Name -eq 'Variable2'}
        }
    }
}
```
```
Describing Test-Variable

  Context Mock Variable1
    [+] Mock Variable1 183ms

  Context Mock Variable2
    [-] Mock Variable2 125ms
      Expected 'Varaible 2 missing' to be found in collection Varaible 1 missing, but it was not found.
      19:             $Result | Should -Contain "Varaible 2 missing"
```

I don't know if it's a bug or I'm doing something wrong ?

Environment
Pester : 4.4.0
PSVersion : 5.1.14409.1012

BrandonLundt:
Very small change to Invoke-Mock. Added a change to Set-Variable mocking to include a scope parameter of Global. This is intended to resolve #1107

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -1895,3 +1895,56 @@ Describe "Restoring original commands when mock scopes exit" {
         }
     }
 }
+
+Describe "Mocking Set-Variable" {
+    Context "Script context" {
+        Mock Set-Variable -ParameterFilter {$false}
+        Set-Variable -Name v1 -Value 1 -Scope 1
+        Set-Variable -Name v2 -Value 2 -Scope script
+        It "Set-Variable forces scope" { 
+            $v1 | Should -Not -BeNullOrEmpty
+            $v1 | Should -Be 1
+            $v2 | Should -Not -BeNullOrEmpty
+            $v2 | Should -Be 2
+        }
+    }
+    Context "Function context" {
+        function Test-Variable {
+            Param()
+        
+            Set-Variable -Name 'Variable1' -Value "Test 01"
+            if ($null -eq $Variable1 -or $Variable1 -eq [string]::Empty) {
+                return "Varaible 1 missing"
+            }
+        }
+        Mock -CommandName 'Set-Variable' -MockWith {return $null} -Verifiable -ParameterFilter {$Name -and $Name -eq 'Variable1'}
+        it 'Mock Variable1' {
+            $Result = Test-Variable
+            $Result | Should -Contain "Varaible 1 missing"
+            
+            Assert-MockCalled -CommandName Set-Variable -Times 1 -Exactly -ParameterFilter {$Name -eq 'Variable1'}
+        }
+    }
+    Context "Function second context" {
+        function Test-Variable {
+            Param()
+        
+            Set-Variable -Name 'Variable1' -Value "Test 01"
+            if ($null -eq $Variable1 -or $Variable1 -eq [string]::Empty) {
+                return "Varaible 1 missing"
+            }
+            Set-Variable -Name 'Variable2' -Value "Test 02"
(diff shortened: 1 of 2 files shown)


nohwnd:
Thanks for the PR, but this is not the solution I described to you on Twitter. You need to add parameter `-Scope 2` if there is none. Ignore the parameters if it is already Script or Global, which are non-relative scopes, or  add 2 to the `-Scope` parameter if it is present and numeric. 

Setting all the mocked variables to Global scope might fix the one symptom, but it does not fix it in a clean way. 

Also we need some tests for this. The two that are in the original issue should be a good start. 

(you can simply fixup the solution and push to the same branch, the PR will update)

BrandonLundt:
So, I'm not sure if I am even close with resolving this issue. Your last comment leads me to believe I have missed the boat entirely.

Attempting to write some tests for this condition, I am not able to get Assert-MockCalled to return true.

What am I looking for is, am I on the right track or should I start over?

nohwnd:
@BrandonLundt seems to be quite close.

BrandonLundt:
Looking at the TeamCity build failures I see:

`[Pester] Invoke-Pester swallows pipeline output from system-under-test.Should swallow test output without -PassThru
[07:14:49][Invoke-Pester swallows pipeline output from system-under-test.Should swallow test output without -PassThru] Exception calling "Create" with "1" argument(s): "The process cannot access the file 'C:\TeamCity\BuildAgent\work\93ac3fcead2a62ac\Test.v2.xml' because it is being used by another process."
[07:14:49][Invoke-Pester swallows pipeline output from system-under-test.Should swallow test output without -PassThru] at <ScriptBlock>, C:\TeamCity\BuildAgent\work\93ac3fcead2a62ac\Functions\TestResults.ps1: line 70
70:         $xmlFile = [IO.File]::Create($Path)`

Which I do not believe to be an issue with my pull request but rather with multiple items running in TeamCity at the same time.

Any ideas how I rerun these build tasks?

nohwnd:
@BrandonLundt yeah, I can re-run it, or you can push some change to the repository. But this already happened with some other build, so it might be a problem connected to migrating the builds to a different infrastructure. @dlwyatt you were doing some kind of upgrade recently, right? Have you observed this problem somewhere else?

dlwyatt:
Not that I've noticed.  The agents are only supposed to be used for a single build and then discarded.  Will check.

PR #1112, opened 2018-10-08
Add TestRegistry

### LINKED ISSUE #402: Feature Request: TestDrive for Registry
One thing that I have been doing frequently in my scripts is writing to the registry. I've been using mocking to unit test anything that I write to the registry and it has worked well, but after using TestDrive in some of my scripts for writing and deleting files I keep wishing that I could do the same thing with the registry.

I was wondering if anyone else would be interested in seeing a feature like TestDrive for the Registry, and if its possible, could we see that as an added feature some time in the future.

andikrueger:
closes #402 

First draft of how a TestRegistry implementation could work. Currently unit-test for TestRegistry are missing.


andikrueger:
@nohwnd Thank you so much for finishing the PR.  I appreciate it very much!

PR #1116, opened 2018-10-11
Add Get-ShouldOperator

### LINKED ISSUE #878: How do I discover Should operators?
As I understand it, using Pester 4 the better practice is to use named parameters with `Should` to compose assert statements.  The following is an example of such use:

```PowerShell
$r | Should -beOfType ([string])  # <= I think this is the preferred syntax
```

I believe it has been stated that this is preferable to composition using arguments to should like the following:

```PowerShell
$r | Should beOfType ([string]) # <= I think this syntax is deprecated
```

Because of the move toward operators as parameters, I expected there to be a way to discover the operators as is possible for most other function's parameters.  I have not yet found a way to do that.  Invoking `help should -Parameter *`, for example, outputs no useful information.  `help about_Should` outputs a number of useful examples.  But that command merely outputs whatever was written.  I see no reason to expect that it currently is, or will ever be, reliably comprehensive of all operators.

How should I, as a normal user, discover the `Should` operators?

brianbunke:
This PR would add a new function, `Show-PesterAssertion`, in an attempt to improve discoverability of Should parameter assertions.

Supercedes #1111.
Resolves #878. (See the issue for far more details.)

### DIFF
--- a/Functions/Get-ShouldOperator.Tests.ps1
+++ b/Functions/Get-ShouldOperator.Tests.ps1
@@ -16,7 +16,10 @@ InModuleScope Pester {
             }
 
             It 'Returns Name and Alias properties' {
-                $get1[0].PSObject.Properties | Select-Object -ExpandProperty Name | Should -Be 'Name','Alias'
+                $get1[0].PSObject.Properties |
+                    Select-Object -ExpandProperty Name |
+                    Sort-Object |
+                    Should -Be 'Alias','Name'
             }
         }
 


brianbunke:
Per discussion in #878, changed name to `Get-ShouldOperator`. This PR will also add help for and rename all internal assertion operator functions (`PesterBe` --> `Should-Be`). I have already updated Be and BeExactly; need to get the rest of the built-ins, and then will remove the WIP label from this PR.

nohwnd:
Maybe `Add-AssertionOperator` could take the command object, and keep reference to it, and take the name and definition from it. Just thinking out loud.

brianbunke:
If I'm understanding correctly, that would be a breaking change to `Add-AssertionOperator`, which is now public.

I assume that change would look something like:

```posh
Add-AssertionOperator -Name Exist -Command (Get-Command Should-Exist)
```

And the internals could figure out the test script block, the help, and maybe even the `SupportsArrayInput`.

This PR has been enough work that I would love to not throw in a breaking change in on top of it. 🙂 But if that's correct, and you'd still like to see it done that way, let me know!

brianbunke:
If I'm understanding correctly, that would be a breaking change to `Add-AssertionOperator`, which is now public.

I assume that change would look something like:

```posh
Add-AssertionOperator -Name Exist -Command (Get-Command Should-Exist)
```

And the internals could figure out the test script block, the help, and maybe even the `SupportsArrayInput`.

This PR has been enough work that I would love to not throw in a breaking change in on top of it. 🙂 But if that's correct, and you'd still like to see it done that way, let me know!

brianbunke:
If I'm understanding correctly, that would be a breaking change to `Add-AssertionOperator`, which is now public.

I assume that change would look something like:

```posh
Add-AssertionOperator -Name Exist -Command (Get-Command Should-Exist)
```

And the internals could figure out the test script block, the help, and maybe even the `SupportsArrayInput`.

This PR is massive enough that I would love to not throw in a breaking change in on top of it. 🙂 But if that's correct, and you'd still like to see it done that way, let me know!

brianbunke:
If not pursuing the above, I think this PR is finally ready. I removed the WIP label from the PR title.

Summarizing:

- Add `Get-ShouldOperator`
  - Returns name and alias of all registered operators
  - `-Name` is a dynamic parameter to allow for tab completion
  - If specifying one operator, returns the PowerShell examples help for that operator
- Add optional `-InternalName` parameter to `Add-AssertionOperator`
  - Allows `Get-ShouldOperator -Name Be` to fetch help examples
- Renames all Should operators' internal names to `Should-xxx`
  - From `PesterBe` to `Should-Be`, etc.
- Adds help synopsis and examples to all Should operators

nohwnd:
> This PR is massive enough that I would love to not throw in a breaking change in on top of it. 🙂 But if that's correct, and you'd still like to see it done that way, let me know!

I was thinking two parameters sets, so the change would not be breaking, but fair enough. Adding that can be done later, or never.

Now I just need to test it.

brianbunke:
Ah, right, I never think of the second parameter set. That could work, but yeah, I'll leave it alone for now.

PR #1138, opened 2018-11-03
Fix features and scenario names do not display when PesterResult is printed to the console

fourpastmidnight:

I used the `-PassThru` parameter of `Invoke-Gherkin` and discovered that feature names simply displayed as `Gherkin.Ast.Feature`, and likewise, scenario names displayed as `Gherkin.Ast.Scenario`. This isn't helpful. So this pull request adds the file `Gherkin.types.ps1xml` which adds "overrides" of the ToString method as ScriptMethod methods to the `Gherkin.Ast.Feature` and `Gherkin.Ast.Scenario` objects.


fourpastmidnight:
How do I resubmit the build? A test failed (not one of the ones related to this pull request) due to a resource contention (a file was locked which was trying to be written by the test&mdash;perhaps tests that need to use files need to be modified to use a random-generated filename to prevent transient errors like this?) I'm guessing I could amend the last commit to force a change to it's SHA to force a build....

PR #1139, opened 2018-11-04
Adds support for 'Scenario Template'

### LINKED ISSUE #1123: Using the Gherkin keywords Scenario Template in place of Scenario Outline causes scenarios to be executed twice

Using the synonymous Gherkin keywords `Scenario Template` in place of `Scenario Outline' may cause the scenario associated with those keywords (or the scenario before it) to be executed twice.

## 3. Expected Behavior

I expected the scenario associated with the keywords (or the scenario preceding the scenario using the keywords) to be executed only once.

## 4.Current Behavior

If you have the following Gherkin file:

```gherkin
Scenario Outline: This scenario will be executed twice

    # A scenario, here
    # With some examples here

@ignore
Scenario Template: This is an ignored scenario

    # Another scenario here
    # More examples here
```

**or** this Gherkin file:

```gherkin
Scenario Template: This scenario will be executed twice

    # A scenario here
    # With some examples here
```

and run Pester with the following command:

    Invoke-Gherkin -Path .\features\Path-to-feature-file.feature -ExcludeTag ignore

then the scenarios indicated in the gherkin files above will be executed twice.

## 5. Possible Solution

If Pester is not already using the latest Gherkin parser (it probably isn't, as other newer Gherkin keywords are not yet supported) then update Pester to use the latest Gherkin parser. Or, for the version of the Gherkin parser in use, ensure that all Gherkin keywords supported by the parser are also supported by Pester.

## 6. Context

I was using the keywords `Scenario Template` and `Scenarios` instead of the older `Scenario Outline` and `Examples` keywords. Note, however, that Pester does not seem to have any issues with interchanging the `Examples` and `Scenarios` keywords.

I had to revert my feature files to use only the `Scenario Outline` keywords.

fourpastmidnight:

Adds support for 'Scenario Template', which can be used in place of
'Scenario Outline'. Fixes #1123.

### DIFF
--- a/Functions/Gherkin.ps1
+++ b/Functions/Gherkin.ps1
@@ -390,11 +390,8 @@ function Import-GherkinFeature {
              $null = & $SafeCommands["Add-Member"] -MemberType "NoteProperty" -InputObject $Step.Location -Name "Path" -Value $Path
         }
 
-        switch ($Child.Keyword.Trim()) {
-            "Scenario" {
-                $Scenario = Convert-Tags -InputObject $Child -BaseTags $Feature.Tags
-            }
-            "Scenario Outline" {
+        switch -Regex ($Child.Keyword.Trim()) {
+            "Scenario(?: (?:Outline|Template))?" {
                 $Scenario = Convert-Tags -InputObject $Child -BaseTags $Feature.Tags
             }
             "Background" {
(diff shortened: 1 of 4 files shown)


Functions/Gherkin.ps1
```diff
@@ -397,6 +397,9 @@ function Import-GherkinFeature {
             "Scenario Outline" {
                 $Scenario = Convert-Tags -InputObject $Child -BaseTags $Feature.Tags
             }
+            "Scenario Template" {
```
Jaykul: This is making me think maybe we should only look at the first word in the keyword ;-)

PR #1141, opened 2018-11-04
Don't run Pester version tests on forks

fourpastmidnight:

I had a PR build fail due to an unrelated unit test failing. But, just
to be sure, I checked out my branch locally and ran the entire test
suite starting at the root.

Lo and behold, I had some failing tests--but all due to ensuring that
the correct module manifest version is being found. I also found that at
least one test was skipped if the git remote did not contain 'github.com/Pester/'.

It seems to me that the other tests in the same area of the suite which
also check module versioning information should also not run if the
current git repository is a fork.

This commit moves those tests inside the `if` statement which checks
whether or not the current git repository is the official Pester
repository, or whether or not it's a fork.

After this change, the test suite ran to completion with no failures.


nohwnd:
makes sense, thx.

PR #1142, opened 2018-11-05
WIP: Fix #1124 pester does not print doc strings or data tables

### LINKED ISSUE #1124: Pester does not print Tables or DocStrings that are part of steps

When using DocString or tables as part of steps, this data is not printed to the console during the test run, unlike in cucumber (for Ruby) or SpecFlow for .NET.

## 3. Expected Behavior

For this step in a gherkin file:

```gherkin
Then  it should be displayed as:
  """

  Major Minor Patch PreReleaseTag BuildMetadataTag
  ----- ----- ----- ------------- ----------------
      1     4     5 rc.1



  """
```

when running the tests, it should be displayed on the console as

```
    [+] Then it should be displayed as:

      Major Minor Patch PreReleaseTag BuildMetadataTag
      ----- ----- ----- ------------- ---------------
          1     4     5 rc.1


```

Per _The Cucumber Book_ on page 78:

> Just like a data table, the entire string between the `"""` triple quotes is attached to the step above it.

## 4.Current Behavior

Currently, only the main text of the step is displayed:

```
    [+] Then it should be displayed as:
```

## 5. Possible Solution

For steps containing tables or DocStrings, display the table or DocString with the step text just as Cucumber does.

## 6. Context

While it's nice that the Gherkin style tests for Pester have reused much of the existing Pester test framework, some nice details about using Gherkin for BDD style tests was "lost in translation" (or perhaps, just not quite finished). This issue merely exists to track/record the desire to bring the Gherkin style tests in Pester up to parity with Cucumber, where possible.

The value in displaying this data, though it does take up considerable console real-estate, is that these style of tests are meant to be more readable than compact.

fourpastmidnight:

Fixes #1124.

For Gherkin Step Definitions which have DocString or DataTable data, this data is now printed to the console by default. A new switch parameter has been added to `Invoke-Gherkin` called `HideStepData` which controls whether or not DocString/DataTable data is printed to the console: when enabled, this data is not printed. The default is for the data to be printed.

### DIFF
--- a/Functions/Gherkin.ps1
+++ b/Functions/Gherkin.ps1
@@ -134,6 +134,10 @@ function Invoke-Gherkin {
             This parameter does not affect the PassThru custom object or the XML output that
             is written when you use the Output parameters.
 
+        .PARAMETER HideStepData
+            Controls whether or not Step Definition tables or DocStrings are printed to the console
+            during the test run.
+
         .PARAMETER PassThru
             Returns a custom object (PSCustomObject) that contains the test results.
             By default, Invoke-Gherkin writes to the host program, not to the output stream (stdout).
@@ -218,6 +222,8 @@ function Invoke-Gherkin {
 
         [Pester.OutputTypes]$Show = 'All',
 
+        [Switch]$HideStepData,
+
         [switch]$PassThru
     )
     begin {
@@ -282,7 +288,7 @@ function Invoke-Gherkin {
         Enter-CoverageAnalysis -CodeCoverage $CodeCoverage -PesterState $pester
 
         foreach ($FeatureFile in & $SafeCommands["Get-ChildItem"] $Path -Filter "*.feature" -Recurse ) {
-            Invoke-GherkinFeature $FeatureFile -Pester $pester
+            Invoke-GherkinFeature $FeatureFile -Pester $pester -HideStepData:$HideStepData
         }
 
         # Remove all the steps
@@ -378,9 +384,14 @@ function Import-GherkinFeature {
 
         .PARAMETER Pester
             Internal Pester object. For internal use only
+
+        .PARAMETER HideStepData
+            Controls whether or not Step Definition text includes DocStrings or Tables.
+            If this parameter is set, then DocStrings or tables will not be appended to
+            Step Definition text and will not be output to the console.
     #>
     [CmdletBinding()]
-    param($Path, [PSObject]$Pester)
+    param($Path, [PSObject]$Pester, [Switch]$HideStepData)
     $Background = $null
 
     $parser = & $SafeCommands["New-Object"] Gherkin.Parser
@@ -443,8 +454,58 @@ function Import-GherkinFeature {
                     & $SafeCommands["New-Object"] Gherkin.Ast.Scenario $ExampleSet.Tags, $Scenario.Location, $Scenario.Keyword.Trim(), $ScenarioName, $Scenario.Description, $Steps | Convert-Tags $Scenario.Tags
                 }
             }
-        } else {
+        } elseif ($HideStepData) {
             $Scenario
+        } else {
+            $Steps = foreach ($Step in $Scenario.Steps) {
+                [string]$StepText = $Step.Text
+                if ($Step.Argument -is [Gherkin.Ast.DocString]) {
+                    $StepText += "$([Environment]::NewLine)  `"`"`"$([Environment]::NewLine) $(foreach ($str in ($Step.Argument.Content -split [Environment]::NewLine)) { " $str$([Environment]::NewLine)" })  `"`"`""
+                }
+                if ($Step.Argument -is [Gherkin.Ast.DataTable]) {
+                    $table = $Step.Argument.Rows |
+                        & $SafeCommands['ForEach-Object'] { ,@($_ |
+                        & $SafeCommands['Select-Object'] -ExpandProperty Cells |
+                        & $SafeCommands['Select-Object'] -ExpandProperty Value) }
+
+                    if ($Step.Argument.Rows[0].Cells.Length -gt 1) {
+                        $transposedTable = for ($i = $table[0].Length - 1; $i -ge 0; $i--) {
+                            ,@(for ($j = 0; $j -lt $table.Length; $j++) {
+                                $table[$j][$i]
+                            })
+                        }
+
+                        [Array]::Reverse($transposedTable)
+                        $cellWidths = $transposedTable |
+                            & $SafeCommands['ForEach-Object'] { $_ |
+                            & $SafeCommands['Measure-Object'] -Property Length -Maximum |
+                            & $SafeCommands['Select-Object'] -ExpandProperty Maximum }
+                    } else {
+                        $cellWidths = @($table |
+                            & $SafeCommands['ForEach-Object'] { $_ } |
+                            & $SafeCommands['Measure-Object'] -Property Length -Maximum |
+                            & $SafeCommands['Select-Object'] -ExpandProperty Maximum)
+                    }
+
+                    $tableText = ""
+                    foreach ($row in $Step.Argument.Rows) {
+                        $rowText = "  |"
+                        for ($j = 0; $j -lt $row.Cells.Length; $j++) {
+                            $rowText += " {0,$(-$cellWidths[$j])} |" -f $row.Cells[$j].Value
+                        }
+
+                        $tableText += "$([Environment]::NewLine)$rowText"
+                    }
+
+                    $StepText += "$($tableText.TrimEnd())"
+                }
+                if ($StepText -ne $Step.Text) {
+                    & $SafeCommands["New-Object"] Gherkin.Ast.Step $Step.Location, $Step.Keyword.Trim(), $StepText, $Step.Argument
+                } else {
+                    $Step
+                }
+            }
+            & $SafeCommands["New-Object"] Gherkin.Ast.Scenario $null, $Scenario.Location, $Scenario.Keyword.Trim(), $Scenario.Name, $Scenario.Description, $Steps | Convert-Tags $Scenario.Tags
         }
     }
     )
@@ -464,7 +525,8 @@ function Invoke-GherkinFeature {
         [Parameter(Mandatory = $True, Position = 0, ValueFromPipelineByPropertyName = $True)]
         [IO.FileInfo]$FeatureFile,
 
-        [PSObject]$Pester
+        [PSObject]$Pester,
+        [Switch]$HideStepData
     )
     # Make sure broken tests don't leave you in space:
     $CWD = [Environment]::CurrentDirectory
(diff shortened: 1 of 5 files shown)


Functions/Gherkin.ps1
```diff
@@ -434,8 +445,48 @@ function Import-GherkinFeature {
...
-        } else {
+        } elseif ($HideStepData) {
             $Scenario
+        } else {
+            $Steps = foreach ($Step in $Scenario.Steps) {
+                [string]$StepText = $Step.Text
+                if ($Step.Argument -is [Gherkin.Ast.DocString]) {
+                    $StepText += "$([Environment]::NewLine)  `"`"`"$([Environment]::NewLine) $(foreach ($str in ($Step.Argument.Content -split [Environment]::NewLine)) { " $str$([Environment]::NewLine)" })  `"`"`""
+                }
+                if ($Step.Argument -is [Gherkin.Ast.DataTable]) {
```
fourpastmidnight: One possible improvement that could be made is to extract this functionality out into a private Function. The same could be said for where this function processes Scenario Outlines to create new Scenarios for each table row of data.

Functions/Gherkin.ps1
```diff
@@ -576,72 +628,74 @@ function Invoke-GherkinScenario {
     Exit-MockScope
 }
 
-function Find-GherkinStep {
```
fourpastmidnight: I meant to completely remove this--not just comment it out. I guess I should first solicit comments on the removal of this unused function--because perhaps it's there for a reason? Depending on the responses I get, I'll either drop this commit or rebase the _actual_ removal of the function.

Functions/Gherkin.ps1
fourpastmidnight: One possible improvement that could be made is to extract this functionality out into a private Function. The same could be said for where this function processes Scenario Outlines to create new Scenarios for each table row of data.

Functions/Gherkin.ps1
fourpastmidnight: I meant to completely remove this--not just comment it out. I guess I should first solicit comments on the removal of this unused function--because perhaps it's there for a reason? Depending on the responses I get, I'll either drop this commit or rebase the _actual_ removal of the function.

Jaykul: This is meant for tooling --hypothetically-- your editor can call this when it sees you writing in a .feature file to show you intellisense, to auto-complete "when ..." sentences, or highlight _missing_ steps, etc.  It's basically the first step toward integrating with the PSES...

fourpastmidnight:
OH!! Definitely useful in the future. Ok, I’ll drop that commit. Thanks!!

Sent from my Windows 10 phone

Functions/Gherkin.ps1
fourpastmidnight: OH!! Definitely useful in the future. Ok, I’ll drop that commit. Thanks!!

Functions/Gherkin.ps1
```diff
@@ -218,6 +222,8 @@ function Invoke-Gherkin {
 
         [Pester.OutputTypes]$Show = 'All',
 
+        [Switch]$HideStepData,
```
fourpastmidnight: @Jaykul: Just wanted to check, based on my previous comment, do you think it's a good idea to definitely add the option to disable the displaying of DocStrings/DataTables in the console to the `OutputType` enum? I'd like to wrap this up.

@nohwnd Do you have any preference as to how this should be handled, especially given that RSpec style tests don't make use of DocStrings nor DataTables (to my knowledge&mdash;I did see something called _implicit docstrings_, but that also implied the use of an extension to RSpec, so I wouldn't think this applies to Pester)?

Functions/Gherkin.ps1
fourpastmidnight: Wow, I recently used `Find-GherkinStep` to find an existing step and execute it from within another step to help keep my test code dry. So this is definitely a useful cmdlet.

fourpastmidnight:
Well this is a bummer. I think printing the DocStrings and Tables to the console messes up the problem matcher. At least there's an option to turn off displaying these in this feature.

This seems to be a common issue however: changes made to the output mess up the problem matcher.

Perhaps whenever a step fails, immediately after the step text, there's some "sentinal" text, like `--- Error Details ---` is printed to the console. In this way, the problem matcher can match this line, then the first two lines following would be the actual data that is used to populate the Problems pane, followed by any additional information.

fourpastmidnight:
OK, I think the problem matcher being broken is not a huge deal. Let me explain why.

First: There's a way to turn off "pretty printing" of Tables and DocStrings. So, if you _really_ want the problem matcher to work, just don't print Tables and DocStrings. Ideal? No. 

(NOTE: it might be nice to output the Tables and DocStrings to the NUnit XML report, _even if_ pretty-printing of Tables and DocStrings to the console is disabled. Hmm, and this may require another option switch to `Invoke-Gherkin`. This would aid tools that take the NUnit XML report and create nice documentation from the specs, such as [Pickles](http://www.picklesdoc.com/) and/or [DonutReport](http://donutreport.github.io/donut/) [(sample DonutReport demo.)](http://donutreport.github.io/donut/demo.html).)

Second: I think I could get more bang for my buck if I were to look into implementing a test adapter VS Code extension that can be used by the VS Code [Test Explorer extension](https://github.com/hbenl/vscode-test-explorer). Again, this could be an "opt-in" type thing in lieu of using the problem matcher. And again we could leverage the NUnit XML Report output. I believe that even though this option requires additional work, longer-term it could inoculate us from breaking the all-too-fragile problem matcher everytime console output changes. I also believe this would provide a much better user experience within VS Code for surfacing test failures and navigating to them from the Test Explorer hub. (As an aside, just now I thought how it's funny that Pester, a test framework which does not require VS Code to run, has code specifically to support VS Code. It would be nice to remove this accidental coupling/knowledge about the outside world from Pester.)

Any thoughts, comments, concerns?

fourpastmidnight:
However, I did just manage to get the VS Code problem matcher to work, and I kind of like the results. I'll push up the necessary changes later, so please hold off on merging this until then.

nohwnd:
@fourpastmidnight are there more changes needed? I renamed it to `WIP:` so I stop re-reading it when I go through open PRs. When it's complete please remove the WIP and ideally tag me in comment. Thx

fourpastmidnight:
@nohwnd: Yes, more changes are needed and forthcoming. I should’ve renamed this awhile back. Sorry about that. And, thanks for renaming.

Sent from my Windows 10 phone

fourpastmidnight:
OK, so now that a separate scenario is created for each row of an example table for a scenario outline, the implementation of this needs to change a great deal.

BUT, that's OK. Because this has caused me to realize that we've mixed two unrelated concerns in Pester's Gherkin test implementation. The mixing of these two concerns was going to cause greater and greater friction at some point anyway. Adding in #1168 just caused this friction to happen sooner. But again, this is a _good_ thing. It provides the opportunity to clean up the code and separate these concerns.

The concerns in question are:

1. Generating the objects necessary to execute the tests, e.g. features, scenarios, steps, etc.
2. Formatting various strings to display the test run execution to the console (and, also in other outputs, such as the NUnit XML reports)

fourpastmidnight:
Hmm, reading _The Cucumber Book_ on around pg. 70 where they delve more into Scenario Outlines and DataTables, `cucumber` has a flag: `--expand` which, instead of "pretty printing" the DataTables, causes the individual scenarios created from each row in the data table to be printed.

So, it would seem after the implementation of #1168, though it's not printed correctly with that implementation, the effect is that `cucumber --expand` is being executed for Gherkin-style tests. So, let's use that to our advantage to control how scenario outlines will be displayed to the console with this PR.

fourpastmidnight:
And [here's a blog article that shows the output from running `cucumber --expand`](http://qastuffs.blogspot.com/2011/09/use-expand-to-expand-scenario-outline.html).

fourpastmidnight:
OK, looking at the implementation of Gherkin.dll, Gherkin does not make a distinction between Scenario Outlines and Scenarios. It parses a "ScenarioLine" and description and then some steps and then "Examples", which consist of an "ExamplesLine" (e.g. the keyword and name for the example table(s)), a description (optional) and one or more "ExamplesTable"s.

This is interesting considering the output shown in the previously linked-to blog article. It's almost as if a Scenario Outline functions much like an RSpec Context would within a Describe (except that, there's no shared state being created--the "context" is merely acting as a container for all the scenarios created from the example table(s)). This is exactly how I was planning on modeling this in Pester for Gherkin style tests. It's good to know I was on the right track.

fourpastmidnight:
Note to self: Rename `-HideStepData` to `-NoMultiline` which closely matches the cucumber switch `--no-multiline` which also suppresses outputting to the console window DataTables and DocStrings of Step Definitions.

fourpastmidnight:
I've come to the realization that the current implementation of Gherkin tests in Pester isn't conducive to running them as cucumber would run them. In fact, the current implementation leaves a lot to be desired in terms of performance and reporting. Now, I'm not "bashing" all the work that went into this--I think what has been done is absolutely phenomenal!, and my hat's off to all who have contributed to make possible what we have with Gherkin-style tests in Pester!

But it seems to me that this organically grown implementation of Gherkin tests living within the Pester framework needs a different approach. How will this fit in with the RSpec style of tests Pester supports? I'm unsure. But I am working on overhauling the Gherkin implementation, with the focus on these key areas:

1. Improved Performance
2. Improved reporting -- both to the console and as reporting artifacts
3. Fix current scoping issues when using Mocks (hopefully 😉)

So, if you think I've been quiet lately, it's because I just finished reading _Windows PowerShell in Action, 3rd Ed._ about one week ago and have been taking my first stab at a new approach to Gherkin style tests in Pester.

When I have something to show, I will submit a new PR and officially abandon this one. The new PR will be far from "Mergeable", more a WIP, but with the intent of garnering feedback.

nohwnd:
I think it would be better to wait for beta of v5 and do it there. We can then work together to do this, if you are able to describe to me what the changes should be, and I to you how the new internals work :D

fourpastmidnight:
I was thinking this would be a v5 change. 😉 Maybe I could create a v5 “issue” item with the details where further discussions can be had.

Sent from my Windows 10 phone

nohwnd:
@fourpastmidnight sure :)

fourpastmidnight:
I've recently made _significant_ progress on this for the 4.x branch of Pester. I'm still wrapping up some loose ends. I will warn however, that I have a lot of stuff coming up that will require some time/attention in the next three weeks, so it may be just a bit longer than I would like. But, wow. _Significant_ progress here!

nohwnd:
@fourpastmidnight that's great :) I still did not have time to look into your code :/ Hopefully when Pester v5 beta is out.

fourpastmidnight:
This is being superseded by #1276.

PR #1144, opened 2018-11-08
Should: Use Generic.List[PSObject] instead of ArrayList for $inputArray

vexx32:

Refactor code slightly in `Should` command. [Documentation for ArrayList](https://docs.microsoft.com/en-us/dotnet/api/system.collections.arraylist?view=netframework-4.7.2#remarks) explicitly states that Generic.List[T] should be used instead:

> We don't recommend that you use the ArrayList class for new development. Instead, we recommend that you use the generic List&lt;T&gt; class.

This allows us to drop the `$null =` output-swallowing construct for part of the function, as `List[T]` does not emit output on `.Add()`.

Additionally, typing the List with `[PSObject]` avoids some extremely edge case problems that can occur with reboxing PSObjects that have instance members attached (such as those added via `Add-Member`).

### DIFF
--- a/Functions/Assertions/Should.ps1
+++ b/Functions/Assertions/Should.ps1
@@ -96,7 +96,7 @@ function Should {
 
     begin
     {
-        $inputArray = New-Object System.Collections.ArrayList
+        $inputArray = New-Object System.Collections.Generic.List[PSObject]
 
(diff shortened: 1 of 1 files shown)


nohwnd:
Thanks!

vexx32:
https://media.giphy.com/media/tXTqLBYNf0N7W/giphy.gif

PR #1146, opened 2018-11-14
Fix using [ScriptBlock]::Create with InModuleScope.

sethvs:

We use **ScriptBlock** with **InModuleScope**.

Now we can use it like this:

```powershell
Describe "{}" {
    $ScriptBlock = {Write-Output -InputObject 'This is a ScriptBlock'}
    Import-Module -Name PowerShellGet
    InModuleScope -ModuleName PowerShellGet -ScriptBlock $ScriptBlock
}
```

the result is:
```powershell
Describing {}
```

But if we need to construct ScriptBlock dynamicly by using **[ScriptBlock]::Create** method
```powershell
Describe "[ScriptBlock]::Create" {
    $ScriptBlockText = "Write-Output -InputObject 'This is a ScriptBlock'"
    $ScriptBlock = [ScriptBlock]::Create($ScriptBlockText)
    Import-Module -Name PowerShellGet
    InModuleScope -ModuleName PowerShellGet -ScriptBlock $ScriptBlock
}
```

the result will be:

```powershell
Describing [ScriptBlock]::Create
  [-] Error occurred in Describe block 49ms
    ParameterBindingValidationException: Cannot bind argument to parameter 'SessionStateInternal' because it is null.
```

This is because, when InModuleScope function gets **$originalScriptBlockScope** of the ScriptBlock
https://github.com/pester/Pester/blob/5a53c46bc7704726e97ad747647dc711652edb96/Functions/InModuleScope.ps1#L74
the result will be $null.

It seems, that ScriptBlock created with Create method doesn't have **SessionStateInternal** property.
For example this code returns null:
```powershell
$ScriptBlockText = "Write-Output -InputObject 'This is a ScriptBlock'"
$ScriptBlock = [ScriptBlock]::Create($ScriptBlockText)
$flags = [System.Reflection.BindingFlags]'Instance,NonPublic'
[scriptblock].GetProperty('SessionStateInternal', $flags).GetValue($ScriptBlock, $null)
```

So, when InModuleScope function tries to return the **$originalScriptBlockScope** to the ScriptBlock
https://github.com/pester/Pester/blob/5a53c46bc7704726e97ad747647dc711652edb96/Functions/InModuleScope.ps1#L90
the **Set-ScriptBlockScope** throws an error, because **-SessionStateInternal** parameter cannot be null.

This PS allows **-SessionStateInternal** parameter to accept null, so that ScriptBlock will have its original SessionStateInternal back, even if it is $null.

### DIFF
--- a/Functions/InModuleScope.Tests.ps1
+++ b/Functions/InModuleScope.Tests.ps1
@@ -40,13 +40,15 @@ Describe "Executing test code inside a module" {
         }
     }
 
-    It "Can use ScriptBlock inside the module scope" {
-        $ScriptBlockOne = { Write-Output "I am a ScriptBlockOne" }
-        $ScriptBlockTwoString = 'Write-Output "I am a ScriptBlockTwo"'
-        $ScriptBlockTwo = [ScriptBlock]::Create($ScriptBlockTwoString)
+    It "Can execute bound ScriptBlock inside the module scope" {
+        $ScriptBlock = { Write-Output "I am a bound ScriptBlock" }
+        InModuleScope TestModule $ScriptBlock | Should -BeExactly "I am a bound ScriptBlock"
+    }
 
-        InModuleScope TestModule $ScriptBlockOne | Should -BeExactly "I am a ScriptBlockOne"
-        InModuleScope TestModule $ScriptBlockTwo | Should -BeExactly "I am a ScriptBlockTwo"
+    It "Can execute unbound ScriptBlock inside the module scope" {
+        $ScriptBlockString = 'Write-Output "I am an unbound ScriptBlock"'
+        $ScriptBlock = [ScriptBlock]::Create($ScriptBlockString)
+        InModuleScope TestModule $ScriptBlock | Should -BeExactly "I am an unbound ScriptBlock"
     }
 
     Remove-Module TestModule -Force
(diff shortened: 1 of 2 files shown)


Functions/InModuleScope.Tests.ps1
```diff
@@ -40,6 +40,15 @@ Describe "Executing test code inside a module" {
         }
     }
 
+    It "Can use ScriptBlock inside the module scope" {
+        $ScriptBlockOne = { Write-Output "I am a ScriptBlockOne" }
```
nohwnd: I don't see any need to mix the two tests cases together. Could you split them to `Can execute bound ScriptBlock inside the module scope` ({} makes scriptblock bound to the current session) and `Can execute unbound ScriptBlock in module scope` please? :)

sethvs:
Is it OK with **Should swallow test output without -PassThru** test in PowerShell v2?

nohwnd:
@sethvs sorry I don't get what you are asking. Could you rephrase please?

sethvs:
**Publish Status to GitHub (Pester)** check is failed on test with name **Should swallow test output without -PassThru** when it was run on **PowerShell v2**.
Just want to say, I haven't touched this test :)

nohwnd:
@sethvs  Now I get it. There is some ongoing problem with the build server, that makes some of the tests fail, I am 
not sure if it's the problem of the test or the server. Plus it's really hard to investigate when the server waits 3 hours to run the build. I tried running it again to see if the error stays the same.

nohwnd:
@sethvs Merged, thanks ! 👍

Issue #1179, opened 2018-12-20
Gherkin: bloated NUnit test report


cgnuechtel:
## 1. General summary of the issue

The created NUnit-Test report after `Invoke-Gherkin` has unnecessary nested elements.
Additionally the order is wrong when steps are used repeatedly.

The reason is the following code in `Gherkin.ps1`:

```powershell
$Pester.AddTestResult($DisplayText, ${Pester Result}.Result, $Elapsed, $PesterErrorRecord.Exception.Message, ${Pester Result}.StackTrace, $Source, $NamedArguments, $PesterErrorRecord)
```

`$Source` is used as `ParameterizedSuiteName` to group the test results.

This is problematic for two reasons:

- It introduces a new nested `test-suite/results` elements and bloats the result XML
- It creates `test-case` elements which order is different to the execution order

## 2. Describe Your Environment

```
Pester version     : 4.5.0 D:\Documents\WindowsPowerShell\Modules\Pester\Pester.psm1
PowerShell version : 5.1.15063.1446
OS version         : Microsoft Windows NT 10.0.15063.0
```

## 3. Expected Behavior

Given the following feature:

```gherkin
Feature: A test feature for reporting 1

  Scenario: Scenario 1

    Given step_001
    When step_002
    Then step_003

  Scenario Outline: Scenario 2

    Given step_<given>
      And and_<given>
    When step_<when>
      And and_<when>
    Then step_<then>
      And and_<then>

    Examples: Examples 1
      | given | when | then |
      | 101   | 102  | 103  |

    Examples: Examples 2
      | given | when | then |
      | 201   | 202  | 203  |

  Scenario: Scenario 3

    Given step_301
    When step_302
    Then step_303
    When step_302
    Then step_304
```

The expected NUnit test report would look like this:

```xml
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<test-results xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="nunit_schema_2.5.xsd" name="Pester" total="20" errors="0" failures="2" not-run="0" inconclusive="0" ignored="0" skipped="0" invalid="0" date="2018-12-20" time="20:12:14">
[...]

PR #1180, opened 2018-12-20
Improve NUnit xml report

cgnuechtel:
Fix for #1179

### DIFF
--- a/Examples/Gherkin/JustForReporting.Steps.ps1
+++ b/Examples/Gherkin/JustForReporting.Steps.ps1
@@ -1,0 +1,58 @@
+Given "step_001" {
+}
+
+When "step_002" {
+}
+
+Then "step_003" {
+}
+
+Given "step_101" {
+}
+
+Given "and_101" {
+}
+
+When "step_102" {
+}
+
+When "and_102" {
+}
+
+Then "step_103" {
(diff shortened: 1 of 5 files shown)


nohwnd:
@cgnuechtel fails this in 9 places (you should be able to see that in the details as well I hope). I guess you are using `.` navigation from array to child item property which does not work in PowerShell 2. You need `Select-Object -ExpandProperty Node` most likely, but I haven't looked at the code.

```
 [-] should contain all scenarios of feature 1 with correct names and test results 13ms
2018-12-20T19:43:23.4716250Z       RuntimeException: Property 'Node' cannot be found on this object. Make sure that it exists.
2018-12-20T19:43:23.4716250Z       at line: 299 in C:\agent\_work\3\s\Functions\Gherkin.Tests.ps1
```

cgnuechtel:
I would like to suggest more small but important improvements with the last two commits.

1. I think the position of the invocation of `$Pester.EnterTestGroup` was not correct, since it produced nested elements in case of error
2. I also think, that the file name should be replaced by the feature name in the test report (in any case of error the file name would occur in the stack trace)

nohwnd:
@cgnuechtel nobody objects, so merging.

PR #1120, opened 2018-10-13
 More comprehensive JaCoCo support

Thilas:
Following work done on #920, here are some changes that bring a more comprehensive support of JaCoCo format.

Basically, it now fully supports JaCoCo format (updated to 1.1):
* add `class` elements for each script file
* add `method` elements for each method in a script file (as well as a `script()` method for code with no method)
* use `sourcefilename` attributes to link coverage with source code
* add `counter` elements wherever it is possible to

All this allows to generate cool reports like the ones created with ReportGenerator thanks to the (very) recent [support of JaCoCo](https://github.com/danielpalme/ReportGenerator/issues/160).

Also, to minimize changes with current Pester's behavior, I removed the `-DetailedCodeCoverage` parameter added in #920 and the current coverage report is back in the console.

And of course, unit tests have been adjusted to check the new xml generated.

Let me know if you have any question. Thanks!

PR #1189, opened 2018-12-29
Clean some stuff from #920 that are no longer needed because of #1120

Thilas:
I removed the following stuff from #920 that are no longer needed because of #1120:

- the `DetailedCoverage` parameter on `Invoke-Pester`
- the `CodeCoverage` parameter on `Describe`

@nohwnd, could you please review? Thanks.


nohwnd:
Thx!

PR #1193, opened 2019-01-01
Make hashtable key access be consistently cased to prevent symptom of PowerShell bug on Unix systems where LANG is set to C.UTF-8

### LINKED ISSUE #1101: Exceptions thrown in Describe/Context BeforeAll or BeforeEach are undiagnosable.

When a BeforeAll or BeforeEach block throws an exception, Pester error output doesn't include any information about the failure. Instead, a secondary exception is thrown due to the call  to AddTestResult failing.

For example:

[-] Error occurred in Context block 3.4s
  ValidationMetadataException: The attribute cannot be added because variable Result with value  would no longer be valid.
  MethodInvocationException: Exception calling "AddTestResult" with "8" argument(s): "The attribute cannot be added because variable Result with value  would no longer be valid."

The same behavior occurs in both Describe and Context Before* blocks with the only difference being the block type (Context or Describe).

When this occurs, the only solution I've found is to edit the catch blocks in It.ps1 and Describe.ps1 and Write-Host the original failure details.  Neither the console output nor the test result file contain this inforamation.

```powershell
Pester version     : 4.4.0 /home/psrptest/.local/share/powershell/Modules/Pester/4.4.0/Pester.psm1
PowerShell version : 6.1.0-rc.1
OS version         : Unix 4.15.0.1022
```

## 3. Expected Behavior

If you're describing a bug, tell us what should happen.

The error details ($_) from the catch block should be reported and/or logged.

## 4.Current Behavior

The subsequent call to AddTestResult itself throws and this is the information reported; as indicated above.

```powershell
Context 'Verify NTLM PSSession' {
    BeforeEach {
        # If this throws, a ValidationMetadataException is thrown and reported instead of the exception 
        # caused by the this call.
        $session = New-PSSession -ComputerName $TestVariables.HostName -Credential $TestVariables.Credential -Authentication Negotiate
    }

    It 'Verifies New-PSSession -Authentication Negotiate connects with no errors' -Skip:$Skip {
            $session.ComputerName | Should -BeExactly $TestVariables.HostName
            $session.Transport | Should -BeExactly 'WSMan'
        }
```

## 5. Possible Solution

Have a solution in mind?

https://github.com/pester/Pester/wiki/Contributing-to-Pester has detailed instructions on how to contribute.

## 6. Context

How has this issue affected you? What are you trying to accomplish?

bergmeister:
Fixes #1101
The issue is that when the environment variable `LANG` is set to `C.UTF-8` on Unix systems, which has now become the default for Ubuntu 18 (and it also the case for AppVeyor's `Ubuntu` image), then PowerShell's hashtable keys access is case sensitive, see [this](https://github.com/PowerShell/PowerShell/issues/7761) PowerShell issue (which will only get fixed in 6.2) due to the way .Net Core respects this culture as being 'invariant'.
This PR ensures that hashtable key casing inconsistency in Pester get fixed so that Pester still works in such environments.
Thanks @ffeldhaus for debugging this issue and providing the relevant details to avoid the symptoms of this bug

### DIFF
--- a/Functions/Environment.ps1
+++ b/Functions/Environment.ps1
@@ -1,6 +1,6 @@
 function GetPesterPsVersion {
     # accessing the value indirectly so it can be mocked
-    (Get-Variable 'PSVersionTable' -ValueOnly).PsVersion.Major
+    (Get-Variable 'PSVersionTable' -ValueOnly).PSVersion.Major
 }
 
 function GetPesterOs {
--- a/Functions/It.ps1
+++ b/Functions/It.ps1
@@ -308,7 +308,7 @@ function Invoke-Test
 
             $result = ConvertTo-PesterResult -Name $Name -ErrorRecord $errorRecord
             $orderedParameters = Get-OrderedParameterDictionary -ScriptBlock $ScriptBlock -Dictionary $Parameters
-            $Pester.AddTestResult( $result.name, $result.Result, $null, $result.FailureMessage, $result.StackTrace, $ParameterizedSuiteName, $orderedParameters, $result.ErrorRecord )
+            $Pester.AddTestResult( $result.Name, $result.Result, $null, $result.FailureMessage, $result.StackTrace, $ParameterizedSuiteName, $orderedParameters, $result.ErrorRecord )
             #todo: disabling progress reporting see above & $SafeCommands['Write-Progress'] -Activity "Running test '$Name'" -Completed -Status Processing
         }
     }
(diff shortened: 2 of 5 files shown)


ffeldhaus:
It would be great if this pull request could be merged (and released) with priority, as currently this is causing a lot of trouble with Appveyor tests using Pester.

bergmeister:
Yes, I agree, that would be great way of celebrating 8 years of Pester ;-).
After this PR we could add a regression test for this as well but having at least one working version of Pester out would be great at the moment, since PowerShell will likely not release a patch for the issue due to the wide impact of the necessary change.
cc @nohwnd

PR #1195, opened 2019-01-03
Enable TestResults.Tests.ps1 tests on all platform because they can be made to work on all platforms

bergmeister:
Enabling tests that already work on all platforms instead of running them just on Windows only, only one small adoption had to be made.
On non-Windows, the `ComputerName` environment variable is not present, instead call the `hostname` binary, which returns the same info but is present on all platforms. Please use a whitespace-ignoring diff since I had to decrease indentation due to the removed if condition, after that the diff is a 1-liner in line 128

image

### DIFF
--- a/Functions/TestResults.Tests.ps1
+++ b/Functions/TestResults.Tests.ps1
@@ -1,230 +1,228 @@
 Set-StrictMode -Version Latest
 
 InModuleScope Pester {
-    if ((GetPesterOs) -eq 'Windows') {
-        Describe "Write nunit test results" {
-            Setup -Dir "Results"
+    Describe "Write nunit test results" {
+        Setup -Dir "Results"
 
-            It "should write a successful test result" {
-                #create state
-                $TestResults = New-PesterState -Path TestDrive:\
-                $testResults.EnterTestGroup('Mocked Describe', 'Describe')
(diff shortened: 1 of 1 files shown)


ffeldhaus:
Wouldn't it be better to use the .NET `[Environment]` instead of `$env`? With `[Environment]` the `MachineName` is present on all platforms and matches the . Also `UserName` and other environment variables could be accessed via `[Environment]`.

Also the list of changes in GitHub shows a huge number of changes, maybe due to changes to whitespaces / tabs or other things. In my opinion those changes should be a separate pull request if really necessary.

bergmeister:
Please read my summary, I changed it to call the hostname binary instead of using `$env` for a particular test that would've failed on Unix otherwise.
A pull request should not leave the indentation in a bad state, the purpose of configuring the diff tool as I described is that one can ignore the reduced indentation to get a very clear diff. I am not touching any other code other than the one that was necessary.

nohwnd:
@bergmeister Great work, thanks!

ffeldhaus:
@bergmeister Sorry for my comment, I've read over your comment regarding the whitespace diff. 
With the reference to `[Environment]` I just wanted to point out that using `[Environment]::MachineName` returns the machine name on all platforms and may have been a better alternative than using `$(hostname)`.

PR #1196, opened 2019-01-04
Exclude Pester Tests from CodeCoverage by default

### LINKED ISSUE #1014: Let CodeCoverage exclude Pester Tests by default
## Describe Your Environment

```powershell
Pester version     : 4.3.1 C:\Program Files\WindowsPowerShell\Modules\Pester\4.3.1\Pester.psm1
PowerShell version : 5.1.16299.251
```

## Expected Behavior

When using `-CodeCoverage *`, Pester test scripts (`*.Tests.ps1`) should be excluded from code coverage reoprting. 

## Current Behavior

When using `-CodeCoverage *`, Pester test scripts (`*.Tests.ps1) are included in code coverage reporting.

```powershell
Invoke-Pester -Path . -CodeCoverage *
WARNING: CodeCoverage path '*' resolved to a non-PowerShell file 'Some.psd1'; this path will not be part of the coverage report.
WARNING: CodeCoverage path '*' resolved to a non-PowerShell file 'Some.pssproj'; this path will not be part of the coverage report.
WARNING: CodeCoverage path '*' resolved to a non-PowerShell file 'Some.zip'; this path will not be part of the coverage report.```
```

## Possible Solution

Exclude `*.Tests.ps1` file from CodeCoverage, just like module manifests (`*.psd1`) or other non-PowerShell files.

Replace line 131 of Coverage.ps1 with:

```powershell
    if (
        ($item -is [System.IO.FileInfo]) -and 
        (('.ps1','.psm1') -contains $item.Extension) -and 
        (-not $item.Name.ToLower().EndsWith('.tests.ps1'))
    )
```

## Context

When using Code Coverage, it is desirable to just use `-CodeCoverage *`, and not having to create a file set with the files to be covered up front.

Reporting code coverage over test scripts (by default) is not very useful.

fflaten:

Using `-CodeCoverage *` includes test coverage for Pester Tests (`*.tests.ps1`) by default which is rarely useful information. This PR changes the behaviour to exclude Pester Tests if not specifically specified using a path that ends with "*.test.ps1" or a specific test-filename.

Fixes #1014

### Design

I could think of three possible solutions to this feature:

1. Introduce a new switch like `-IncludeTestsInCoverage` in `Invoke-Pester`. That would require more changes and make `Invoke-Pester` even more complex while still not being able to activate it for a specific path if you're using multiple paths/coverageinfo in a single `Invoke-Pester` call. Decided to skip this.
2. (**Initial implementation**) Exclude/Include Pester Tests based on the path provided. Since you can specify an array of CoverageInfo-objects or paths then we could ignore tests when using wildcard, unless you specifically add one that ends with `*.tests.ps1`. This way you can include some tests, while ignoring others. You may also specify a single test-file directly.
3. Extend the CoverageInfo-object to include a new `IncludeTests` with a default value of false. That way users can enable it for a specific file or wildcard-path. We could also allow the use of `*.tests.ps1` in a string path input to activate it by checking the path in `New-CoverageInfo` and setting `IncludeTests` to true if it matches.

What do you think? Keep solution 2 or change to solution 3?

### Other questions

- Due to the behaviour being controlled by the path provided in the current draft, I included a warning message when using `*` or `*.ps1` at the end of the path to explain why tests are not covered and how to include them. Keep it or remove it and add the tip to the help for `-CodeCoverage` in `Invoke-Pester` and/or wiki?

Example:

```
PS> Invoke-Pester -Path $null -CodeCoverage C:\Git\Pester\Functions\Coverage*
WARNING: CodeCoverage for Pester Tests (*.tests.ps1) are excluded for path 'C:\Git\Pester\Functions\Coverage.tests*'. Include a path ending with '*.tests.ps1' or a specific file to include Pester Tests.
Executing all tests in ''
Tests completed in 0ms
Tests Passed: 0, Failed: 0, Skipped: 0, Pending: 0, Inconclusive: 0
Code coverage report:
Covered 27,02 % of 396 analyzed Commands in 1 File.
Missed commands:

File         Function                         Line Command
----         --------                         ---- -------
Coverage.ps1                                     1 if ($PSVersionTable.PSVersion.Major -le 2)...
...
```

- I've used some internal functions for the unit tests to isolate it, so I didn't think that it belonged in the normal "Code Coverage Analysis"-describe block. What do you think?

### DIFF
--- a/Functions/Coverage.ps1
+++ b/Functions/Coverage.ps1
@@ -69,7 +69,13 @@ function Get-CoverageInfoFromUserInput
 
 function New-CoverageInfo
 {
-    param ([string] $Path, [string] $Class = $null, [string] $Function = $null, [int] $StartLine = 0, [int] $EndLine = 0)
+    param ([string] $Path, [string] $Class = $null, [string] $Function = $null, [int] $StartLine = 0, [int] $EndLine = 0, [bool] $IncludeTests = $false)
+
+    # If user did not specify IncludeTests-value then auto-detect based on path
+    if ($PSBoundParameters.ContainsKey("IncludeTests") -eq $false) {
+        $testsPattern = '\.tests\.ps1$'
+        $IncludeTests = $Path -match $testsPattern
+    }
 
     return [pscustomobject]@{
         Path = $Path
@@ -77,6 +83,7 @@ function New-CoverageInfo
         Function = $Function
         StartLine = $StartLine
         EndLine = $EndLine
+        IncludeTests = $IncludeTests
     }
 }
 
@@ -94,11 +101,13 @@ function Get-CoverageInfoFromDictionary
     $endLine = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'EndLine', 'End', 'e'
     [string] $class = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Class', 'c'
     [string] $function = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Function', 'f'
+    $includeTests = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'IncludeTests', 'Tests' 't'
 
     $startLine = Convert-UnknownValueToInt -Value $startLine -DefaultValue 0
     $endLine = Convert-UnknownValueToInt -Value $endLine -DefaultValue 0
+    [bool] $includeTests = Convert-UnknownValueToInt -Value $includeTests -DefaultValue 0
 
-    return New-CoverageInfo -Path $path -StartLine $startLine -EndLine $endLine -Class $class -Function $function
+    return New-CoverageInfo -Path $path -StartLine $startLine -EndLine $endLine -Class $class -Function $function -IncludeTests $includeTests
 }
 
 function Convert-UnknownValueToInt
@@ -121,9 +130,13 @@ function Resolve-CoverageInfo
 
     $path = $UnresolvedCoverageInfo.Path
 
+    $testsPattern = '\.tests\.ps1$'
+    $includeTests = $UnresolvedCoverageInfo.IncludeTests
+
     try
     {
-        $resolvedPaths = & $SafeCommands['Resolve-Path'] -Path $path -ErrorAction Stop
+        $resolvedPaths =  & $SafeCommands['Resolve-Path'] -Path $path -ErrorAction Stop |
+                        & $SafeCommands['Where-Object'] { if($_.Path -match $testsPattern) { $includeTests } else { $true } }
     }
     catch
     {
--- a/Functions/Coverage.ps1
+++ b/Functions/Coverage.ps1
@@ -101,7 +101,7 @@ function Get-CoverageInfoFromDictionary
     $endLine = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'EndLine', 'End', 'e'
     [string] $class = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Class', 'c'
     [string] $function = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Function', 'f'
-    $includeTests = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'IncludeTests', 'Tests' 't'
+    $includeTests = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'IncludeTests', 't'
 
     $startLine = Convert-UnknownValueToInt -Value $startLine -DefaultValue 0
     $endLine = Convert-UnknownValueToInt -Value $endLine -DefaultValue 0
--- a/Functions/Coverage.ps1
+++ b/Functions/Coverage.ps1
@@ -61,7 +61,12 @@ function Get-CoverageInfoFromUserInput
     }
     else
     {
-        $unresolvedCoverageInfo = New-CoverageInfo -Path ([string]$InputObject)
+        $Path = $InputObject -as [string]
+
+        # Auto-detect IncludeTests-value from path-input
+        $IncludeTests = $Path -match '\.tests\.ps1$'
+
+        $unresolvedCoverageInfo = New-CoverageInfo -Path $Path -IncludeTests $IncludeTests
     }
 
     Resolve-CoverageInfo -UnresolvedCoverageInfo $unresolvedCoverageInfo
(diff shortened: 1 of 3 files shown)


fflaten:
After submiting the PR I realised that we could also extend the `CoverageInfo`-object (user input) with a new `IncludeTests`-key that has a default value of false. If we want, we could also enable the new option when only a path is provided if it ends with `.tests.ps1`. This might be a cleaner solution.

I've updated the PR description so we can discuss it before I spend time updating the PR.

nohwnd:
@fflaten Yes, include tests is imho a good idea, because then it can be reverted to the original behavior. Please implement it.

fflaten:
Draft v2 using design option 3 published, including auto-detect for path/string input. Looks better, @nohwnd ?

Getting lots of errors on Gherkin-tests in PSv2 on my Win10 Insider build now (should be unrelated), but works fine in azure pipeline. Looks like it's the same cause as the errors I get on Nunit schema validation tests recently. .NET complaining about the uri-section in powershell.exe.config (exception: Configuration system failed to initialize). Just me?

nohwnd:
@fflaten how are you running the v2 tests on windows 10? To test properly on v2 you need Windows 7 or Vista (or appropriate server), running with `PowerShell.exe -Version 2` is not accurate.

nohwnd:
@cgnuechtel Does [this](https://github.com/pester/Pester/pull/1196/files#diff-473116ef56eb89cfff55f6099234e384R76) affect Gherkin? Should there be match for scenarios/features (or what types of files you have)?

fflaten:
@nohwnd Didn't have a Win7 computer available so I took a shortcut and used `-Version 2` as it is 95% accurate (+ CI should catch the rest). That's also why I haven't posted it as an issue with the Nunit schema tests, since it looks to be caused by the new .NET application config-style that PowerShell 5 uses. Just wanted to mention it in case it wasn't only a `-version 2`-bug :-)

Functions/Coverage.ps1
```diff
@@ -69,14 +69,21 @@ function Get-CoverageInfoFromUserInput
 
 function New-CoverageInfo
 {
-    param ([string] $Path, [string] $Class = $null, [string] $Function = $null, [int] $StartLine = 0, [int] $EndLine = 0)
+    param ([string] $Path, [string] $Class = $null, [string] $Function = $null, [int] $StartLine = 0, [int] $EndLine = 0, [bool] $IncludeTests = $false)
+
+    # If user did not specify IncludeTests-value then auto-detect based on path
+    if ($PSBoundParameters.ContainsKey("IncludeTests") -eq $false) {
```
nohwnd: This is a bit confusing to me. I can see that `Get-CoverageInfoFromUserInput` calls `New-CoverageInfo` in two places. Once directly and once through the dictionary parser. In the dictionary parser the bound parameter will always be specified, so this only applies to the direct call. 

The direct call will never specify `IncludeTests` because in the else branch we got just a string. So imho this should move into the `Get-CoverageInfoFromUserInput`, and the IncludeTestsParemeter should become mandatory.

Functions/Coverage.ps1
```diff
@@ -94,11 +101,13 @@ function Get-CoverageInfoFromDictionary
     $endLine = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'EndLine', 'End', 'e'
     [string] $class = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Class', 'c'
     [string] $function = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Function', 'f'
+    $includeTests = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'IncludeTests', 'Tests' 't'
```
nohwnd: Please remove the `Test` and `t`, aliases. They don't indicate the intent of ignoring tests and could be more useful for different purposes in the future.

Functions/Coverage.ps1
```diff
@@ -121,9 +130,13 @@ function Resolve-CoverageInfo
...
     $path = $UnresolvedCoverageInfo.Path
 
+    $testsPattern = '\.tests\.ps1$'
+    $includeTests = $UnresolvedCoverageInfo.IncludeTests
+
     try
     {
-        $resolvedPaths = & $SafeCommands['Resolve-Path'] -Path $path -ErrorAction Stop
+        $resolvedPaths =  & $SafeCommands['Resolve-Path'] -Path $path -ErrorAction Stop |
+                        & $SafeCommands['Where-Object'] { if($_.Path -match $testsPattern) { $includeTests } else { $true } }
```
nohwnd: This would imho be easier to understand as `$includeTests -or $_.Path -notmatch $testsPattern`. It should also be faster when tests are included because then we don't match the paths.

Pester.psm1
```diff
@@ -836,13 +836,17 @@ any code coverage information, because it's not supported by the schema.
 
 Enter the path to the files of code under test (not the test file).
 Wildcard characters are supported. If you omit the path, the default is local
-directory, not the directory specified by the Script parameter.
+directory, not the directory specified by the Script parameter. Pester test files
```
nohwnd: This is a bit hard to understand when you don't know the internals. I would reword it along the lines of:  Pester test files are by default excluded from code coverage when a directory is provided. But when you provide a test file directly, code coverage will be measured. To include tests  in code coverage of a directory, use the dictionary syntax and provide IncludeTests = $true option, as shown below.

Pester.psm1
```diff
@@ -836,13 +836,17 @@ any code coverage information, because it's not supported by the schema.
...
+are excluded by default when using path input, unless a specific test file or
+a path that ends with *.tests.ps1 is provided.
 
 To run a code coverage test only on selected classes, functions or lines in a script,
 enter a hash table value with the following keys:
 
--- Path (P)(mandatory) <string>. Enter one path to the files. Wildcard characters
+-- Path (P)(mandatory) <string>: Enter one path to the files. Wildcard characters
    are supported, but only one string is permitted.
+-- IncludeTests (T) <bool>: Includes code coverage for Pester test files (*.tests.ps1).
```
nohwnd: Remove the T alias here.

Functions/Coverage.ps1
fflaten: Result is the same (autodetect only used for path input), so I agree that it should be moved to keep it clean.

Pester.psm1
fflaten: I expected users to understand it since the first line in the help mentions _strings_ or hash table as valid input. I agree that it can be improved, but I'm unsure about this:

> When you provide a test file directly, code coverage will be measured.

I might assume that `@{ Path = "file.tests.ps1" }` would also include tests, which it won't when hashtables are used. Should it? I consider the dictionary as advanced input where I expect it to be set if needed. How about this?

> Pester test files are by default excluded from code coverage when a directory is provided. When you provide a test file directly using string, code coverage will be measured. To include tests in code coverage of a directory, use the dictionary syntax and provide IncludeTests = $true option, as shown below.

nohwnd: Yup, agree.

fflaten: Updated help

nohwnd:
@fflaten looks all done to me, am I right?

fflaten:
I believe it's good to go, unless we're waiting on a reply from @cgnuechtel

nohwnd:
@fflaten it can be patched later if needed. Thanks!

Issue #360, opened 2015-05-25
How to stop Pester at first failed test?


hero101111:
I need Pester to stop executing tests when one of them fails, but I cannot find any information on this. Can Pester do this?

dlwyatt:
There's no option for that at the moment.

hero101111:
I see. Maybe in a future version then :+1: Thanks for the quick answer!

austoonz:
I thought this is what happens if there's multiple assertions in the same 'It' block.
If one of those assertions fails, the remaining assertions won't be processed.

This might not be what the question relates to though.

nohwnd:
It fails on first failed assertion (or any other terminating exception for that matter), but the whole test suite does not fail on first exception.  So stopping the while suite on first failure is not possible in Pester.

W1M0R:
This would be a great addition.

nohwnd:
@W1M0R Thanks for the comment. Could you please expand on what benefits/use cases do you see or have for it ?

flatwound:
For me it would be useful that pester exits / throws after the first failed test, as we are using a bunch of sequential tests (e.g. install our product, check if it was installed correctly, logon to the admin interface etc...). This whole chain of tests assumes that the previous test has completed successfully. In the case of installation failed, it makes no further sense to test if we can logon to the admin interface or perform any other tests...

kilasuit:
@flatwound - You can always wrap the testing into seperate test scripts for each stage and have a controller script that only executes them if the prior test script for the previous stage passes.

Personally though it really seems like you should be using DSC with its notion of DependsOn.
That would in my opinion be better suited to your needs.

davidwallis3101:
Did this ever get added as a feature? - I don't really want to split my testing out in to seperate files, but I have a prereq set of steps I wish to pass and also like above in some cases I would like some tests to assume the previous completed successfully.

nohwnd:
No it did not. We might add it to V4. Here is a quick workaround that seems to work quite fine. It would need to check the internal state and print the actual description of the failed test to be useful, but a a quick workaround it might be good enough for you so I am sharing it anyway.
```powershell
$script:__anyFailed__ = $false
function Assert-PreviousTestsSucceeded {
    if ($script:__anyFailed__){
        throw "Some previous test failed"
    }
}

function MyShould  {
    process {
    try {
       $_ | Should $args[0] $args[1] $args[2]
    }
    catch {
        $script:__anyFailed__ = $true
        throw
    }}
[...]

PR #1198, opened 2019-01-05
Stop execution of scenario if one step failed/inconclusive (#360)

cgnuechtel:
As described in #360 it does not make any sense to execute the next steps of a scenario if the current step fails or is inconclusive.

(A step is a Given/When/Then clause.)
(Also with the current pull request Pester should behave more like Cucumber.)

### DIFF
--- a/Functions/Gherkin.Tests.ps1
+++ b/Functions/Gherkin.Tests.ps1
@@ -369,11 +369,11 @@ Describe "Check test results of steps" -Tag Gherkin {
     }
 
     It "Test result 11 is correct" {
-        $testResults[10] | Should -Be 'Passed'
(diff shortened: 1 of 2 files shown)


nohwnd:
Fix #1198

cgnuechtel:
very fast merge 😮
thank you

nathan-given-health-catalyst:
@cgnuechtel - is this available now? How do I use it?

cgnuechtel:
@nathan-given-health-catalyst my fix was for Gherkin execution, not for PSpec. But I guess for PSpec it was [fixed](https://github.com/pester/Pester/issues/360#issuecomment-324476249) earlier.

PR #1199, opened 2019-01-06
Modify package and class names in jacoco report to add support for coverage gutters extension

ThePSAdmin:

This PR slightly modifies the jacoco detailed report so it's more java idiomatic with regards to how class and package names are formed and their relation to the folder structure.  This fix makes it easier for code coverage analysis tools to find the reported on files, for instance the VSCode Extension "Coverage Gutters" now works when creating a coverage report with pester called "cov.xml" in the projects root.


nohwnd:
@ThePSAdmin Thanks for the PR, it fails on Windows PowerShell, please click the details and have a look at the problem. If you cannot figure it out let me know I will have a look. If you push to the same branch the build will automatically run (takes at max 10 minutes to complete).  

Once it builds is there some guide that would help me test it with Coverage Gutters, pls? 

@Thilas opinions please :)

ThePSAdmin:
That's fixed now. Turns out, it was this issue https://github.com/PowerShell/PowerShell/issues/2459

To test the coverage gutters, install this extension: https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters

Open the pester project in vscode, click the "watch" icon in the status bar at the bottom, then generate a coverage file with the command:

```powershell
Invoke-Pester .\Functions\Coverage.tests.ps1 -CodeCoverage "C:\dev\pester\functions\Coverage.ps1" -CodeCoverageOutputFile cov.xml
```

Once the report is generated, gutters should appear in vscode:

pester_coverage

nohwnd:
@ThePSAdmin Great thanks for the info. I will test it in the evening.

Do you also happen to use Azure DevOps ? Does the output still work there even after modification?

ThePSAdmin:
I do, but I haven't tested this. I'll try to do it this evening.

Thilas:
@nohwnd, I'll look at this by tonight as well.

Thilas:
I just made a bit of testing and so far all this seems very fragile to me (although I love th:

- at first, I generated the code coverage of `Coverage.ps1` only but Gutters couldn't display ; the reason was because I had a build of Pester in `build` directory (i.e. a copy of Pester files) so Gutters found multiple files that match paths mentionned in cov.xml
  - I would expect these path to be relative to the xml file location
  - anyway, after removing the copy, Gutters was working fine on `Coverage.ps1`
- I tried then to generate the coverage of `Coverage.ps1` + `Pester.psm1` files but I couldn't: I get some "file path not found" errors
  - I tried to make some changes at the xml file but unsuccessfully (I can't spend more time on it right now). @ThePSAdmin, any idea?

Otherwise regarding the PR itself, it really is mainly about convention we want to follow. FYI, I tested my use case (generating a report using [ReportGenerator](https://github.com/danielpalme/ReportGenerator)) and it still works fine. And I'm pretty sure, it will still works with Azure DevOps (even though I couldn't test it yet).

nohwnd:
Is there any standard for how the output file should look like? Any most popular tool that produces this output? I would follow how they do it in that case.

ThePSAdmin:
I think the issue is that Jacoco is a coverage tool designed for java, and classes/namespaces are pretty tightly coupled with the directory structure. This means that there has to be at least one subdirectory (i.e a namespace of just a single word)  where the source files are located to comply with the conventions. 
I know very little about Java, so I gathered this information by reading stuff like [this](http://www.ntu.edu.sg/home/ehchua/programming/java/j9c_packageclasspath.html) when I was researching how to get the gutters to work. There is also an example of a java project in the coverage gutters repo [here](https://github.com/ryanluker/vscode-coverage-gutters/tree/master/example/java).

I'm not sure we can get much better than this using the jacoco format because of these constraints. Using something like lcov would probably be easier as it is more language agnostic. I wouldn't mind giving it a go implementing it, but having two kids under 3 years old puts some hard constraints on my community work, so it wouldn't be tomorrow :)

nohwnd:
If it shows reasonable stuff in vsts and works with the gutters then great. On the other hand if the gutters are finicky, or we would be breaking the intergration with VTST then VSTS should win. Probably some research is needed first: What is the most widely used format for CC? Can we use that?  :)

ThePSAdmin:
I just had a look at the lcov spec and it looks pretty easy to do. I might even be able to do a poc tomorrow and see how the results compare in Azure DevOps.

Thilas:
More work is needed to get gutters works with our JaCoCo implementation (e.g. having one `package` per folder or something like this). Right now, even with @ThePSAdmin's PR, it is not enough (as mentioned earlier).

So either we do that or as suggested by @ThePSAdmin, we can try some other format. I already worked with Cobertura. And even if it comes from java world too, it is not as tight to it as JaCoCo. Bonus: it is supported by both gutters and VSTS.

Though, I'd prefer to work on fixing current JaCoCo implementation using this PR. It should not take too long I think.

Thilas:
FYI, I'm pretty sure lcov is not supported by VSTS...

Thilas:
I looked into gutters extension source code and I now understand what it expects. Given that, the main issue we have is with powershell files that are at the root of the workspace (e.g. `Pester.psm1`). Here is why...

Gutters expects that `{package}/{sourcefilenameWithoutExtension} == {class}` and then looks for such files in the workspace relatively to its root. For instace:

- for `Coverage.ps1`, the package is `Functions`, the class must be `Functions/Coverage` and then gutters looks for a file `Functions/Coverage.ps1` in the workspace which can be found because it is a relative path that is correct.
- but for `Pester.psm1`, there is no way to have a package that leads to a valid relative path except for 2 ways I could think of:
  1. package = `.`, so class = `./Pester` and path `./Pester.psm1` is valid
  2. or package = `../{workspaceRoot}`, so class = `../{workspaceRoot}/Pester` and path `../{workspaceRoot}/Pester.psm1` is also valid

Either way (`.` or `../{workspaceRoot}`), the solution could be applied (or not) on all packages. For instance, on `Coverage.ps1`, package could be `./Function`, class `./Function/Coverage` and so on...

Some questions then:

1. what solution should be chosen: `.` or `../{workspaceRoot}`?
2. should this solution be applied to all packages or only for the root package?
3. should we choose another way and ask gutters for an enhancement?

To me, we should go with option 1 (`.`) and apply it only on root package. Coverage of `Pester.psm1` + `Coverage.ps1` files would then looks like this:
```xml
<report>
    <package name=".">
        <class name="./Pester" sourcefilename="Pester.psm1" />
        <sourcefile name="Pester.psm1" />
    </package>
    <package name="Functions">
        <class name="Functions/Coverage" sourcefilename="Coverage.ps1" />
        <sourcefile name="Coverage.ps1" />
    </package>
</report>
```

FYI, I did test this with ReportGenerator successfully but not yet with VSTS.

What do you think?

Thilas:
Actually, I thought about it and I would recommend recommend to stick with option 3 (which is close to @ThePSAdmin's proposition on this PR):

- determine the top common path
- use it to build JaCoCo's packages, classes and paths (i.e. 1 package per directory)
- it can lead then to coverage files that cannot be used by gutters (when a powershell script is at the top of the root workspace)
- in other cases, it should work fine with gutters
- in any cases, it is valid JaCoCo files (to be confirmed though on VSTS at least)

Here is a sample of what it would look like (my root git folder is `Pester`):
```xml
<report>
    <package name="Pester">
        <class name="Pester/Pester" sourcefilename="Pester.psm1" />
        <sourcefile name="Pester.psm1" />
    </package>
    <package name="Pester/Functions">
        <class name="Pester/Functions/Coverage" sourcefilename="Coverage.ps1" />
        <sourcefile name="Coverage.ps1" />
    </package>
</report>
```

I think it makes more sense this way, and we can still submit an idea to gutters teams in order to handle such case.

nohwnd:
Thanks for figuring this out, option 3 sounds good.

nohwnd:
@ryanluker
- it can lead then to coverage files that cannot be used by gutters (when a powershell script is at the top of the root workspace)

Could you please confirm that we are hitting limitation in coverage gutters? I was looking for a similar issue in your repo but could not find any.

ThePSAdmin:
Sorry about the commit history. I forgot to pull and rebase before pushing. Any tips to remediate would be appreciated :)

Anyway, I've changed it now to fit option 3. Since all the test scripts in the tests are in the root folder, the tests pass. But I'm guessing we would probably need another test script in a subfolder to test this fully. I've tested with coverage gutters and it works.

nohwnd:
@ThePSAdmin no problem I will squash and merge it a single commit anyway. 🙂 @Thilas please validate one more time if you can.

ryanluker:
@nohwnd 
gutters should be fine consuming these jacoco files as long as then are present in one of the open workspace projects in visual studio code and have a matching coverage file name (ie: cov.xml). 

In the future there will also be a manual option so if people store their coverage files outside their workspace folders they can also use the extension https://github.com/ryanluker/vscode-coverage-gutters/issues/178. I believe that was what was being discussed by 
```it can lead then to coverage files that cannot be used by gutters (when a powershell script is at the top of the root workspace)```.

**Comment by Thilas** [comment]
OK, I'll check later this evening.

File: Functions/Coverage.ps1 [hunk 1]
```diff
@@ -3,7 +3,8 @@ if ($PSVersionTable.PSVersion.Major -le 2)
     function Exit-CoverageAnalysis { }
     function Get-CoverageReport { }
     function Write-CoverageReport { }
-    function Enter-CoverageAnalysis {
+    function Enter-CoverageAnalysis
```
Thilas: I guess you used some automatic formatting but it created a lot of changes and introduced discrepancies with the rest of the source code. Could you undo that?

Functions/Coverage.ps1
```diff
@@ -650,163 +651,196 @@ function Normalize-Path
...
     [long] $endTime = [math]::Floor((New-TimeSpan -start $nineteenSeventy -end $now).TotalMilliseconds)
     [long] $startTime = [math]::Floor($endTime - $PesterState.Time.TotalMilliseconds)
 
-    $package = @{
-        Classes = [ordered] @{ }
-        Instruction = @{ Missed = 0; Covered = 0 }
-        Line = @{ Missed = 0; Covered = 0 }
-        Method = @{ Missed = 0; Covered = 0 }
-        Class = @{ Missed = 0; Covered = 0 }
+    $folderGroups = $PesterState.CommandCoverage | Group-Object -Property {
```
Thilas: I'm pretty sure you should use `& $SafeCommands['Group-Object']` here.

Functions/Coverage.ps1
```diff
@@ -650,163 +651,196 @@ function Normalize-Path
...
+        Split-Path $_.File -Parent
```
Thilas: Same thing here: `& $SafeCommands['Split-Path']`

Functions/Coverage.ps1
```diff
@@ -650,163 +651,196 @@ function Normalize-Path
...
+        foreach ($file in $package.Classes.Keys)
         {
-            $null = Add-XmlElement $sourceFileElement 'line' -Attributes ([ordered] @{
-                nr = $line
-                mi = $class.Lines.$line.Instruction.Missed
-                ci = $class.Lines.$line.Instruction.Covered
-            })
+            $class = $package.Classes.$file
+            $sourceFileElement = Add-XmlElement $packageElement 'sourcefile' -Attributes ([ordered] @{
+                    name = (Split-Path -Path $file -Leaf)
```
Thilas: Same here: `& $SafeCommands['Split-Path']`

Functions/Coverage.ps1
```diff
@@ -650,163 +651,196 @@ function Normalize-Path
...
-        $package.Classes.$file.Instruction.Covered += $covered
-        $package.Classes.$file.Methods.$function.Instruction.Missed += $missed
-        $package.Classes.$file.Methods.$function.Instruction.Covered += $covered
-        $package.Classes.$file.Lines.$line.Instruction.Missed += $missed
-        $package.Classes.$file.Lines.$line.Instruction.Covered += $covered
+        $packageList.Add($package)
     }
 
     $commonParent = Get-CommonParentPath -Path $CoverageReport.AnalyzedFiles
+    $commonParentLeaf = Split-Path $commonParent -Leaf
```
Thilas: Same here: `& $SafeCommands['Split-Path']`

Functions/Coverage.ps1
```diff
@@ -650,163 +651,196 @@ function Normalize-Path
...
-        })
+        foreach ($file in $package.Classes.Keys)
+        {
+            $class = $package.Classes.$file
+            $classElementRelativePath = (Get-RelativePath -Path $file -RelativeTo $commonParent).Replace("\", "/")
+            $classElementName = "{0}/{1}" -f $commonParentLeaf, $classElementRelativePath
+            $classElementName = $classElementName.Substring(0, $($classElementName.LastIndexOf(".")))
+            $classElement = Add-XmlElement $packageElement 'class' -Attributes ([ordered] @{
+                    name           = $classElementName
+                    sourcefilename = (Split-Path -Path $classElementRelativePath -Leaf)
```
Thilas: Same here: `& $SafeCommands['Split-Path']`

Thilas:
I did a quick review. I also did some test and results looks good to me.

Thilas:
@ryanluker, what I meant with my comment is that for a JaCoCo file that looks like this:
```xml
<report>
    <package name="PesterRoot">
        <class name="PesterRoot/Pester" sourcefilename="Pester.psm1" />
        <sourcefile name="Pester.psm1" />
    </package>
    <package name="PesterRoot/Functions">
        <class name="PesterRoot/Functions/Coverage" sourcefilename="Coverage.ps1" />
        <sourcefile name="Coverage.ps1" />
    </package>
</report>
```
if `Pester.psm1` is located at the root of the workspace (let's say root is at `C:\PesterRoot`), then gutters will try to look for a file whose relative path is `PesterRoot/Pester.psm1` which does not match any file in the workspace (because the actual relative path is `Pester.psm1`.

Thilas:
There are still quite some whitespaces-only changes but except from that it looks good to me.

nohwnd:
@ThePSAdmin Merged, thanks.

Issue #1194, opened 2019-01-02
Feature: Allow users to specify custom output format


davinci26:
Hey everyone,

Many thanks for your amazing product! I would like to propose a new feature to Pester

### 1. General summary of the issue
Allow users to specify a custom `$OutputFormat` function when invoking Pester. The invocation will look something like:
```powershell
Invoke-Pester -EnableExit -OutputFormat 'Custom' -OutputFunction Export-UserDefinedReport
```
### 2. Possible Solution

This feature will allow the user to specify a custom output format. This can be achieved with the following changes:

1. Expand the `outputformat` parameter to also accept the literal "Custom". The `outputformat` parameter will default to NUnitXml (same as now)
2. Add an additional input parameter ` $OutputFunction` that allows the user to specify the logging function. The logging function should follow the same contract as the NUnit function:
   `Export-UserDefinedReport -PesterState $PesterState -Path $Path`
3. Modify `function Export-PesterResults` to call the user provided function
### 3. Context
This feature is useful for users that are required to use an output format that is either proprietary or too specific to be added to Pester.

### 4. Implementation
I would love to hear your feedback! When the design is somewhat finalized I can go ahead and implement the feature.

dlwyatt:
I suspect you can already accomplish this by using the -PassThru parameter, and then sending that object on to your function. That would be more resilient to changes in Pester’s internals as well.

nohwnd:
@davinci26 yes, what @dlwyatt says is true. You can get the result object from Pester by specifying the `-PassThru` parameter to `Invoke-Pester`. This object is also used internally to produce the Nunit output. [Here](https://github.com/pester/Pester/blob/2dfb8a810a9bc8fdc02be2ec6ac9a6ec12705c2a/Pester.psm1#L1190) you can see that the $pester object is passed into a function and transformed to the NUnit XML, and [here](https://github.com/pester/Pester/blob/2dfb8a810a9bc8fdc02be2ec6ac9a6ec12705c2a/Pester.psm1#L1204) the same object is returned when `-PassThru` is specified. So the functionality can be implemented both internally and externally. (Some internal properties are filtered out from the object when it's published, but hopefully nothing you need for your output.) 

Please give it a try and if you need more help re-open this issue.

davinci26:
Hey again,

Hopefully I am not becoming annoying :) with my Issues.

[...]

Issue #1087, opened 2018-08-08
NUnit 3.0 Support


rdbartram:
## 1. Description
I would like to bring in support for NUnit 3.0 among other reasons, I just wrote the whole thing to extend the Export-NUnitReport cmdlet to include properties and output fields. I noticed after writing that that we are currently NUnit 2.5 compliant and that means I can't include the output property

## 3. Expected Behavior
should be able to export NUnit 3.0 report

## 4.Current Behavior
Only exports NUnit 2.5

## 5. Possible Solution
I can rewrite the testresults.ps1 to either replace support for 2.5 with 3.0 or we provide a parameter somewhere for the user to choose. 

Have a solution in mind?

https://github.com/rdbartram/pester/

## 6. Context

After the test file has been uploaded in VSTS it is easier troubleshoot when more information is saved in the test file.

nohwnd:
That would be nice. The [OutputFormat](https://github.com/pester/Pester/blob/055bb3e4cfe20a165300f12b882807a35e3f286a/Pester.psm1#L789) allows the user to choose the output format, to be used, please add `NUnitXML30` and implement your changes.

rdbartram:
Understood. I'll start working on it

nohwnd:
@rdbartram Are you still interested in doing this?

nohwnd:
Related #301

nohwnd:
This migration should also address the problem raised in #1083 , should the fail be reported with `Failed` or `Failure`?

rdbartram:
I had this as part of my TestCase Output PR but then removed it for separation of concern. As soon as I get the other PR in and I know that I can get a PR into pester, I'll create the next one. The code is still on my machine working internally

nohwnd:
@rdbartram it's still unclear to me how to do the testcases properly, that's why it's still not merged. If you already have the code for this issue, could you PR that? /cc @davinci26

Issue #301, opened 2015-04-10
NUnit XML does not contain output logging for the cmdlets


theficus:
The NUnit XML that's created by Pester doesn't seem to contain any output logging from tests including information, warnings, or errors. This makes reviewing test runs difficult since while I can tell what passed and what didn't, there's zero fidelity regarding what actually happened when the test ran and that is important information.

I've used NUnit in the past to run unit tests and I know it can contain this information (obviously at the cost of much larger XML result outputs).

Is it possible for Pester to capture output to the host and attach it to the XML results that are created to provide more context about what's happening?

dlwyatt:
Can you provide examples of what you're seeing in the file now, and what you'd like to see?

theficus:
Any Write-Warning/Write-Verbose/Write-Host that runs I'd like to see in the test case output. In the XML it would look something like this:

```xml
<test-case name="..." executed="True" time="1" asserts="0" success="True" result="Success">
    <message>... powershell output goes here ...</message>
</test-case>
```

This way when viewing the test results I can see the actual output of the test execution.

Similarly when an error occurs, those error details including stack trace should be written.

Today it looks like the bare minimum of information is written to the NUnit XML which is ok, but I'd like to see more so it's more in line with what I would expect from test cases run in NUnit or MSTest.

dlwyatt:
I'll think about it, but capturing that output (particularly when we still need to support PowerShell v2) is tricky.  Write-Host, in particular, is a problem (which is why it's generally recommended to not use that command at all, prior to the latest builds of PowerShell v5.)

theficus:
Understood. 

Even if some other command needed to be used for logging I think it's important to have more details captured in the output XML. Pass/fail results are useful for a quick glance but when it comes to auditing what happened you either need to re-run the tests and hope they behave the same way, or capture the full output to a separate file and x-ref between that and the NUnit output which is definitely not ideal. :)

theficus:
If Write-Host would present compatibility problems, capturing things like Write-Verbose/Write-Warning as standard output into the NUnit XML would certainly be valuable as well.

nohwnd:
There is Output element in NUnit 3. In the schema of 2.5 that we use I don't see output. So to be able to do this we probably need to migrate to NUnit 3.  https://github.com/nunit/docs/wiki/Test-Result-XML-Format#output

davinci26:
Is there actively working on this issue? @nohwnd @rdbartram

nohwnd:
[...]

PR #1210, opened 2019-01-12
Improve export

cgnuechtel:
This pull requests improves the export of Pester test results.

I am opening this pull request because of the following comments:

- from fourpastmidnight: https://github.com/pester/Pester/pull/1186#issuecomment-451665563
- from nohwnd: https://github.com/pester/Pester/pull/1186#issuecomment-451670480

(both from PR #1186)

The code has the following characteristics:

- It creates a test report object wich contains a tree of test results which are suitable for the export in different formats
- 99 percent of the code is equal for PSpec and Gherkin
- It refactors the existing NUnit export with maintenance in mind
- Also it fixes some NUnit report bugs for Gherkin
- It introduced a shared test code module for XML tests (which will be removed on release)

Also, I carefully compared the output of my changes with the last output of the master.
It is the same (without time values).

I am still convinced, that Pester should support a small amount of some standard outputs say like Cucumber.

I think the following formats would be fine:

- NUnit 2
- Nunit 3
- HTML (single page)
- JSON (for tools like Donut)

More output formats could be provided as extra modules.

The following issues/pull requests I have been found about output/NUnit results.

- #301 
- #1087
- #1194
- #1186
- #1089

Hopefully I'm not too overwhelming.
The code I have been provided can be uses as it is.
Or can used as starting point of discussion about the topic.

By the way: the object which is returned by `passthru` parameter seems not to be sufficient for me so I used the internal Pester state object.

Please give feedback to the code!
I would highly appreciate suggestions for improvements!

### DIFF
--- a/Dependencies/TestUtilities/TestUtilities.psm1
+++ b/Dependencies/TestUtilities/TestUtilities.psm1
@@ -1,0 +1,24 @@
+# Helper function to get a XML node from a XPath expression
+function Get-XmlNode($xmlDocument, $xPath) {
+    return (Select-Xml -Xml $xmlDocument -XPath $xPath | Select-Object -ExpandProperty Node)
+}
+
+# Helper function to get the inner text of a XML node from a XPath expression
+function Get-XmlInnerText($xmlDocument, $xPath) {
+    return (Get-XmlNode $xmlDocument $xPath).InnerText
+}
+
+# Helper function to get the value of a XML node from a XPath expression
+function Get-XmlValue($xmlDocument, $xPath) {
+    return (Get-XmlNode $xmlDocument $xPath).Value
+}
+
+# Helper function to get the number of children of a XML node from a XPath expression
+function Get-XmlCount($xmlDocument, $xPath) {
+    return (Get-XmlNode $xmlDocument $xPath).Count
+}
+
+# Special helper function to get the text of the directly following pre element
+function Get-NextPreText($xmlDocument, $xPath) {
+    return Get-XmlInnerText $xmlDocument "$xPath/following-sibling::*[position()=1][name()='pre']"
+}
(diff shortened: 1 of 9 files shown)


nohwnd:
@cgnuechtel Thanks for the PR, it will take me some time to reivew it.

clcaldwell:
@cgnuechtel 

I am not sure if this was intentional - but there is some interesting behavior you have introduced now that -OutputFormat is [string[]] .

`Invoke-Pester -Path .\UnitTests\ -OutputFormat NUnitXml,html -OutputFile .\tmp\results.xml `
Creates the result.xml, but ignores the html report.

`Invoke-Pester -Path .\UnitTests\ -OutputFormat NUnitXml,html -OutputFile results.xml,result.html `
Creates both files, as expected.

`Invoke-Pester -Path .\UnitTests\ -OutputFormat NUnitXml,html -OutputFile results.html,result.xml `
And
`Invoke-Pester -Path .\UnitTests\ -OutputFormat html,NUnitXml -OutputFile results.xml,result.html `
Creates both files, but the contents are switched (xml results in html <> html results in xml).

Similarly:
`Invoke-Pester -Path .\UnitTests\ -OutputFormat NUnitXml -OutputFile results.html `
Creates the xml file, but outputs html results into it.

So, I guess the question is: If you are going to offer multiple output formats, should you validate that the results are getting written into a valid file - or do some correction for mis-orded inputs on -OutputFile?

cgnuechtel:
@clcaldwell As I remember, that was intended. The output file is just a name and has no special checks.

@nohwnd Any plans to include this pull request or is it outdated now?

clcaldwell:
@cgnuechtel Just to clarify what I mean with this:
If -OutputFormat only accepts 1 value, it would make sense to just output into whatever file is given (as the current behavior is). However, now that you are accepting a collecting of strings as input (in this PR), the outputs are no longer intuitive.

nohwnd:
Thank your for the PR, as I said in my previous comments, I am not very interested in adding another output because I would have to support it in the future. Please feel free publishing this on your own as a module.

PR #1220, opened 2019-01-20
Filter by scriptblock position

nohwnd:
ScriptBlockFilter filters top-level describes by position of the Fixture instead of filtering by name, this is for plugin usage to accomodate tests that expand properties in the describe name, and so AST and runtime values in the name do not match.

See https://github.com/PowerShell/PowerShellEditorServices/issues/827

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,35 +1,39 @@
 ## 4.5.0 (January 10, 2018)
 
- - Add support for code coverage in VSCode [GH-1199]
- - Remote Tests from CodeCoverage by default [GH-1196]
- - Add short-circuting to Ghrerkin scenario [GH-1198]
- - Fix error in parallel TestRegistry [GH-1197]
- - The rest listed in betas below
+- Add support for code coverage in VSCode [#1199](https://github.com/pester/Pester/issues/1199)
+- Remote Tests from CodeCoverage by default [#1196](https://github.com/pester/Pester/issues/1196)
+- Add short-circuting to Ghrerkin scenario [#1198](https://github.com/pester/Pester/issues/1198)
+- Fix error in parallel TestRegistry [#1197](https://github.com/pester/Pester/issues/1197)
+- The rest listed in betas below
 
 ## 4.5.0-beta2 (January 5, 2018)
 
-- Add more info to code coverage passthru [GH-1190]
-- Add scenario for every example in Gherkin [GH-1168]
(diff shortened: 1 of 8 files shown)


nohwnd:
```powershell 
# file run.ps1
get-module pester | remove-module
import-module .\Pester.psd1

# runs a and b but not e
$o = New-PesterOption -ScriptBlockFilter @(
    # describe a
    @{ Path = "C:\Users\nohwnd\Desktop\filtering on sb position.ps1"; Line = 1 }
    # describe b
    @{ Path = "C:\Users\nohwnd\Desktop\filtering on sb position.ps1"; Line = 5 }
    # describe c, but it does not apply on child blocks
    @{ Path = "C:\Users\nohwnd\Desktop\filtering on sb position.ps1"; Line = 6 }
)
invoke-pester -Script "~\Desktop\filtering on sb position.ps1" -PesterOption $o

```
```powershell
# file filtering on sb position.ps1

nohwnd:
I don't see anything in my changes that should cause such a bug, and I don't see it running recursively in your output. So we understand each other what this change is supposed to do? I just added a filter that works like the TestName filter, but it filters on different criteria. So the files that you give it will still run, it does not filter the paths until the scriptblock is reached, if you want one file to run you should still give it just a one file to run.

bergmeister:
It was a mistake that I made locally when i wasn't aware that I had to specify the path inside the filter as well, sorry, I edited my comment.

bergmeister:
After I upstreamed my PSES and vs-code branches, I have it working now end to end. The big question now to @rkeithhill is if we need to have a check on the Pester version or if we should download Pester the minimum required v ersion as a fallback similar to how the extension does it for PSSA?
image

rkeithhill:
I think we'd need to add the version check regardless of whether Pester ships with the extension.  A user could have loaded a specific (older) version of Pester into the PSIC that doesn't support this.

RE shipping Pester with the question.  That is probably something @rjmholt and/or @TylerLeonhardt should answer.

TylerLeonhardt:
Yeah we tack on the end of the PSModulePath so even _if_ we ship Pester, a version check is the right way to go here.

I don't think we should ship Pester personally but I would love to use the extension to get people on the newer version in an semi-unobtrusive way.

bergmeister:
@nohwnd Thanks, I created [this](https://github.com/PowerShell/vscode-powershell/issues/1710) issue for the technical discussion of how to implement it in the extension

PR #1223, opened 2019-01-21
Remove the Quiet parameter

it-praktyk:

The pull requests removes - deprecated in the Pester v4 - the Quiet parameter.


nohwnd:
The parameter is currently deprecated, and removing it would be a breaking change. Why change it now and not wait for v5?

PR #1230, opened 2019-01-30
WIP: Improve logic for removal of symlinks during TestDrive tear down.

JamesWTruher:

Under some circumstances, it appears that the delete method `FileInfo/DirectoryInfo` will throw which causes test failures. This change adds an invocation of  Remove-Item as a last-ditch effort to remove a symlink when the items' Delete method fails. I've also enabled the commented out tests for symlink removal under the appropriate circumstances by calculating whether the user is elevated on Windows. On non-Windows, symlink creation does not require elevation.

### DIFF
--- a/Functions/TestDrive.Tests.ps1
+++ b/Functions/TestDrive.Tests.ps1
@@ -174,22 +174,24 @@ InModuleScope Pester {
 InModuleScope Pester {
 
     Describe "Clear-TestDrive" {
-        # not all environments need admin privileges to create symlinks
-        if ($PSVersionTable.PSVersion.Major -lt 6 -or $IsWindows) {
+
+        # Determine whether we should skip these tests
+        # run only on PSVersion -ge 5 (since that's when Symlinks were available)
+        # if on Windows, determine whether we are administrator
+        # run the tests on non-Windows platforms, since symlinks don't require elevation there.
+        $SkipTest = $true
+        if  ( (GetPesterOs) -eq "Windows" -and (GetPesterPSVersion) -ge 5 ) {
             $windowsIdentity = [Security.Principal.WindowsIdentity]::GetCurrent()
             $windowsPrincipal = new-object 'Security.Principal.WindowsPrincipal' $windowsIdentity
             $IsAdmin = $windowsPrincipal.IsInRole("Administrators") -eq 1
+            # skip tests when not admin
             $SkipTest = ! $IsAdmin
         }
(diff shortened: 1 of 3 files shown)


Functions/TestDrive.Tests.ps1
```diff
@@ -171,32 +171,61 @@ InModuleScope Pester {
...
 # # to test this is imho not a good idea. Tested on powershell 2 and 5 -- nohwnd
-# InModuleScope Pester {
-#     Describe "Clear-TestDrive" {
+InModuleScope Pester {
+
+    Describe "Clear-TestDrive" {
+        BeforeAll {
+            $SkipTest = $true
+            # test on non-windows PowerShell Core
+            if ( $PSVersionTable.PSVersion -ge "6.0" -and $IsWindows -eq $false ) {
```
nohwnd: Try using `GetPesterPsVersion` and `GetPesterOs` to avoid fails. It works properly on Powershell 2. 
```
$SkipTest = (GetPesterOs) -ne "Windows" -and (GetPesterPsVersion) -ge 6
```

JamesWTruher: thanks!

nohwnd:
Marked it as `WIP`, please let me know when it is ready. :) If you need help passing the tests on v2 I can help you.

I briefly looked through the code: 
- One test asserts that it does not throw, it would be better that it checked that the symlinks are gone, as the original test did. 
- `IsInRole` produces bool, comparing it to 1 to get bool is confusing.
- please use `-not` instead of `!`

JamesWTruher:
@nohwnd I've found myself a Win7/PSv2 system but can't seem to get the failure in the `Describe` block as what's showing up in CI. https://nohwnd.visualstudio.com/Pester/_build/results?buildId=217 - do you have insight as to what's happening here?

nohwnd:
If you cannot trigger the error try running `.\bin\pester.bat` that one runs in powershell strict mode, same as Pester does on the build server. I just added a flag to the build on master so it does not filter the call stack on failure, so we can see in the log unaltered callstack. 

Judging just by the error, it can't find the data that `Should` failure adds on the error record, but I don't see any direct connection between your code and this. Upgraded your branch to the latest master, hopefully it gives more useful info.

nohwnd:
Okay, fixed the failure, cleaned up the code a bit, and now it all builds. Merging :) Thanks!

nohwnd:
@JamesWTruher You are welcome. Thanks for your contributions :)

PR #1231, opened 2019-02-01
Add a new exception and clarify exception wording to It, Describe, and Context

### LINKED ISSUE #496: Error message is not helpful when missing Name parameter value
When you use Describe/Context/InModuleScope without the required name string, you get an error message that says there's no scriptblock. 

This is a fantastic error message when you've misplaced your opening brace, but it doesn't help when the name is missing. 

Please change it so it detects the missing name and returns a helpful error message. Thanks!

jd-porter:
Fix #496 
Add new exceptions and clarify exception wording when a parameter to It, Describe, or Context is missing. Clarify for the use whether a name or script block is missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build/
 Test.xml
 vendor/packages/
+.vscode/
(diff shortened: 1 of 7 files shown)


nohwnd:
@jd-porter You are no longer just a fan, you are a full blown contributor now! 🎉 🍾

jd-porter:
Not yet, I probably should get a twitter account, though :)

PR #1233, opened 2019-02-01
Allow mocks to work with conflicting parameter names

### LINKED ISSUE #1007: SessionStateUnauthorizedAccessException when Mocking Get-Module
I came across this while trying to repro #1006.

## 1. Repro

```PowerShell
function f { Get-Module foo }

Describe something {
    It 'mock Get-Module' {
        Mock Get-Module -Verifiable { 'mocked' }
        f
        Assert-MockCalled Get-Module
    }
}
```

```none
Pester version     : 4.3.1 C:\Users\un1\Documents\WindowsPowerShell\Modules\PEster\Pester.psm1
PowerShell version : 5.1.14409.1012
OS version         : Microsoft Windows NT 6.3.9600.0
```

## 3. Expected Behavior

Test should pass. 

## 4.Current Behavior

```none
Describing something
  [-] mock Get-Module 64ms
    SessionStateUnauthorizedAccessException: Cannot overwrite variable PSEdition
 because it is read-only or constant.
    at f, C:\Users\un1\Desktop\test2.ps1: line 1
    at <ScriptBlock>, C:\Users\un1\Desktop\test2.ps1: line 6
```

renehernandez:

Fix #1007 

Allow Mock to overcome names in parameters that shouldn't be used by prefixing them with the *_* identifier.


renehernandez:
@nohwnd the powershell 2 run failed before running the tests and it is not clear to why or where exactly. Any inputs?

Functions/Mock.ps1
```diff
@@ -1544,3 +1547,75 @@ function Remove-MockFunctionsAndAliases {
...
+
+    $paramMetadatas = @()
+    $paramMetadatas += $Metadata.Parameters.Values
+
+    foreach ($paramMetadata in $paramMetadatas) {
+        if ($paramMetadata.IsDynamic) {
+            continue
+        }
+
+        $conflictingParams = Get-ConflictingParameterNames
```
nohwnd: Normally I would advocate for using functions like you did, but it is easy enough to find where the variable we are sharing was modified, and the functions add a tiny bit of runtime overhead. BUT mostly they make it look more complicated than it is (IMHO). Please replace `Get-ConflictingParameterNames` with just the variable, and the `Get-ParameterPrefixForConflictingParameters` and the script scoped variable, with just using the  underscore in place. The prefix will not change change once a release is done so we don't need the DRYness.

Functions/Mock.ps1
```diff
@@ -1544,3 +1547,75 @@ function Remove-MockFunctionsAndAliases {
...
+        if ($conflictingParams -contains $paramMetadata.Name) {
+            $paramName = $paramMetadata.Name
+            $newName = "$(Get-ParameterPrefixForConflictingParameters)$paramName"
+            $paramMetadata.Name = $newName
+            $paramMetadata.Aliases.Add($paramName)
+
+            $Metadata.Parameters.Remove($paramName) | Out-Null
```
nohwnd: use `$null =` instead, `Out-Null` is slow, and user might have mocked it, so you'd need to use out-null from the safe commands table which adds even more overhead.

Functions/Mock.ps1
```diff
@@ -1544,3 +1547,75 @@ function Remove-MockFunctionsAndAliases {
...
+    param(
+        [Parameter(Mandatory)]
+        [hashtable]
+        $BoundParameters
+    )
+
+    $parameters = $BoundParameters.Clone()
+    $prefix = Get-ParameterPrefixForConflictingParameters
+
+    Get-ConflictingParameterNames |
```
nohwnd: To avoid conflict with mocked `Where-Object` and `Foreach-Object` you should use the commands from the SafeCommands table, but using pipeline is slow anyway. So a better alternative is to use foreach (keyword) and if, which solves both problems at the same time. Please rewrite it.

nohwnd:
@renehernandez Thanks for your PR. 🙂 You have to use `Mandatory = $true` in posh v2, I described it better in the code review.

Functions/Mock.ps1
renehernandez: Well I introduced the `Get-ConflictingParameterNames` so I can mock it on the tests and use a different array of conflicting values, because if it will fail to run some of the test if *PSEdition* value is used

nohwnd: @renehernandez okay, in that case can you get rid of only the prefix function, please?

Functions/Mock.ps1
renehernandez: Done

Functions/Mock.ps1
```diff
@@ -1544,3 +1547,75 @@ function Remove-MockFunctionsAndAliases {
         & $script:SafeCommands['Remove-Item'] "function:/$($bootstrapFunction.Name)"
     }
 }
+
+
+function Repair-ConflictingParameters {
+    [CmdletBinding()]
+    [OutputType([System.Management.Automation.CommandMetadata])]
+    param(
+        [Parameter(Mandatory)]
```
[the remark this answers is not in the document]
renehernandez: Done

Functions/Mock.ps1
renehernandez: Done

Functions/Mock.ps1
renehernandez: Done

PR #1235, opened 2019-02-03
Added -like operator for Should-Throw

### LINKED ISSUE #1003: Should -Throw wildcard charachters

It would be nice if the following could be used:
```powershell
{throw 'this is a test'} | Should -Throw "*is*test"
```

```powershell
Pester version     :  
PowerShell version : 5.1.14409.1012
OS version         : Microsoft Windows NT 6.2.9200.0
```

## 3. Expected Behavior

```powershell
Describe 'Test Should -Throw' {
    It '1' {
        { throw "This is a test" } | Should -Throw 'This is a test' # Test will pass
    }
    it '2' {
        {throw 'this is a test'} | Should -Throw 'test' # Test will pass
    }
    it '3' {
        {throw 'this is a test'} | Should -Throw "*is*" # Test will not pass
    }
    it '4' {
        {throw 'this is a test'} | Should -Throw "*is*test" # Test will not pass
    }
}
```

It would be great if all these tests could be evaluated as True.

## 4.Current Behavior

Test 3 and 4 fail.

jd-porter:
Added -Like parameter to Should-Throw, allowing wildcard patterns. Users can now write ``` Should -Throw -Like "*pattern*" in tests. Fix #1003

### DIFF
--- a/Functions/Assertions/PesterThrow.Tests.ps1
+++ b/Functions/Assertions/PesterThrow.Tests.ps1
@@ -80,6 +80,29 @@ InModuleScope Pester {
             }
         }
 
+
+        Context "Matching error pattern" {
+            It "given scriptblock that throws exception with the expected pattern it passes" {
+                $expectedErrorMessage = "expected error message"
+                $expectedErrorPattern = "*error*"
+                { throw $expectedErrorMessage } | Should -Throw -like $expectedErrorPattern
+            }
+
+            It "given scriptblock that throws exception with the expected pattern in UPPERCASE it passes" {
+                $expectedErrorMessage = "expected error message"
+                $expectedErrorPattern = "*error*"
+                $errorMessage = $expectedErrorMessage.ToUpperInvariant()
+                { throw $errorMessage } | Should -Throw -like $expectedErrorPattern
+            }
+
+            It "given scriptblock that throws exception with a non-matching pattern it fails" {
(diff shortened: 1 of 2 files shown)


jd-porter:
Looks like some tests broke in this last commit -- I'll fix them up and resubmit.

nohwnd:
@jd-porter You don't have to create a new PR when you break the code, PR is just a pointer to a branch and it's totally okay having many commits in one PR. I squash merge them all in one commit and edit the resulting message anyway. :)

I'd recommend making a new branch for the functionality (we can see that referred to as a feature branch or a topic branch). This way your master is not "locked" to this PR.

You do that by `git checkout -b"add-like-to-should-throw"`. On that  branch you then implement your changes and push (`git push` will show you that there is no remote and will give you the command you need to run in order to push the branch to the origin remote, so run that). Once the branch is online you can make a PR out of it, Github will show you a message with a `Create a PR` button. Once you've done that all the changes done in the branch will be reflected in the PR. You can push more commits, revert, rebase, force push, or whatever else.

nohwnd:
Looking on the original issue, the agreement there is to do this in v5, and the task should be as simple as using `-like` instead of the current comparison. No new switch is needed, we can control whether the comparison behaves like `-like` or `-eq` by using or not using wildcards. So your fix should target the v5 branch, but the tests there don't run yet. Sorry for the confusion. I am gonna un-mark the issue.

PR #1240, opened 2019-02-08
Allow aliases in Parameter Filters blocks

### LINKED ISSUE #1128: Allow use of parameter aliases in Mock -ParameterFilter

Powershell parameters can be defined with aliases:

```powershell

function Get-Something{    
   param(
        [Alias('Path')]
        [string] $PathName)
   #function body
}

Get-Something -Path c:\temp.txt
```

When mocking with `-ParameterFilter` you cannot currently use the alias name to filter, but must use the real parameter name.

```powershell
Mock Get-Something -ParameterFilter {$PathName = "c:\temp.txt"} {return 123} #works
Mock Get-Something -ParameterFilter {$Path = "c:\temp.txt"} {return 456} #fails
```

Pester version : 4.4.0 3.4.0 3.3.5 C:\Program Files\WindowsPowerShell\Modules\Pester\4.4.0\Pester.psd1 C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\Pester.psd1 C:\Program Files\WindowsPowerShell\Modules\Pester\3.3.5\Pester.psd1
PowerShell version : 5.1.15063.1292
OS version : Microsoft Windows NT 10.0.15063.0

## 3. Expected Behavior

`Get-Content` has a parameter named `-Tail` which has an alias `-Last`

```powershell
(Get-Command get-content).ParameterSets[0] | select -ExpandProperty parameters | select Name,Aliases | where Name -eq Tail

Name Aliases
---- -------
Tail {Last}
```

Therefore I would expect this test should pass:

```powershell
It "Uses parameter aliases in Parameter-Filter" {
   Mock Get-Content {return "default-get-content"}
   Mock Get-Content -ParameterFilter {$Last -eq 100} -MockWith {return "aliased-parameter-name"}

   Get-Content -Path "c:\temp.txt" -Last 100 | Should -Be "aliased-parameter-name"
}                                       
```

## 4.Current Behavior

When using an alias in `-ParameterFilter` the variable never has a value.

The output from the above test is

```
[-] Uses parameter aliases in Parameter-Filter 50ms         
  Expected strings to be the same, but they were different. 
  Expected length: 22                                       
  Actual length:   19                                       
  Strings differ at index 0.                                
  Expected: 'aliased-parameter-name'                        
  But was:  'default-get-content'                           
  -----------^    
```

renehernandez:

Fix #1128

### DIFF
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -232,8 +232,9 @@ about_Mocking
                 $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, 'PositionalBinding=$false')
             }
 
-            $repairedMetadata = Repair-ConflictingParameters -Metadata $metadata
-            $paramBlock = [Management.Automation.ProxyCommand]::GetParamBlock($repairedMetadata)
+            # Will modify $metadata object in-place
+            Repair-ConflictingParameters -Metadata $metadata
+            $paramBlock = [Management.Automation.ProxyCommand]::GetParamBlock($metadata)
 
             if ($contextInfo.Command.CommandType -eq 'Cmdlet') {
                 $dynamicParamBlock = "dynamicparam { Get-MockDynamicParameter -CmdletName '$($contextInfo.Command.Name)' -Parameters `$PSBoundParameters }"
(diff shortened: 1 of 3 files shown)


Functions/Mock.ps1
```diff
@@ -232,8 +232,9 @@ about_Mocking
                 $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, 'PositionalBinding=$false')
             }
 
-            $repairedMetadata = Repair-ConflictingParameters -Metadata $metadata
-            $paramBlock = [Management.Automation.ProxyCommand]::GetParamBlock($repairedMetadata)
+            # Will modify $metadata object in-place
+            Repair-ConflictingParameters -Metadata $metadata
```
renehernandez: This is a cosmetic change from the #1007 implementation. We are now making explicit the fact that the metadata object is modified in place

renehernandez:
@nohwnd 21 tests failed in the powershell 3 job. They are related to Gherkin, which are not related to these changes. I looked in the job history and those Gherkin tests failed in the job associated to the previous pull request as well.

renehernandez:
The powershell 2 job is apparently stuck waiting for an agent to run, but these changes shouldn't be validated against powershell 2 anyways. Any inputs?

renehernandez:
Ok I realized now that the pull request should be targeted against v5 branch instead of master. Will close and open a new pull request

renehernandez:
> Ok I realized now that the pull request should be targeted against v5 branch instead of master. Will close and open a new pull request

@nohwnd Actually I am not sure how to proceed here. I checked the latest status of mocking in v5 and it doesn't have yet the previous mocking fixes that I have implemented

nohwnd:
@renehernandez This should target the master, not v5. The agent running powershell v2 is stuck for some reason, I just restarted the VM so hopefully it will come back and start building. Otherwise I need to investigate. The tests & code that is PowerShell 3+ specific should go into an `if` that will use the `GetPesterPSVersion` function (it's called something like that). So it can still work like before on v2. If tests started failing in Gherkin then there is probably some behaviour that has changed by your changes, and it needs to be investigated an addressed. I should be able to help with that later.

nohwnd:
Okay restart did not help, but I solve it. The problem is that VSTS agent is not compatible with PowerShell 2 so what I am doing is patching the app via compatibility shimming to think it runs on v3 and then giving it PowerShell 6 instead, so it is all just a big hack. The problem was that I specified a version of the Agent.Listener executable in the shim and the agent updated and so it started seeing the real powershell version and refused to start up.

renehernandez:
> If tests started failing in Gherkin then there is probably some behaviour that has changed by your changes, and it needs to be investigated an addressed. I should be able to help with that later.

@nohwnd The gherkins testst were already failing in the previous build job before mine, as can be checked [here](https://nohwnd.visualstudio.com/Pester/_build/results?buildId=236). I have also looked at the stacktrace of the error by they don't seem to be hitting anything related to my mocking changes

renehernandez:
@nohwnd Now only the PowerShell 3 job is failing in the Gherkins tests

nohwnd:
@renehernandez okay will have a look, that error looks weirdly unrelated. I can also see that the master started failing, so maybe it's not a problem of this PR at all.

renehernandez:
@nohwnd Any update on the Gherkin issues?? How can I help with that?

nohwnd:
I looked at the failed builds and frankly I have no idea why they started failing. I need to have a closer look on it, hopefully I’ll be able to do that today.

renehernandez:
@nohwnd I guess I should looked into the exceptions sooner. I had ran into that issue before and I found a solution at this [uservoice thread](https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/14915283-job-cmdlets-fail-with-utf-8-codepage)

The problem seems to be that older version of powershell weren't capable of properly managing different encodings with BOM between the current process and a triggered background job

nohwnd:
@renehernandez aaah, that's why there is apparently no change, especially no change in the xmls.

nohwnd:
@renehernandez merged, thanks!

PR #1248, opened 2019-02-16
Shorten error message of string comparison when it is too long

### LINKED ISSUE #532: Error messages with long strings could be displayed better
### Scenario

Tests with pattern

``` powershell
$CreatedLongString | Should Be $ExpectedLongString
```
### Problem

The error message is cumbersome to read (text is too long and spreads multiply lines).
So sometimes I'm trying to split it with `n and then compare each line.
For example in [platyPS](https://github.com/PowerShell/platyPS/blob/master/test/Pester/FullLoop.Tests.ps1#L76)

But then I need to carry around the line number and display it somehow additionaly.
It's becoming hairy pretty fast.
### Proposal

We can take the same approach as xUnit: show a short snippet from the place in the stirng, where it's different (like 32 chars) with additional offset for the beginning of the snippet.

jd-porter:
Modify Be.ps1 to show excerpts of string comparisons when actual or expected values are very long.

Fixes #532


nohwnd:
@jd-porter is this still work in progress? Please add some tests. :)

nohwnd:
@jd-porter Are you interested in adding the tests? If not let me know please, so I can finish this :)

jd-porter:
Can do. I'll finish them up this weekend and resubmit the PR
________________________________

jd-porter:
Updated be.tests and I think we're good to go!

nohwnd:
Merged, thanks!

jd-porter:
oh nice! Thank you for the merge! I saw that the merge had a failure in the Powershell 2 tests last night and wasn't quite sure how I could test that on my system at home... Is that test outdated?

vors:
Nice 🎉 🌮 🇹🇼

nohwnd:
@jd-porter You could do that if you install a virtual machine put some old system in there, like Windows 7 without SP (I think). But the whole problem with the build was that the PowerShell Jobs that we use to test Pester in Pester are a bit flaky on PS2, so the test run failed. I just re-run it and it passed just fine.

PR #1252, opened 2019-02-19
Add alias for extension-less external commands in windows

### LINKED ISSUE #1043: Add alias without .exe when mocking external command

Running a test like this: 

```powershell
Describe "-" {
    It "-" {
        Mock Notepad { Write-Host "I am a mock" }
        notepad
    }
}
```
Will result in notepad being executed because `notepad` will resolve into `notepad.exe` command, but Mock will not pick it up since there is no alias for `notepad`. This is counter-intuitive and currently there is no good way of mocking executables without wrapping them into functions, which is especially painful now that powershell is multiplatform and we usually want to call the executable without explicitly specifying the extension (think `git`).

Pester version     : 4.3.1 4.2.0 C:\Program Files\WindowsPowerShell\Modules\Pester\4.3.1\Pester.psd1 C:\Program Files\WindowsPowerShell\Modules\Pester\4.2.0\Pester.psd1
PowerShell version : 5.1.16299.248

## 3. Expected Behavior

Mock should pickup the call to notepad even though the underlying command is notepad.exe.

## 4.Current Behavior

The executable is executed.

## 5. Possible Solution

Add alias without extension to the mock.

renehernandez:

Fix #1043

### DIFF
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -338,29 +338,7 @@ about_Mocking
 
         $mockTable["$ModuleName||$CommandName"] = $mock
 
-        $scriptBlock = { $ExecutionContext.InvokeProvider.Item.Set("Function:\script:$($args[0])", $args[1], $true, $true) }
-        $null = Invoke-InMockScope -SessionState $mock.SessionState -ScriptBlock $scriptBlock -ArgumentList $Mock.BootstrapFunctionName, $mockScript
-
-        $mock.Aliases += $CommandName
-
-        $scriptBlock = {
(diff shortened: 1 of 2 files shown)


Functions/Mock.ps1
```diff
@@ -338,29 +338,7 @@ about_Mocking
 
         $mockTable["$ModuleName||$CommandName"] = $mock
 
-        $scriptBlock = { $ExecutionContext.InvokeProvider.Item.Set("Function:\script:$($args[0])", $args[1], $true, $true) }
```
renehernandez: I extracted all this block the `Invoke-CommonSetupInMockScope` and then added the new functionality there in order to group common code and reduce the size of the `Mock` function`.

nohwnd:
@renehernandez Thank you!

PR #1260, opened 2019-03-10
Fix .Length mispelling in duplicate certificate thumbprint check in signModule.ps1

leojonathanoh:

Fixes a misspelling in `signModule.ps1`: where the check for duplicate thumbprint(s) is performed, the property to enumerate over the array of certifiates should be `.Length`.


nohwnd:
Merged thanks!

Issue #1163, opened 2018-12-14
Improve Get-ShouldOperator


nohwnd:
`Get-ShouldOperator` shows:
```
NAME
    Should-Be
```
on the top, and does not show aliases. It would be nice to replace the name section with this: 

```
NAME
    Be

ALIASES
    EQ
    EQUAL
    BeEq
```

(the two last aliases don't actually exist on Be)

Related to #878

Stephanevg:
still available? if so, i'll be happy to check this one

nohwnd:
@Stephanevg  yup. Looks like you are the first one.

omiossec:
I will be happy too

Stephanevg:
@nohwnd I think you meant: ```Get-ShouldOperator -Name be``` Shows:

```
NAME
    Be

ALIASES
    EQ
    EQUAL
    BeEq
```

right?

Cause I have this when I use ```Get-ShouldOperator```:

```

Get-ShouldOperator

Name                      Alias
----                      -----
Be                        EQ
BeExactly                 CEQ
BeGreaterThan             GT
...
```

Stephanevg:
I am sorry to ask, but either I didn't fully understand what problem we need to fix here, or I am missing something.

I originally thought that this issue would be as easy as editing the 'comment Based Help' of the BE operator, but It seems there is a bit more to it.

Perhaps you can give me a hint of where I should look? ;)

Let me explain my rationale:


for example, any other 'ShouldOperator' return the same output. 
Where ```.NAME ```shows the ```$InternalName``` (which is also often the name of the file where the function resides).

### Be exactly

```powershell
[DBG]:   Get-ShouldOperator -Name BeExactly

NAME
    Should-BeExactly

SYNOPSIS
    Compares one object with another for equality and throws if the
    two objects are not the same. This comparison is case sensitive.


    -------------------------- EXAMPLE 1 --------------------------

    PS C:\>$actual = "Actual value"

    PS C:\>$actual | Should -Be "Actual value"

    This test will pass. The two strings are identical.




    -------------------------- EXAMPLE 2 --------------------------

    PS C:\>$actual = "Actual value"

    PS C:\>$actual | Should -Be "actual value"

    This test will fail, as the two strings do not match case sensitivity.

```

This actually highlights that the issue is not an isoolated issue, as in opposition of what I thought.

## Process understanding:

It seems to me, that I missing a (obvious) part

The assertion operator nd it Alias is declared in ```be.ps1``` with the following code:

[...]

PR #1272, opened 2019-03-24
WIP DO NOT MERGE : Fix for 1163 - Add alias section to Get-AssertionOperator output

Stephanevg:
# WIP.PLEASE DO NOT MERGE YET.

Hi, this is a fix for https://github.com/pester/Pester/issues/1163

I think I have it working, but the output of the command is not quite how it should be.
Currently, by default, it doesn't include the Alias property.

I tried playing a bit with something around the lines of `$Help | Select Name,Synopsis,Alias,Examples -ExpandProperty Examples` but that out put is not as it was before.

So before I go further in this, it would be cool if you can give some feedback already and perhaps you have an idea / way of how we can return all the properties without 1) destroying the object with a `Select` statement. 2) how to get the output exactly as a regular output from a Get-Help cmdlet (If that is what you want of course)

### DIFF
--- a/Functions/Get-ShouldOperator.ps1
+++ b/Functions/Get-ShouldOperator.ps1
@@ -65,12 +65,16 @@ function Get-ShouldOperator {
 
     END {
         if ($Name) {
+           
             $operator = $AssertionOperators.Values | Where-Object { $Name -eq $_.Name -or $_.Alias -contains $Name }
             $help = Get-Help $operator.InternalName -Examples -ErrorAction SilentlyContinue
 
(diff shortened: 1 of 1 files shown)


nohwnd:
I tried adding the aliases but it seems that it ignores the runtime values when there is a module manifest so I am not sure if this can be done without extensive reflection. 

The approach shown in the PR does not work. 

So I guess the step forward would be just modifying the name in the help output.

nohwnd:
This is still WIP. Closing this up because of clean up. Thanks for spending time on this. :)

PR #1276, opened 2019-03-26
Many gherkin runner fixes and improvements

### LINKED ISSUE #1124: Pester does not print Tables or DocStrings that are part of steps

fourpastmidnight:

This is a replacement PR for #1142 which will fix #1124. This PR simplifies a lot of the code on the Gherkin side, while also improving the console output during the test run so that the output closely resembles the output of a Ruby Cucumber test run. These improvements include:

* A change to the theme to resemble the default Ruby Cucumber theme
* Adopts the output format of the Ruby Cucumber pretty-print formatter
    * The margin is still 2 spaces
    * All lines to be output (such as descriptions for features, scenarios, examples, etc.) are appropriately indented.
* Improvements to Step output
    * Steps have their arguments highlighted.
        * I often found myself using quotes to denote where my arguments to my steps were because there was no indication in the console output. This feature, adopted from Ruby Cucumber, eliminates the need to do that.
    * DocStrings and DataTables are now printed to the console. You can turn this off by specifying `-NoMultiline`.
    * Failed, Undefined, and Pending steps all have a consistent, one-line "error" message and one or two-line stack trace showing the step-definition and/or feature file where the error originated.
        * Previously, changes made to Pester Gherkin often resulted in the problem matcher for Pester becoming broken, especially for Gherkin-style tests. This change does mean that the existing Pester problem matcher will not work for Gherkin-style tests. However, the benefit of this change is that for Gherkin tests, there is now a consistent way of displaying errors in the console, and a problem matcher can be made for each of the Failed, Pending, or Undefined step statuses in the Visual Studio Code Problems pane. See below for the JSON for the new VS Code problem matchers for Pester Gherkin.
* First-class support for Scenario Outlines
    * Scenario Outlines are now printed to the console, including their steps with their token arguments.
    * Scenario Outline Example Scenario names are now generated based off of the scenario outline example tables.
    * Scenario Outline Example Scenarios can be displayed as they are in the feature file, in tabular form, or, using the new `-Expand` parameter to `Invoke-Gherkin`, in expanded form, where they are displayed as any other scenario (though, output indented from and below their parent Scenario Outline).
    * Scenario Outline Example Scenarios are also displayed as nested test-suites as children of the scenario outline in the NUnit XML.
    * When scenario outline example scenarios fail, and `-Expand` is not used, that example scenario will have its stack trace displayed in the console. (NOTE: this only occurs for failing example scenarios, and not pending or undefined ones. This also means that, when not using `-Expand`, pending and undefined "warnings" will not be displayed in the VS Code problems pane. I'm open to changing this. Ruby Cucumber doesn't display these--BUT, there probably is value to be had by displaying these, especially for VS Code development.)

As mentioned above, if you develop in Visual Studio Code and want to use the Problems Pane for errors and warnings in your test code, this PR introduces new consistent Failing, Pending and Undefined step error/warning output. You can use the following JSON to define Pester Gherkin problem matchers for these step statuses:

```json
"problemMatcher": [
    {
      "owner": "powershell",
      "fileLocation": ["absolute"],
      "severity": "error",
      "pattern": [
        {
          "regexp": "^\\s*Error:\\s+(.*)$",
          "message": 1
        },
       {
          "regexp": "^\\s+at\\s+[^,]+,\\s*(.*?):line\\s+(\\d+)$",
          "file": 1,
          "line": 2
        }
      ]
    },
    {
      "owner": "powershell",
      "fileLocation": ["absolute"],
      "severity": "warning",
      "pattern": [
        {
          "regexp": "^\\s*(TODO: \\(Pester::Pending\\))$",
          "message": 1
        },
        {
          "regexp": "^\\s*at\\s+[^,]+,\\s*(.*?):line\\s+(\\d+)$",
          "file": 1,
          "line": 2
        }
      ]
    },
    {
      "owner": "powershell",
      "fileLocation": ["absolute"],
      "severity": "warning",
      "pattern": [
        {
          "regexp": "^\\s*>>> (No matching step definition found.)$",
          "message": 1
        },
        {
          "regexp": "^\\s*at\\s+(.*?):line\\s+(\\d+)$",
          "file": 1,
          "line": 2
        }
      ]
    }
  ]
```


renehernandez:
@fourpastmidnight @nohwnd Is the file *.vscode/launch.json* expected to be added to the git repo?

Functions/Gherkin.Output.ps1
```diff
@@ -0,0 +1,705 @@
...
+    Process {
+        $StepText = $StepResult.TypeNames
+        $StepParameters = $StepResult.Parameters
+        $Result = $StepResult.Result
+
+        # Split the step text into pieces, where the pieces are either generic step text, or input to the step
+        # definition. This will allow "highlighting" of the replaceable text parameters of the step
+        # definition.
+        $StepTextParts = [hashtable[]]@()
+        if ($Result -notin 'Inconclusive' -and $StepParameters.Keys.Count) {
```
renehernandez: @fourpastmidnight  The `-notin` operator is only available in PS 3+. That is why the PS 2 job is failing to start

renehernandez: @fourpastmidnight Is $Result an array in this context? 

If so, you should switch to do `$Result -notcontains 'Inconclusive'`. Otherwise, if it is a single value, just do a comparison with `-ne`

fourpastmidnight:
Well, I’m finding it very helpful while working on Pester, allowing me to debug at the click of a button as I improve the Gherkin side of Pester. So, that’s currently my intention.

If it is seemed this is not wanted in the repo, the file could be dropped from the commit that added it. But I firmly believe this is an improvement to help others who may wish to help with Pester be able to be up and running with debug capabilities right after cloning the repo.

Sent from my Windows 10 phone

fourpastmidnight:
Ooh, good catch on the -notin usage. Thanks!

Sent from my Windows 10 phone

Functions/Gherkin.Output.ps1
fourpastmidnight: Oooh, good catch on the usage of `-notin`. Also, I had previously implemented these changes on a privately maintained copy of Pester (to initially just get it to work!--and this implementation is much better than my original). Anyway, I was going to try to replace usages of `Inconclusive` with `Undefined` since that is what Cucumber uses. BUT, the NUnit Xml reports I think require `Inconclusive` as a value for the test result, so I'm going to stick with a single value and I need to remove that `-notcontains` and make it `-ne` as you suggest. Again, thanks for the second set of eyes!

fourpastmidnight: Thanks again for pointing those issues out. They weren't the main thing causing the tests to break, but most certainly would not have helped my cause with PSv2. This is fixed in the latest version of this commit.

nohwnd:
IDK, it does not look like it harm anything. I am not so sure about the extended formatting options though. But right now I have the formatting off and the psscript analyzer off as well because it took 30s to save a single file... :/

Functions/Gherkin.Core.ps1
```diff
@@ -52,6 +52,86 @@ function Invoke-GherkinHook {
...
+
+    $AddMember = $SafeCommands['Add-Member']
+    $NewObject = $SafeCommands['New-Object']
+
+    New-PesterState -TagFilter $Tag -ExcludeTagFilter $ExcludeTag -TestNameFilter $ScenarioName -SessionState $SessionState -Strict:$Strict  -Show $Show -PesterOption $PesterOption |
+        & $AddMember -MemberType NoteProperty -Name Features -Value (& $NewObject System.Collections.Generic.List[PSObject] ) -PassThru |
+        & $AddMember -MemberType ScriptProperty -Name FailedScenarios -PassThru -Value {
+            [string[]]@(
+                $this.TestResult |
+                    & $SafeCommands['Group-Object'] Context |
```
fourpastmidnight: Interesting, so it seems `Group-Object` in PowerShell Core (6.1) ends up _sorting_ the groups by the property(ies) selected for grouping! This causes the unit tests to fail on the AppVeyor build. So I'm left wondering what my options are.

I could just force the issue and sort the results. But I'm concerned that doing this would have unintended consequences?? Is the order important in the result object? For example, the ordering on PSv2 - PSv5 is the order in which the tests were executed. Is it OK to change this ordering?

I could "synthesize" a name to preserve the ordering when using `Group-Object` on PSv6. The downside to this is the extra processing involved.

fourpastmidnight: I think I've decided I'm going to synthesize a key name upon which to perform grouping which would preserve the ordering when using `Group-Object` in PowerShell Core. I guess I should also file an issue with the PowerShell team.

fourpastmidnight: OK, so apparently, this is a breaking change in PowerShell v6. In order to dramatically improve performance from quadratic to `n log n`, they first sort the objects and then group. I did post a comment on the issue ([#7409](https://github.com/PowerShell/PowerShell/issues/7409?ts=2)), but I doubt that'll be reversed since 1) it results in dramatically better performance and 2) the work was already done, discussed, approved, and obviously released. So my decision to synthesize a key to guarantee the ordering is the right one.

On a side note, the extra processing should be minimal. The biggest cost is looping over all the results and the creation of objects in order to keep the original ordering. It's just a shame that this extra work needs to be done at all, as it clutters up the code.

renehernandez: @fourpastmidnight I am not that familiar with the Gherkin code, but is preserving the ordering important? Is preserving the ordering a guarantee of the Pester implementation? 

@nohwnd @Jaykul Any input to add?

nohwnd: @renehernandez I have nothing.

fourpastmidnight: I may have managed to find my way around this, and with a _MUCH_ simpler solution. I've noticed for some time now, that even though we return (basically) a Tuple  from `Import-GherkinFeature`, that the actual `$Feature` object itself has the scenarios added as a property.

With all my changes that I'm making, I'm adding a synthesized property for the scenario result as a whole. Now, as long as PSv2 doesn't drop that at some point before we get to (or as I'm) creating the TestResults, then I can simply use that to find scenarios based on their result, instead of all this hoopla. And the ordering will remain in tact, so the question becomes moot. The real question here is will this work with PSv2.

fourpastmidnight: _RIGHT_, since I added the synthetic Result property, it was subsequently dropped by PSv2. And so in `Invoke-GherkinStep`, I had to check for the property, and if it didn't exist, add it back. BUT, when I do that, of course, I never "replace" the Scenario object on the Feature with the "new" PSObject created by adding the synthetic property--because I don't have access to the feature. IOW, this won't work. Too bad. In v5, this'll totally work, though (although, things will be much different in v5 if I can ever get time to work on it).

fourpastmidnight: Hmm, I think we can work around the PSv2 issue by adding `-PassThru` to the `Add-Member` invocation in `Convert-Tags` and subsequently returning the "new" PSObject created by that command. Then I may be able to get rid of the check in `Invoke-GherkinStep`, as well as make this commit work, too.

fourpastmidnight: OK, So, I managed to work around everything by making sure that Pester doesn't lose the `PSObject` wrapper over `Gherkin.Ast.Scenario` and `Gherkin.Ast.Background` objects for extended properties that are added (e.g. the `Result` property). Doing this allows me to get rid of the check for this property in `Invoke-GherkinStep`. This also means that this property is available on the `Scenario` objects in the Pester Results object, which allows me to use _MUCH_ simpler code that works for PSv2 all the way through PSv6 without any shenanigans.

The most recent Azure Pipelines build failed because one of the steps in the PSv2 build (the finalize step) took too long to run for some reason--but all the tests passed. So now I just need to clean up the most recent 4 commits and squash them.

fourpastmidnight: This is resolved with the latest commit.

fourpastmidnight:
Apparently, I lost the PSObject wrapper somewhere along the way again... 😐

Functions/Gherkin.Core.ps1
```diff
@@ -0,0 +1,1304 @@
...
+        .PARAMETER Tags
+            Tags for filtering hooks
+    #>
+    [CmdletBinding()]
+    param([string]$Hook, [string]$Name, [string[]]$Tags)
+
+    if ($GherkinHooks.${Hook}) {
+        foreach ($GherkinHook in $GherkinHooks.${Hook}) {
+            if ($GherkinHook.Tags -and $Tags) {
+                :tags foreach ($hookTag in $GherkinHook.Tags) {
```
renehernandez: What is the purpose of this `:tags` declaration here?

renehernandez: As well as in line 40 next to the break statement

Functions/Gherkin.Core.ps1
```diff
@@ -0,0 +1,1304 @@
...
+            Invoke-Gherkin -CodeCoverage @{ Path = 'ScriptUnderTest.ps1'; StartLine = 10; EndLine = 20 }
+
+            Runs all *.feature specifications in the current directory, and generates a coverage report for all commands on lines 10 through 20 in the "ScriptUnderTest.ps1" file.
+
+        .LINK
+            Invoke-Pester
+            https://kevinmarquette.github.io/2017-03-17-Powershell-Gherkin-specification-validation/
+            https://kevinmarquette.github.io/2017-04-30-Powershell-Gherkin-advanced-features/
+    #>
+    [CmdletBinding(DefaultParameterSetName = 'Default')]
```
renehernandez: Where is the *Default* parameterset name defined? None of the parameters below specify it in the `Parameter` attribute definition.

Were you referring instead to **__AllParameterSets**?

Functions/Gherkin.Core.ps1
fourpastmidnight: So, I didn't add this label, it was already there. It's more than likely this code appears 'added' because it moved from its previous location in the file.

However, I can answer this question.

The `$Tags` variable that's being enumerated over in the inner loop contains the tags as specified for the `-Tags` parameter to `Invoke-Gherkin`. Now, for scenarios and features and whatnot, if _any_ tag on the scenario, feature, etc. matches one of the tags specified in the `-Tag` parameter, then that scenario, feature, etc., is selected for execution. (Processing of other filters may eventually further exclude these scenarios/features, but, as far as `-Tags` works, my previous statement is accurate.) So, once an "include" tag matches just one of the tags on the hook, we can stop looking for other matching tags because the hook will be included. So we "execute" the hook script block to register it with Pester Gherkin. And because it's been included and registered, we can skip looking for any further tags--there's no point.

Anyway, the one thing I just noticed with this code is that the `ExcludeTag` tags are not processed! I'm thinking this might be a bug. Any thoughts @Jaykul  or @nohwnd? If this is a bug, then I should look to fix this while I'm in here making changes.

Functions/Gherkin.Core.ps1
fourpastmidnight: This code is making use of some standard behavior of PowerShell. Because the parameter `FailedLast` is explicitly added to the `RetestFailed` parameter set, that's the only parameter set the parameter is added to. So far, so good, this is standard PowerShell. Now, also standard with PowerShell, any parameter without a specific parameter set assignment is added to _all_ parameter sets defined on the Cmdlet. So, now the `RetestFailed` parameter set includes _all_ parameters for `Invoke-Gherkin`. Now in the `CmdletBinding` attribute, `Invoke-Gherkin` declares that the default parameter set is named `Default`. Even though no parameters are explicitly assigned to that parameter set, as previously mentioned, all parameters not explicitly assigned to a parameter set are assigned to every parameter set defined by the Cmdlet. The very fact that we said the default parameter set name is `Default` implies that there is a `Default` parameter set and since all other parameters other than `FailedLast` have no explicit parameter set mapping, these parameters are added to both the `Default` and `Retestfailed` parameter sets. I actually didn't think that doing this would work (and never really bothered to examine this before since `Invoke-Gherkin` just works).  This is a fairly neat short-hand trick to creating a parameter set.

Anyway, check it out for yourself, load Pester in a PowerShell ISE session and check out the help in the side-pane. Pretty neat stuff.

fourpastmidnight: BTW, please also note that I did not add this default parameter set name. Again, the code in the file moved, making it look as if this code was added by me.

renehernandez: @fourpastmidnight Sounds good to me :)

fourpastmidnight:
Just pushed up the last of the changes I planned to make, that is until I found #1285. So before I wrap up this PR, I would like to fix that issue since I'm in here and have made changes to the unit tests in that area.

nohwnd:
Okay. I see that most of the changes are in gherkin related files only, so merging this should be no problem.

fourpastmidnight:
Well, all the tests are passing, but I decided to take what I have so far and actually use it with tests I wrote for 3 modules I recently developed. And there are things that are wrong with this that I need to fix:

* Multiline arguments were not indented correctly
    * Fixed, but not yet pushed. I was using the old PesterState.IndentationLevel indent instead of the new one for Gherkin: GherkinIndentationLevel. Updated all the step printing code.
* For some reason, when running the `Gherkin-Background.feature` in the Pester test suite, everything looks ok. But, when running against actual code, the background steps don't appear to be worknig correctly. They're printed at the start of the feature, but then the steps are run "out of order" with the scenario (e.g. they execute after the scenario), leading to test failures. AND, they're printed again (after the scenario!?!). What's up with that? And why aren't the Pester tests seeing this behavior?
* I have one failing unit test in one of my modules (due to the Background issue I just mentioned). The failed scenarios count is 95 even though only one scenario failed. The steps count is correct. Weird.

One thing that I'm not a huge fan of is that because `Write-Host` does not allow you to redirect the output, there's no way to verify the console output other than "looking at it" and hoping you've done enough manual testing with enough variation to catch everything. I have a thought that there should be some way to "redirect" output to somewhere for testing/reporting purposes, while still "forwarding" the output to Write-Host to colorize the output. For example, I would be interested in capturing the test run console output to a file (instead of using NUnit Xml and relying on non-existent/not really nice NUNitXml -> HTML converters). Of course, this issue goes away with PowerShell 5.1 (I think--if not 5.1, then 6+). But in the meantime, as long as Pester needs to support PSv2 (and, even in Pester v5 we're supporting PSv3+), we're stuck. I'd love to use something akin to `Tee-Object`, or create some sort of proxy for `Write-Host` that would allow this in a way that's more flexible than the straight usage of `Tee-Object`.

fourpastmidnight:
And one more issue:

I was trying to keep the console output clean, regular and consistent. This meant removing extraneous blank lines from description text. The only problem is, if the line was blank, I don't print it. This means that when you have longer descriptions, you end up having a "wall of text". Not ideal, IMO. So, I need to figure out what to do about this.

fourpastmidnight:
Oh, and another bug that's fixed but not yet pushed: in the new function `New-GherkinPesterState`, I made the `PesterOption` parameter a `[hashtable]`, but it is in fact supposed to be a `[PSCustomObject]`.

fourpastmidnight:
OK, I did some validation of behavior in Ruby Cucumber. While Ruby does properly indent the Feature names, Scenario names, etc., it does not "reformat" descriptions--which makes perfect sense. So, I'll be pushing an update that removes the blank line suppression and line-trimming that I was doing in the various Write-* functions when descriptions were written to the console.

fourpastmidnight:
OK, the only remaining thing I'd like to put into this PR is the fix for #1285.

fourpastmidnight:
Oh, and I need to verify there's not a bug in timing the step execution duration.

fourpastmidnight:
So, there's a general bug in Pester with respect to recording test time. I'm not sure how it works, but the _method_ I was using was not actually capturing the time. For example, I noticed that in `It.ps1`, their are bookend calls to `$Pester.EnterTest()` and `$Pester.LeaveTest()`. I don't see these bookend calls being made in the current `Gherkin.ps1` file, and so they aren't (weren't) in my `Gherkin.Core.ps1` file either. Now, I have since added those calls to `Invoke-Scenario` in `Gherkin.Core.ps1`, but even so, the `$PesterState.Time` field was never updated from it's initial value of 0. I do see that times are being recorded when entering/leaving TestGroups. But, as I understand it, those are ephemeral and go away (well, they're used to construct the NUnit XML, so the times captured in the TestGroupStack show up in the XML report). I don't even see in Pester RSpec where the `$PesterStat.Time` field is being updated--though, I know this is working.

Anyway, this whole thing led me to resort to adding up all the individual step-definition execution times.

Because of that, I found very interesting results. Running this PR against one of the modules I recently wrote, Psake tells me that the task used to execute the tests took a total time of 35.5 seconds (there's _some_ overhead from Psake in there, but not much--not more than 2s - 4s probably). What's _interesting_ is that the total step-definition execution time averages 4.5s! What this means is that Pester itself has an overhead of 20s - 25s (roughly). That's huge. Now, I know, Pester's doing a lot to get everything setup, and all test runners will have overhead. But, this does seem a bit large.

I know where some of this overhead comes from--such as reloading all the step definition script files on every feature invocation. I plan to resolve this in Pester v5 Gherkin. Hopefully with that one change alone, we'll see the total overhead decrease by 50%, down to 10s - 12s.

There's also a significant amount of time spent printing to the console--though, this PR _greatly_ reduces that time--IOW, the total overhead was _much_ greater before this PR. I figure, at least a 30% - 50% perceived speed-up. Let me put it this away, it's enough of a speedup that _I_ notice it runs _much_ faster than before.

I haven't pushed up the timing fix just yet. There were a couple of other issues I came across as I begin to use this PR against some modules I've written recently to really try this PR out because not everything is testable (e.g. console output). 😐

nohwnd:
20s seem like a lot. How many tests are there? Are you measuring it via measure-command?

fourpastmidnight:
It does seem like a lot. So as I prefaced, the 35s total time was for the task I ran in Psake. I _assumed_ 2s - 4s overhead in the task--but I need to do more specific, targeted instrumentation to make better determinations. And I couldn't tell you if Psake is using `Measure-Command`, which I understand has some shortcomings. As for the number of tests, there are 39 scenarios executing 290 steps--so not a lot. So I would think that for medium- to large test suites, the overhead (whatever it is) is significantly smaller.

It would still be interesting to see how much time it takes for the actual test run to occur; so I could easily add a stopwatch to my Psake test task and wrap it around my call to Invoke-Gherkin to know what the real test run execution time was. I hope it's less than 25s considering the steps execute in less than 5s. I'll post back when I find out.

fourpastmidnight:
I did notice, looking at my PSake test task, one thing that takes a good amount of time (contributing to the 35s Task execution time): the generation of an HTML report from the NUnit XML.

So with that, and adding the stop watch around the call to Invoke-Gherkin, I'm very happy to report that while the steps executed in 5.472s, the total test execution time was 13.114s. So, that's not bad at all! Good. And that's for Gherkin, where we're reloading the step definitions on each feature that's being invoked. Eliminating that extra work will bring down the overhead even more for small test suites such as mine.

And, I want to apologize--I wasn't trying to be inflammatory above about the total execution time. But I am trying to improve Pester (on the Gherkin side), and if I can shave off a few seconds, that's great. So, it was with the best of intentions.

nohwnd:
No that’s fine, no need to apologize, I was just genuinely interested in how you are measuring this, because there is some overhead, but it’s usually <10ms per test. :)

fourpastmidnight:
Hey all. Sorry, I got pulled into something at work that demanded a lot of my free time. I'm hoping to get back to this in short order.

fourpastmidnight:
Hmm, I see a fix a few commits back (3e20be2) but it seems to be failing in the same manner again? Oh well, I have some more fixes I need to make before this can be merged, so let's see if the build will "magically" turn green--indicating flaky tests due to the issue identified in the aforementioned commit.

fourpastmidnight:
Yay, I'm finally getting time to work on this again!

So, pursuant to some of the conversation above earlier this year, here's what's left on this PR:
* [ ] Determine what's wrong with recording of test execution time
    * I have determined that I am not correctly capturing test execution time.
    * To this end, I have added a bunch of profiling code that times (almost) everything
        * The next question is, should I keep this code? It adds a new `-Profile` parameter to the cmdlets that triggers turning on profiling and returns the results on the TestResult object.
* [ ] Fix the NUnitXml report output for Gherkin tests.

fourpastmidnight:
Hmm, so I _am_ correctly capturing step execution time. BUT, there's a lot of overhead outside of step execution time. I'm guessing that it's all the printing to the console. I need to add some more profiling around the writes to the console to know for sure.

I'm doing this mostly out of an exercise of curiosity at this point--since the total test run time for a suite of tests I'm familiar with is about 00:01:30 but the actual step execution time is only 18s!!

If printing to the console is where the biggest amount of time is spent (and that's the most likely culprit at this point), then I want to know to ensure that it's not something else in the framework. There's not much that can be done about the console printing time (other than not printing to the console by not showing that output! 😄 ).

Jaykul:
Yeah, have you tried running with output turned off and -passthru instead?

fourpastmidnight:
Well, it’s not that I think Pester is slow or anything, or that I want my tests to run faster (though, faster is always better!). It’s just that I saw a rather large discrepancy. Later than I expected. I’m pretty sure console printing is what adds a ton of overhead, and that’s just the nature of the beast. I have no doubt that running with little to no output the test run will be significantly faster. I just wanted to be sure that there’s nothing else in the framework that could be unnecessarily contributing to slower execution times. FWIW, I really do think it is the console, but it doesn’t matter what I think; it only matters what the numbers show. 😊

However, I’m guessing that now that I know the step executions are extremely fast, but the overall test run time is much slower (when printing to the console), the runner should show the overall test run time since that’s the time the end use experiences. After all, this very discrepancy is what led me to profile Pester, to understand why 18s was being reported by my new timing code, but yet I had to wait ~1:30min.

The only reason I came across this is because I started capturing elapsed time at the step/pester test result level and thought that adding all the step execution times together would result in a pretty close approximation of the overall test run time sans some framework overhead which I (rightly) thought would be minimal. However, my assumption apparently doesn’t hold up when you take into account console printing.

Anyway, thanks for the suggestion!!

Sent from my Windows 10 phone

fourpastmidnight:
Alright, get the console writing code instrumented, too. And my suspicions were correct. The console printing code is where the bulk of the time is spent on the test run.

The good news is that this PR does reduce the amount of console printing, so it'll be a bit faster than before. But there's still a lot of printing, so for a faster CI build, it would be better to use the `-Show None` option and then use a report generator to generate an HTML report you can then post along with the build results in, say, Azure DevOps Server (or your favorite CI server of choice).

Which brings me to the next and last part of this PR: fixing NUnit XML generation from the Gherkin runner.

Jaykul:
For what it's worth (in case you're being tempted):

While there's some sense in measuring and reporting "wall clock" duration of "all the work," regardless of the confusion it sometimes causes, for developer's sanity the timings of the tests need to be the execution time of the test code -- because it's the only thing they have control over. 

That is, if a Pester user wants to speed their tests up, the most pertinent information is which tests (and test suites) take a long time to run, and the discrepancy between the sum of the test times and the wall clock time.

Jaykul:
P.S. probably should have broken this up into multiple PRs. Next time ;-)

fourpastmidnight:
@Jaykul 

Well, I can still make the NUnit XML report generation fix a separate PR, and is probably a good idea. :) I was leaning towards that anyhow.

Hmm, I was thinking the same thing regarding test timings and what should be displayed. And so I was thinking about showing two different timings at the end: test execution time, and the entire test run time. But I was afraid that showing that would cause more questions than anything else (given, for example, my suite of 960 steps runs in 18s but with full console output, the test run takes 1:30 - 2:00).

I'm open to whatever behavior the community deems appropriate, however.

fourpastmidnight:
Ok, looking at Ruby Cucumber as my guide, they are basically doing what I'm already doing: gathering all the execution times of the test cases (i.e. scenarios), whose durations are really aggregations of the execution time of the individual steps making up the test cases.

So, I'm going to leave my existing timing code as-is.

Since I have decided to split off the NUnit XML Generation bug as a separate PR, I just need to fix up any broken unit tests and promote this PR from a Draft to a "ready-to-review" PR.

fourpastmidnight:
I keep having one test fail on the appveyor build related to `Clear-TestDrive` removing symbolic links. When I run the tests locally, this test is skipped. I'm thinking this appveyor box is not A Windows box whereas my local dev machine is.

Considering I didn't touch any code in that area of Pester, I fail to understand why this test would be failing.

fourpastmidnight:
Other than trying to find the magical combination that will allow all the unit tests to pass in all execution contexts, this PR is ready.

fourpastmidnight:
@nohwnd OK, there's something with the PowerShell Core AppVeyor build and the Pester test suite. The same test keeps failing over and over on this build: `It "Clear-TestDrive removes problematic symlinks"`. This test does not fail on my machine running PowerShell 6.2.3.  It also passes the TravisCI build for both the Mac and Linux machines.

Functions/Gherkin.Output.ps1
```diff
@@ -0,0 +1,907 @@
...
+        $ErrorLines.Message += $ExceptionLines
+
+        if ('PesterAssertionFailed', 'PesterGherkinStepFailed' -contains $ErrorRecord.FullyQualifiedErrorId) {
+            $ErrorLines.Message += "$($ErrorRecord.TargetObject.Line)`: $($ErrorRecord.TargetObject.LineText.Trim())".Split([string[]]($([System.Environment]::NewLine), '\n', "`n"), [System.StringSplitOptions]::RemoveEmptyEntries)
+            $ErrorLines.Trace += @($StepResult.StackTrace -split '\r?\n')
+        }
+
+        $ErrorLines
+
+        # ! Leaving this here because I may need this if it becomes evident that showing the portion of the
```
fourpastmidnight: I'm interested in knowing if I need to put any of this code back in in any way? Near as far as I can tell, the rest of the stack trace usually contained Pester internals information, which isn't really helpful to someone in trying to understand why their test is failing. (Though, I must admit, the reason why I started working on Pester is because Pester internals (and my lack of PS knowledge) were causing my tests to fail! Lol...ironic.)

fourpastmidnight: OK, I'm going to reinstate these commented-out lines. But I have a feeling that the stack trace is not going to show any further details due to how I'm processing the stack trace in `ConvertTo-GherkinStepResult'. I'll need to rediscover 1) why I commented out those lines to begin with (it had something to do with getting consistent output for use with VS Code Task problem matchers without them breaking every time someone changed something in this area of the codebase); and 2) what the behavior ought to be when the code breaks in the code/system under test, in a script block and in a feature. (The last two are already working fine, but what does/should the stack trace look like for broken code (e.g. NullReferenceException, InvalidOperationException, etc.)?)

fourpastmidnight: Alright. I think I found the error of my ways. (Pun intended.) I'll need to do a bit of legwork on this section here, but given this PR is fairly large, it's still good to get this review going, and this is a relatively small section (but nonetheless important).

fourpastmidnight: OK, I have been able to reinstate these commented-out lines. However, their location has changed. And in the process, after reorganizing the code, I've been able to remove ~400 lines of code. This will be resolved with the next commit. So I'm going to go ahead and resolve this now.

Functions/Gherkin.Core.ps1
```diff
@@ -0,0 +1,1454 @@
...
+    if ($Pester.ExcludeTagFilter -and @(& $CompareObject $Feature.Tags $Pester.ExcludeTagFilter -IncludeEqual -ExcludeDifferent).Length -gt 0) {
+        return
+    }
+
+    $Scenarios = @(
+        :scenarios foreach ($Child in $Feature.Children) {
+            $null = & $AddMember -MemberType NoteProperty -InputObject $Child.Location -Name Path -Value $Path
+
+            foreach ($Step in $Child.Steps) {
+                $null = & $AddMember -MemberType NoteProperty -InputObject $Step.Location -Name Path -Value $Path
```
fourpastmidnight: Apparently, this line, and line 493 above, never worked in PowerShell v2.0. The Path property is never added to the ETS on the Location object referenced by the Location property of the respective objects. This is because the Location property on those objects is a read-only property (it's declared as `Gherkin.Ast.Location Location { get; private set; }`). There must have been some compensating code (very far removed) somewhere to make the stack trace nice, for example. Anyway, these lines are causing 4 unit test failures. I'm evaluating how to make this work in PowerShell v2.

fourpastmidnight: Just to provide a bit more detail, I ran this code in PowerShell v5.x. In the Debugger windows, the Steps (for example) did not show a `Path` property on the Location property either. _However_, when performing a `$Child.Steps[0].Location | Get-Member -View Extended` call, the `Path` property was returned. So it _is_ available--which makes sense since the tests do pass on PowerShell v3 and above. And this is expected since the way the ETS works changed between PSv2 and PSv3.

fourpastmidnight: OK, there was some (far removed) compensating code that handled this PowerShell v2 behavior. Instead of having 2 ways to deal with this, I used a PSv2 compatible approach. This will be fixed in the next set of commits.

fourpastmidnight:
OK, @nohwnd , the only test failing is the Clean-TestDrive test on the version of PowerShell Core installed on the appveyor CI box&mdash;again, which passes on my machine running PowerShell Core v6.2.3. Everything else is passing.

@Jaykul and @renehernandez, if either or both of you could spare a few moments to check this over, perhaps even pull down the PR and give the Gherkin runner a try, I'd be very appreciative.

nohwnd:
@fourpastmidnight I see, we had a problem with that symlink tests before. Not sure what the solution was, but feel free to ignore it for now. I will solve that during merge.

fourpastmidnight:
Just an FYI that I rebased this branch on top of `master` and the TravisCI build failed because `apt-get` failed very early in the build process, so, not a real failure of Pester in and of itself.

fourpastmidnight:
Interestingly, the failed test shown in the AppVeyor build is a difference between PowerShell Desktop Edition and PowerShell Core. I fixed the test locally for PowerShell Core, but then it broke for PowerShell desktop.

fourpastmidnight:
OK, I have a fix for the above. I'll put it in a separate branch.

fourpastmidnight:
Looking through recently merged pull requests, I see #1428 removed the Pester header. When I rebased this PR, I accidentally re-introduced it. I'll be sure to remove it in a future push to this branch.

fourpastmidnight:
Rebased this on top of PR #1475 so the tests will all pass.

Also the last commit on this PR fixed an issue with output that I didn't come across until recently.

Other than that, I've been dogfooding this PR for over a year and have had great success using this version of the Gherkin runner.

I'm itching to get this merged!!

@renehernandez, @Jaykul and @nohwnd , I'd appreciate if any of you could review this PR when you have a few minutes. Maybe try it on a project you have that's using  Gherkin-style tests.

fourpastmidnight:
OK, since #1475 was merged to master (thanks @nohwnd!), I once again rebased this on master. This PR is finally ready to be merged. If this is merged, it should result in a minor version bump of Pester.

nohwnd:
Nice! I guess no-one else was able to validate your changes? 


@Jaykul this is about to be merged, could you run it against your testsuites? So I don't break you when this is released?

Jaykul:
I can. One moment ...

Jaykul:
Looks good.

I don't quite understand what happened with the colors though. I mean, 
1. Did we remove the color from `Feature` and `Scenario` output on purpose?
2. What is the meaning of the highlights? Is that just based on regex in the step matching?

This version:
WindowsTerminal_2020-04-06_22-15-02

Before:
WindowsTerminal_2020-04-06_22-15-29

nohwnd:
@jaykul do you still use gherkin a lot? Right now Pester v5 tries to be accomodating to be layered with another freamwork "face", and while it was useful to think about the internals as being modular to keep the internal architecture clean, it brings a lot of overhead, and unnecessary renaming, just to support <1% of users. 

I am thinking about removing it entirely from v5, and splitting it from v4 to it's own repo under pester organization where hopefully @fourpastmidnight and you could rule and improve it as you see fit? Either forking v5, or maybe I could still keep the "runtime" part distinct from RSPec Pester and rename the items during build and not during runtime. And you'd do the same to build Gherkin Pester. It would then ship in a nuget under a different name, say PesterGherkin.

nohwnd:
If you know other people that should be included in this conversation please tag them, I can move it to a new thread later if needed.

fourpastmidnight:
Ok, wow, there's a lot to respond to here. In this comment, I'm going to address the things raised about the coloring.

@Jaykul:

Yes, the highlighting is showing what text in your steps matched the regexes in your step definition. This comes from Ruby Cucumber. It's _super_ useful. In fact, I caught a bug in my own test suite the other day because of this highlighting (the regex was capturing more than I intended). It can save countless minutes of debugging by being able to visually see what step text matched a step definition regex.

To prevent a regex match from being highlighted, you can use a non-capturing regex: `(?n:match-some-stuff-here)`. In fact, I have come to believe (and, this is _my_ opinion, I realize) that it may be a best practice to write step definitions like this in the majority of cases (>95%):

```powershell
Given '(?n)^This is some (?<Parameter>parameterized) text to be matched$' {
    Param([string]$Parameter)

    Set-StepPending
}
```

I have been considering whether or not case-insensitivity would also be a best-practice (as a starting point) since it can be turned back on for individual capture groups, if needed. In any event, I find this leads to consistent matching step definitions.

As for the Feature and Scenario colors, yes, the color was changed to match the default Ruby Cucumber coloring.

There are many implementations/ports of Ruby Cucumber (see [Related Tools here](https://cucumber.io/docs/tools/related-tools/)). Pester is not listed among them. I believe with this PR in place, we could actually get listed here. Anyway, back to coloring. All the coloring has been consolidated to `en-US\Gherkin.psd1`. One thing I would love to enable in the future is for someone to be able to use their own coloring options by providing a similar data structure to Invoke-Gherkin (either directly through a parameter or via a Gherkin extension to the PesterOptions object).

Also, I _think_ you mentioned this before&mdash;that you don't necessarily want to see the DocStrings/DataTables printed to the console? This can be achieved by passing the `-NoMultiline` switch to `Invoke-Gherkin`.

fourpastmidnight:
@JayKul  BTW, I'm not opposed. I could put the Cyan color back in for the Feature/Scenario title.

fourpastmidnight:
OH--_and_ I missed the part where you said you _liked_ the highlighting. Sorry. 😄 I originally read this at 1 AM and didn't see that bold, highlighted **like**.....gee--wonder how  I missed that...

fourpastmidnight:
@Jaykul 

I can't seem to find where you commented on the label. BUT, I went back to look at the history of those lines. I renamed `Gherkin.ps1` to `Gherkin.Core.ps1`, so the implementation of `Invoke-GherkinHook`, where you found the issue has not changed since the file rename. So this is a latent bug in the Gherkin runner that hasn't been found until now!

I'll add a commit to fix that. I'm thinking all we need to change is line 44, from `break :tags` to `break tags`. I don't use hooks often (and I'm guessing many people may not), so I've never run into an issue due to this. Thanks for catching that!

Also, I remember the reason why light cyan is not part of the scenario/feature titles: here's an example of my test run where I use scenario outlines and data tables:

image

fourpastmidnight:
@nohwnd Re: future of Gherkin in Pester.

I have mixed feelings about this. I've actually wanted to approach you and ask for your thoughts on this very subject. Let's start a new issue to pick up this conversation. I feel like it's one worth having if only to "clear the air" and properly set the stage for not only the future of Gherkin, but possibly even Pester itself. 😄  I think some of what you mentioned makes a lot of sense. But like I said, I do have mixed feelings. So let's do a Cost-Benefit analysis in another thread. (I don't want to speak for @renehernandez, but he did comment on this PR; so perhaps he should be an initial participant, too?)

fourpastmidnight:
As an FYI, I found a cosmetic bug: the highlighting isn't always highlighting what _should_ be highlighted under all circumstances. And it's not improperly highlighting things that aren't matched&mdash;it simply isn't necessarily highlighting all things that _have_ matched.

Again, it's more cosmetic than anything else. Regex is Regex, so whatever is matching is matching and properly setting parameter values. I don't believe that what I found should hold up merging of this PR since it is only cosmetic.

_IF_ this PR is merged before I implement a fix, then I'll just open a new Issue and PR for the fix. If not, then I'll add it to this PR&mdash;but I don't want to hold up this PR any longer for a cosmetic blemish. 😉

fourpastmidnight:
Well I'll be.... Something in my tests caused the highlighting to not work properly, because all of a sudden, it is. Weird. And slightly disturbing. I'll keep an eye on this and see if there isn't something to be done to make it more robust. At least it's only cosmetic and has nothing to do with actual test execution itself.

fourpastmidnight:
@JayKul  Just checking back in. So, if we used the bright cyan color for Scenario and Feature titles, that _shouldn't_ clash too much with the surrounding output. I'm not opposed to making that change if it would help make those stand out better. I suppose the authors of Ruby Cucumber thought that indentation helped to dilineate those items&mdash;but I can see it both ways.

Apart from that, do you feel this is ready to go?

Jaykul:
I'll admit I haven't given it a thorough re-code-review recently, but it worked for the test cases I have, so I'm happy with it 😁

fourpastmidnight:
OK, I sort of figured out why highlighting sometimes "doesn't work" The short answer is, it _IS_ working. Only it doesn't look like it.

It's been over a year since I wrote this code, so I don't quite remember how it works&mdash;and the 1-year later me doesn't quite grok it anymore (which means, I probably need to either refactor the code so I can understand it better, or barring that, put in some comments 😮 ). Anyway, the _slightly longer_ answer is, if there's a problem with the test, then highlighting might not be accurate.

This is more cosmetic than anything. BUT, I should probably revisit that code. It _should_ be possible to make highlighting _always_ work, and work correctly, even if the test is failing.

I still think that this slight issue shouldn't hold up merging this PR. This is something that can wait to be fixed at a later time. It's more cosmetic than anything. And if you've got a failing test, you're either going to fix the failing steps and/or debug the test to determine what's wrong to fix it. And once you fix it, everything will "just work".

fourpastmidnight:
@nohwnd Since JayKul indicates he likes the results of this PR and all his test cases are passing, and my own test cases have been using this PR for about a year now, perhaps it's time to get this merged? 😄 

I have bug #1285 to fix, as well as a few other enhancements for the Gherkin runner all hinging on this PR being merged.

nohwnd:
Sorry that I did not merge it, and since there were no pokes about this for over a year, do you still need this merged? I'd like to avoid breaking the current stable v4 release.

fourpastmidnight:
@nohwnd Well, I poked and poked and poked about this over the last 2 years, but you never merged it pending "v5 breaking changes". Given that Pester v5 removes Gherkin, and given that there are no breaking changes in this PR, I'm still in favor of merging. Also, @Jaykul in his comments above seemed to be OK with these changes; and I've been regularly using "this version" of Pester for over 2 years.

Looking back over the commit history, I don't believe there are any _major_ breaking changes. One _minor_ change is that steps with no step definition are marked as Undefined and still cause a test-run failure. Prior to this, steps with no step definition were simply marked as failed. The behavior is mostly the same--i.e. test run is marked as failed, but the reporting details are updated to show a distinct Undefined state. Also there's an added Pending state that can be _explicitly_ set by test implementors, so this is an addition (not a breaking change) that users can _opt-in_ to. Again, all-in-all, I see this is a worthy feature to merge into v4.

nohwnd:
Ok, will try to merge and publish :)

fMichaleczek:
@nohwnd @fourpastmidnight thank you both, I will wait and test the new package 🙂

nohwnd:
@fMichaleczek pushed as [4.10.2-beta1](https://www.powershellgallery.com/packages/Pester/4.10.2-beta1)

fMichaleczek:
I will test it before the end of the week end.

fourpastmidnight:
@nohwnd Thanks for merging this!!  @fMichaleczek I look forward to any feedback. Once this review has passed and becomes a non-prerelease version in the 4.x branch, I look forward to updating my fork to the latest 4.x line and seeing where to go from here with the Gherkin "port" of Pester. Thanks @nohwnd for your help with this. Thanks @Jaykul and @renehernandez for feedback over the time I was working on this.

fMichaleczek:
@fourpastmidnight I'm a newbie in Gherkin but I like the concept. I will try to see how I can use it.

PR #1278, opened 2019-03-27
Add parameters to mock to relax commands

### LINKED ISSUE #601: Input requested:  Mocking commands with annoying, strongly-typed parameters
This has long been the most annoying type of test to write in Pester, and I want to do something about it.  Haven't been able to think of a perfect solution, but there are a couple of options available, and I'd like to get some feedback from the community.  First, an example of the problem:
## Looks great.  Doesn't work.

``` posh
# Somewhere in your code to be tested:

$adapter = Get-NetAdapter
# Do something
$adapter | Set-NetAdapter
```

If you look at the parameter metadata for Set-NetAdapter (or run Trace-Command), you'll see that we're binding the InputObject parameter via the pipeline in this scenario, which needs to be a [ciminstance[]] object with a particular PSTypeName.  Now let's mock these commands in a test somewhere:

``` posh
Mock Get-NetAdapter { [pscustomobject]@{ Name = 'Mocked' } }
Mock Set-NetAdapter

# Do stuff

Assert-MockCalled Set-NetAdapter -ParameterFilter { $InputObject.Name -eq 'Mocked' }
```

Reasonable looking code.  Mock the Get command to return a dummy object and eliminate one external dependency, and mock the Set command in conjunction with Assert-MockCalled to take care of the other.  Except it doesn't work, with an error something like this:

```
Set-NetAdapter : Parameter set cannot be resolved using the specified named parameters.
At line:1 char:6
+ $adapter | Set-NetAdapter
+      ~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (@{Name=Mocked}:PSObject) [Set-NetAdapter], ParameterBindingException
    + FullyQualifiedErrorId : AmbiguousParameterSet,Set-NetAdapter
```

The exact error may vary a bit, but they all boil down to the fact that the simple custom object we returned from our mock of Get-NetAdapter can't be passed to the InputObject parameter of the mock, because Set-NetAdapter's expecting a different object type (even when mocked; it gets the parameter metadata from the original command.)

The way things stand today, you have two main options for getting around this situation:
- Have your mocked Get- command return an object compatible with what the other cmdlets later on will expect.  In the case of a [ciminstance] with a particular PSTypeName, that's not too painful; we can make that happen.  But sometimes you might be dealing with types that don't have public constructors, or have constructors that do things you don't want them to do as soon as the object is created.
- Write a function with similar, but not quite exactly the same metadata as the Set-NetAdapter command, then mock that instead.  In this case, you'd probably change InputObject's type to [object[]] and get rid of the PSTypeName requirement.  Works, but can sometimes be tricky due to scopes and function lookup order, and requires a fair bit of code.
### Option 1: Automatically make all parameters [object] or [object[]], get rid of validation attributes on the mock.

On the plus side, you'd completely avoid the types of errors discussed in this post, but may introduce other errors.  For example, you might bind to a different parameter set than what would have happened in the original command, or your test may succeed when a call to the live command will fail.  If you're using parameter filters and Pester needs to call the original command due to not matching any filter, it may be possible for the command to error for similar reasons.  (Not certain of this one.)
### Option 2: User-controlled version of Option 1

Instead of making _all_ parameters "safe" for mocking, only change the ones the test author specifies when they call the Mock command.  Something like `Mock Set-NetAdapter -SafeParameter InputObject` .  This way you only use the feature when you need it, and otherwise won't get the weird binding / false positive problems mentioned above.

Possible downside to this:  mock parameter metadata is set one time for the mock, regardless of how many times you use the Mock command for that target.  (Technically,this happens once per Describe block, or in the upcoming v4, once per test script; mocks get cleaned up after the top level test container goes away.)  Even if we update Pester's code to make sure that the most recent Mock's options are "live", there can still only be one version defined at a time.
### Option 3:  Your Idea Here.  :smile:

If you have a better idea on how to make mocking these commands less painful, please share!

### LINKED ISSUE #980: Mocking 'Get-PhysicalDisk' fails
## Issue

Mocking 'Get-PhysicalDisk' fails

## Environment
Pester version     : 4.1.1 C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Pester.psm1
PowerShell version : 5.1.17074.1002
OS version         : Microsoft Windows NT 10.0.17074.0

## Expected Behavior
run:
```Powershell
Describe "Get-PhysicalDisk mocking Tests" {
    Context "calling direct" {
        It "should return a result" {
            { Get-PhysicalDisk } | Should -Not -Throw
        }
    }
    Context "mocking it" {
        Mock Get-PhysicalDisk { return "hello" }
        It "should return 'hello'" {
            Get-PhysicalDisk | Should Be "hello"
        }
    }
}
Describe "Get-Disk mocking Tests" {
    Context "calling direct" {
        It "should return a result" {
            { Get-Disk } | Should -Not -Throw
        }
    }
    Context "mocking it" {
        Mock Get-Disk { return "hello" }
        It "should return 'hello'" {
            Get-Disk | Should Be "hello"
        }
    }
}
```

should return:
```powershell
Executing script .\Test.Integration.Tests.ps1

  Describing Get-PhysicalDisk mocking Tests

    Context calling direct
      [+] should return a result 524ms

    Context mocking it
      [+] should return 'hello' 116ms

  Describing Get-Disk mocking Tests

    Context calling direct
      [+] should return a result 2.61s

    Context mocking it
      [+] should return 'hello' 116ms
Tests completed in 3.35s
Tests Passed: 3, Failed: 1, Skipped: 0, Pending: 0, Inconclusive: 0
```

## Current Behavior

current result:
```powershell
Executing script .\Test.Integration.Tests.ps1

  Describing Get-PhysicalDisk mocking Tests

    Context calling direct
      [+] should return a result 524ms

    Context mocking it
      [-] Error occurred in Context block 94ms
        ParseException: At line:93 char:8
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +        ~
        Missing ] at end of attribute or type literal.
        At line:93 char:9
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +         ~
        Parameter declarations are a comma-separated list of variable names with optional initializer expressions.
        At line:93 char:9
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +         ~
        Missing ')' in function parameter list.
        MethodInvocationException: Exception calling "Create" with "1" argument(s): "At line:93 char:8
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +        ~
        Missing ] at end of attribute or type literal.
        At line:93 char:9
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +         ~
        Parameter declarations are a comma-separated list of variable names with optional initializer expressions.
        At line:93 char:9
        +     [Get-PhysicalDisk.PhysicalDiskUsage]
        +         ~
        Missing ')' in function parameter list."
        at Mock, C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Functions\Mock.ps1: line 322
        at <ScriptBlock>, C:\Users\peppe\OneDrive\GitHub\CS-Analytics\MachineMetrics\Tests\Integration\Test.Integration.Tests.ps1: line 8
        at DescribeImpl, C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Functions\Describe.ps1: line 161

  Describing Get-Disk mocking Tests

    Context calling direct
      [+] should return a result 2.61s

    Context mocking it
      [+] should return 'hello' 116ms
Tests completed in 3.35s
Tests Passed: 3, Failed: 1, Skipped: 0, Pending: 0, Inconclusive: 0
```

renehernandez:

Fix #601 
Fix #980

### DIFF
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -236,7 +236,7 @@ about_Mocking
             }
 
             # Will modify $metadata object in-place
-            $originalMetdata = $metadata
+            $originalMetadata = $metadata
             $metadata = Repair-ConflictingParameters -Metadata $metadata -RemoveParameterType $RemoveParameterType -RemoveParameterValidation $RemoveParameterValidation
             $paramBlock = [Management.Automation.ProxyCommand]::GetParamBlock($metadata)
 
@@ -334,7 +334,7 @@ about_Mocking
             Scope                   = $pester.CurrentTestGroup
             PesterState             = $pester
             Metadata                = $metadata
-            OriginalMetadata        = $originalMetdata
+            OriginalMetadata        = $originalMetadata
             CallHistory             = @()
             DynamicParamScriptBlock = $dynamicParamScriptBlock
             Aliases                 = @()
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -1580,13 +1580,11 @@ function Repair-ConflictingParameters {
         if ($RemoveParameterType -contains $paramMetadata.Name) {
             $paramMetadata.ParameterType = [object]
 
-            if (GetPesterPsVersion -ge 3) {
-                for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
-                    $attr = $paramMetadata.Attributes[$i]
-                    if ($attr -is [PSTypeNameAttribute]) {
-                        $null = $attrIndexesToRemove.Add($i)
-                        break
-                    }
+            for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
+                $attr = $paramMetadata.Attributes[$i]
+                if ($attr -is [PSTypeNameAttribute]) {
+                    $null = $attrIndexesToRemove.Add($i)
+                    break
                 }
             }
         }
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -1573,14 +1573,31 @@ function Repair-ConflictingParameters {
             $repairedMetadata.Parameters.Add($newName, $paramMetadata)
         }
 
+        $attrIndexesToRemove = New-Object System.Collections.ArrayList
+
         if ($paramMetadata.Name -in $RemoveParameterType) {
             $paramMetadata.ParameterType = [object]
 
-            # TODO: Add remove of pstypename attribute
+            for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
+                $attr = $paramMetadata.Attributes[$i]
+                if ($attr -is [PSTypeNameAttribute]) {
+                    $null = $attrIndexesToRemove.Add($i)
+                    break
+                }
+            }
         }
 
         if ($paramMetadata.Name -in $RemoveParameterValidation) {
+            for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
+                $attr = $paramMetadata.Attributes[$i]
+                if ($attr -is [System.Management.Automation.ValidateArgumentsAttribute]) {
+                    $null = $attrIndexesToRemove.Add($i)
+                }
+            }
+        }
 
+        foreach ($index in $attrIndexesToRemove) {
+            $null = $paramMetadata.Attributes.RemoveAt($index)
         }
     }
 
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -1576,7 +1576,7 @@ function Repair-ConflictingParameters {
 
         $attrIndexesToRemove = New-Object System.Collections.ArrayList
 
-        if ($paramMetadata.Name -in $RemoveParameterType) {
+        if ($RemoveParameterType -contains $paramMetadata.Name) {
             $paramMetadata.ParameterType = [object]
 
             for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
@@ -1588,7 +1588,7 @@ function Repair-ConflictingParameters {
             }
         }
 
-        if ($paramMetadata.Name -in $RemoveParameterValidation) {
+        if ($RemoveParameterValidation -contains $paramMetadata.Name) {
             for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
                 $attr = $paramMetadata.Attributes[$i]
                 if ($attr -is [System.Management.Automation.ValidateArgumentsAttribute]) {
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -1553,16 +1553,17 @@ function Repair-ConflictingParameters {
         $RemoveParameterValidation
     )
 
-    $repairedMetadata = [System.Management.Automation.CommandMetadata]::new($Metadata)
+    $repairedMetadata = New-Object System.Management.Automation.CommandMetadata -ArgumentList $Metadata
     $paramMetadatas = @()
     $paramMetadatas += $repairedMetadata.Parameters.Values
+
+    $conflictingParams = Get-ConflictingParameterNames
 
     foreach ($paramMetadata in $paramMetadatas) {
         if ($paramMetadata.IsDynamic) {
             continue
         }
 
-        $conflictingParams = Get-ConflictingParameterNames
         if ($conflictingParams -contains $paramMetadata.Name) {
             $paramName = $paramMetadata.Name
             $newName = "_$paramName"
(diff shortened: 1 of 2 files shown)


renehernandez:
@nohwnd The PowerShell 2 is failing and I don't understand the reason :(. My lack of PS 2 experience is biting me again :)

Functions/Mock.ps1
```diff
@@ -330,6 +333,7 @@ about_Mocking
             Scope                   = $pester.CurrentTestGroup
             PesterState             = $pester
             Metadata                = $metadata
+            OriginalMetadata        = $originalMetdata
```
renehernandez: Having the original `Metadata` at hand, would facilitate future operations where we may want to override the mock customizations in nested blocks

Functions/Mock.ps1
```diff
@@ -1540,28 +1544,65 @@ function Repair-ConflictingParameters {
...
+        }
+
+        $attrIndexesToRemove = New-Object System.Collections.ArrayList
+
+        if ($paramMetadata.Name -in $RemoveParameterType) {
+            $paramMetadata.ParameterType = [object]
+
+            for ($i = 0; $i -lt $paramMetadata.Attributes.Count; $i++) {
+                $attr = $paramMetadata.Attributes[$i]
+                if ($attr -is [PSTypeNameAttribute]) {
```
renehernandez: This removal is required, otherwise PowerShell won't be able to match the new `[object]` type definition for the parameter in the corresponding parameter set.

nohwnd:
@renehernandez the error is either 
1) using the `-in` operator, that is v3 thing, you need to use `-contains` and switch the order of the operands
2) might be the `-1` in one of the params, pretty sure it is not it but in case it is I's put it in `(-1)`

renehernandez:
@nohwnd The issue was the usage of `-in` operator. That is fixed. I had to comment the whole tests examples for `RemoveParameterType` because `PSTypeName` attr seems to not be available in PS 2.

renehernandez:
@nohwnd The current error in the PS 2 run is in the *Gherkin.Tests.ps1* suite. Any advice there?

nohwnd:
@renehernandez the tests on v2 are a bit flaky, I've just re-run them and they passed

renehernandez:
@nohwnd Awesome. 

Then, what actions are next? Should we move along with the idea of resetting the mocks? It sounds to me like that is to big of change in behavior and should be tackle separated for v5.

Any ideas?

nohwnd:
@renehernandez  you can do as many commits as you want I squash them into one when I merge anyway :) 

Yeah I think this should be released with the current limitations. Moving forward to redefining the mocks is hopefully a non-breaking change and can be done later if this proves to be useful. 

I fixed up the tests a bit so they prove more, and so I understand what is happening one year from now. I hope to release it later today.

DarkLite1:
@nohwnd  Are these changes also added to `Pester 5`?

nohwnd:
@DarkLite1 yes I merged them to v4 and v5.

nohwnd:
Neither is released with those changes yet though.

DarkLite1:
In `Pester 4.8.1` I tried the following command:
```powershell
Mock Remove-IpamObjectHC -RemoveParameterValidation ReferenceObject 
```

When in the `$sut` I try to do this:
```powershell
$OldSettings[-1] | Remove-IpamObjectHC @EnvParam -NoServiceRestart
```
Pester throws me the following error:
```powershell
PesterMock_43c736a9-b6ec-4ea3-93bc-7f083443d52b : Cannot bind argument to parameter 
'ReferenceObject' because it is null.
At S:\Test\IPAM Fixed address.ps1:219 char:40
+ ...    $OldSettings[-1] | Remove-IpamObjectHC @EnvParam -NoServiceRestart
```
I would've expected it to not look at the parameter `ReferenceObject` at all, maybe I'm missing something. Is there an example available somewhere?

nohwnd:
I'll have to try this next week, unless @renehernandez jumps on it :)

renehernandez:
@DarkLite1 Thanks for bringing this issue to our attention. I can confirm reproducing the issue with a simpler case:

```
Function Remove-IpamObjectHC {
    [CmdLetBinding()]
    Param (
        [Parameter(Mandatory, ValueFromPipeline)]
        [ValidateNotNullOrEmpty()]
        [Object[]]
        $ReferenceObject
    )
}

It 'removes ValidateNotNull from Mock' {
    Mock Remove-IpamObjectHC -RemoveParameterValidation ReferenceObject { }

    { $null | Remove-IpamObjectHC } | Should -Not -Throw
}
```

@nohwnd Will investigate :)

renehernandez:
@DarkLite1 @nohwnd #1321 keeps track of the issues and solution

PR #1280, opened 2019-03-31
Add TestDrive parameter to In so we can use it outside of blocks

nohwnd:


DarkLite1:
Would it be possible to add an example on how to use the `TestDrive` with the new Pester? Normally we do something like this:
````pwoershell
Describe $sut {
    in $TestDrive {
        Describe 'send an error mail to tha admin when' {
            It 'an ADGroupName has an incorrect quota limit at the end of its name' {
````

But this doesn't work anymore. What is the correct syntax for the new Pester?

nohwnd:
The syntax would be `In -TestDrive` but this might be for the old new runtime, so I am not sure if it works properly, probably not.

Issue #1290, opened 2019-04-19
Ability to specify conflicting parameter names?


FISHMANPET:
## 1. General summary of the issue

I believe this is related to #1007.  I see in the fix for that, the variable $script:ConflictingParameterNames was hardcoded which currently just contains PSEdition.  I happen to be mocking a function that contains a parameter called host, and I'm getting the same error that was originally reported in that bug.  As far as I can tell there's no way to specify conflicting parameters.  If manually add "host" to list of ConflictingParameterNames in Mock.ps1 my test succeeds.

This is the function I'm trying to mock:
```
function Send-SplunkHEC
	{
		<#
			.Synopsis
				Send event to Splunk HTTP Event Collector
			.DESCRIPTION
				Send event to Splunk HTTP Event Collector
			.PARAMETER uri
				URI for HEC endpoint
			.PARAMETER header
				contains auth token
			.PARAMETER host
				Part of Splunk Metadata for event.  Device data being sent from
			.PARAMETER source
				Part of Splunk Metadata for event.  Source 
			.PARAMETER sourceType
				Part of Splunk Metadata for event.  SourceType
			.PARAMETER metadata
				Part of Splunk Metadata for event.   Combination of host,source,sourcetype in performatted hashtable, will be comverted to JSON
			.PARAMETER eventData
				Event Data in hastable or pscustomeobject, will be comverted to JSON
		#>
		[CmdletBinding()]
		Param
		(
			# Param1 help description
			[Parameter(Mandatory)]
			[string]$uri,

			[Parameter(Mandatory)]
			[Collections.Hashtable]$header,

			[Parameter(Mandatory,ParameterSetName='Components')]
			[String]$host,

			[Parameter(Mandatory,ParameterSetName='Components')]
			[String]$source,

			[Parameter(Mandatory,ParameterSetName='Components')]
			[String]$sourceType,

			[Parameter(Mandatory,ParameterSetName='hashtable')]
			[Collections.Hashtable]$metadata,

			# This can be [Management.Automation.PSCustomObject] or [Collections.Hashtable]
			[Parameter(Mandatory)]
			$eventData
		)

		Begin{}
		Process
		{
			if ($metadata){$bodySplunk = $metadata}
			else {$bodySplunk = @{'host' = $host;'source' = $source;'sourcetype' = $sourcetype}}
			#Splunk takes time in Unix Epoch format, so first get the current date,
			#convert it to UTC (what Epoch is based on) then format it to seconds since January 1 1970.
[...]

PR #1292, opened 2019-04-22
Add detection for Constant or ReadOnly variable names

FISHMANPET:
Replace hard coded list of ConflictingParameterNames with code to detect all Constant and ReadOnly variables.  This comes from discussion in #1290, and is specifically my implementation of Rene's option 2

The Conflicting Parameter Names works around situations where the parameter in the mocked function has the same name as a fixed variable, and when Pester tries to use that variable it throws an error.  In older versions of Pester the error would include SessionStateUnauthorizedAccessException, and it looks like in newer ones, the message is "Cannot overwrite variable <variable> because it is read-only or constant."

Since this conflict currently only happens with read-only or constant variables (as the error message describes) I added code to specifically search for all variables that are "Constant" or "ReadOnly" and return their names in the Get-ConflictingParameterNames function.  I removed the PSEdition from the hardcoded string $script:ConflictingParameterNames, but I left the construct in, in case other conflict situations arise in the future.

There are two different ways to detect these variables, the human readable way that currently executes, and the more computer sciency way I left commented out.  The Options property of a variable is actually an Enum (https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.psvariable.options?view=powershellsdk-1.1.0) so you can do bitwise operations against it.  Or it seems to easily cast itself to a string, allowing string comparisons.  I'm not sure which would be more appropriate and in line with existing code so I put them both for reviewers to look at.

Also since I'm not very familiar with the overall code base and how verbose vs concise it is, I went with verbose, but if concise is desired I can shorten the code up.

I also don't see a lot of inline comments and the code seems self-explanatory to me, but if needed I can add some inline comments.

The current Mock.Tests all pass.  I looked at the test that was written when the ConflictingParameterNames code was written, and it looks like it only tests the repair of conflicted parameters, not the detection of them.  I'm not sure how I would unit test the code I wrote, or if it's needed, but if you have ideas I can try to implement those as well.


FISHMANPET:
OK I've clearly broken something I don't understand, it appears the tests that mock Get-Variable are causing my code, which uses Get-Variable to not work properly.  My first guess would be that when executing the test, which requires mocking Get-Variable, that instead of my code being executed, it's executing the mock of Get-Variable, which causes Mock to be called again, until it hits the max script call depth or some other failsafe that causes the code to fail.

nohwnd:
@FISHMANPET To avoid conflict when user mocks `Get-Variable` there is a hashtable of references called `SafeCommands`. That hashtable holds references to all the commands so you can invoke them even when they are mocked. 

You should also capture the list of variable names on the module load so you do it only once, (and possibly avoid using pipelines and cmdlets to make it faster). :)

As far as the code goes, neither of the options is ideal, the best way is to check using `-band` but instead of numbers you should use the actual enum values e.g. `[System.Management.Automation.ScopedItemOptions]::ReadOnly`.

Functions/Mock.ps1
```diff
@@ -1635,11 +1635,23 @@ function Reset-ConflictingParameters {
 }
 
 $script:ConflictingParameterNames = @(
-    "PSEdition"
+    #
 )
 
 function Get-ConflictingParameterNames {
```
renehernandez: Let's change the function goal a little bit here. Instead of just finding at runtime the read-only/constant variables, assign them to the original `$script:ConflictingParameterNames` variable.

With this change, we should change the function verb, should we use `Set-ConflictingParameterNames` or `Resolve-ConflictingParameterNames`?

nohwnd: It is still a `Get` you are getting the data and then saving it somewhere. `Set` would mean to set the variables in the system, and `Resolve` would mean that we are resolving it from something. I think `Get` is okay. :)

FISHMANPET: So you're thinking to to move `Get-ConflictingParamterNames` somewhere where it will get called when `Mock` is called, and have it set the `$script:ConflictingParameterNames` variable, then replace the two current instances of `Get-ConflictingParamterNames` with a call to the variable directly?

In that case `Get-ConflictingParamterNames` wouldn't be returning anything, and would instead be setting the variable, so I would think `Set-ConflictingParamterNames` would be more appropriate, but maybe I'm misunderstanding the change.

FISHMANPET: @renehernandez I see in another comment you mentioned calling this in Pester.psm1, I thought about that, but I'm not sure if that'd be entirely safe.  In theory Constant and ReadOnly variables are the automatic variables and they won't change, it is possible that a new automatic variable could be instantiated between loading the Pester module and running a test.  Additionally, the `New-Variable` command can create ReadOnly and Constant variables, so a user could indiscriminately be adding these types of potentially conflicting variables to their environment at anytime, so the safest time to look for them is when you call a Mock I would think

FISHMANPET:
@nohwnd I can't picture a way of selecting certain types of variables from the output of `Get-Variable` without piping into `Where-Object` since `Get-Variable` doesn't have a way to filter within the cmdlet, especially since whatever I write will have to run on PS2 (why I had to use `Select-Object` instead of just getting the `Name` Parameter from the collection directly).  But I've also spent the last 2 weeks buried in code that pipes all sorts of things into `Where-Object` so if there's a different way you can point me in the direction and I can figure something out.  For what it's worth I see there's already a number of pipes into `Where-Object` in `Mock`, particularly one for filtering out dynamic variables, a very similar situation to this one I think

Functions/Mock.ps1
nohwnd: I was thinking about it like this: `$script:ConflictingParameterNames = Get-ConflictingParamterNames` and then just refer to the script scoped variable anywhere else. So the function would still be a `get` function. No need to explicitly set the variable inside of the function. It is possible to do, but it is a very non-standard approach.

--- 

I don't think that detecting the read-only/constant variables at runtime is a good idea, it would be a very systematic solution, but the conflicts are rare, and we need this fix for variables that are built-in but not necessarily for the variables that user introduces. My main concern is performance of this solution, Mock is slow as it is, and searching through the variables in clean session takes some 3-8ms on my computer (ps 6 on macos), that is quite a lot to add to every mock declaration even though it is very unlikely to have this conflict.

renehernandez:
@FISHMANPET To avoid using `Where-Object` and pipelines you could iterate over the result of `Get-Variable` with a `foreach` loop and assign the corresponding names to the arrays inside the loop. Something along the lines is shown below:

```
[System.Collections.ArrayList]$constantVars = @() 
[System.Collections.ArrayList]$readOnlyVars = @() 
foreach ($var in (Get-Variable)) {
    if ($var.Options -like "*Constant*") {
        $null = $constantVars.Add($var.Name)
    }
    elseif ($var.Options -like "*ReadOnly*") {
        $null = $readOnlyVars.Add($var.Name)
    }
}
````

**Comment by nohwnd** [comment]
@renehernandez do we need two arrays? The check should also be done against the actual enum, not against string, e.g `$o -band [System.Management.Automation.ScopedItemOptions]::ReadOnly -or  $o -band [System.Management.Automation.ScopedItemOptions]::Constant`

**Comment by FISHMANPET** [comment]
I don't think all the tests on this will pass, they don't on my machine at least, but I think it's because the tests need to be rewritten for the changed behavior rather than the code not working (though I'm not sure it works either...)

I changed the two times that `Get-ConflictingParameterNames` was called to just return the `$script:ConflictingParameterNames`

So that leaves me not being sure if `Get-ConflictingParameterNames` is even being called in the right place, and also not sure how to test and mock the `$script:ConflictingParameterNames` variable.  Maybe it does make sense to have a `Set-ConflictingParameterNames` that runs in the module load or with Invoke-Pester, and then have `Get-ConflictingParameterNames` in Mock.ps1 that just returns the value of   `$script:ConflictingParameterNames`, that I can figure out how to mock (is that why it was setup that way in the first place, to allow it to be Pestered?)

File: Functions/Mock.ps1 [hunk 2]
```diff
@@ -1635,11 +1635,23 @@ function Reset-ConflictingParameters {
...
-    $script:ConflictingParameterNames
+    $ConstantVariables = Get-Variable | Where-Object {$_.Options -like "*Constant*"}
+    $ReadOnlyVariables = Get-Variable | Where-Object {$_.Options -like "*ReadOnly*"}
```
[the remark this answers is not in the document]
FISHMANPET: Did this and now the test that mocks get-variable doesn't crash

Functions/Mock.ps1
```diff
@@ -1560,7 +1560,7 @@ function Repair-ConflictingParameters {
     $paramMetadatas = @()
     $paramMetadatas += $repairedMetadata.Parameters.Values
 
-    $conflictingParams = Get-ConflictingParameterNames
+    $conflictingParams = $script:ConflictingParameterNames
```
renehernandez: As you mentioned in your last comment, using the `Get-ConflictingParameterNames` functions allows us to mock it during our tests executions.

Please revert this change

Functions/Mock.ps1
```diff
@@ -1618,7 +1618,7 @@ function Reset-ConflictingParameters {
     )
 
     $parameters = $BoundParameters.Clone()
-    $names = Get-ConflictingParameterNames
+    $names = $script:ConflictingParameterNames
```
renehernandez: As you mentioned in your last comment, using the Get-ConflictingParameterNames functions allows us to mock it during our tests executions.

Please revert this change

Functions/Mock.ps1
```diff
@@ -1634,14 +1634,18 @@ function Reset-ConflictingParameters {
...
 function Get-ConflictingParameterNames {
-    $script:ConflictingParameterNames
+    foreach ($var in (& $script:SafeCommands['Get-Variable'])) {
+        if (($var.Options -band [System.Management.Automation.ScopedItemOptions]::Constant) -or ($var.Options -band [System.Management.Automation.ScopedItemOptions]::ReadOnly)) {
+            $null = $script:ConflictingParameterNames.Add($var.Name)
+        }
+    }
 }
 
+Get-ConflictingParameterNames
```
renehernandez: Calling the function here would have a similar purpose as to call it from Pester.psm1 during module import. 

But having it run directly from Mock seems counterintuitive, at least from a developer perspective, since it is not common for functions to be invoked at the script level from any Functions script.

Please move to Pester.psm1

Functions/Mock.ps1
nohwnd: Imho this is fine, you just need to set the script scope variable before running the test, instead of mocking the function. In fact you can just resolve the variable without qualifier here, which will allow you to set it in the above scope by using the `InModuleScope` block without overwriting data for the rest of the module. This also means that you don't have to remember the data that were in the variable and putting them back. 

example: 

```powershell
Get-Module m | Remove-Module

$module = New-Module -Name m -ScriptBlock { 
    $script:ConflictingParameterNames = "a", "b"

    function Get-Names {
        "without qualification"
        $ConflictingParameterNames
        "with qualification"
        $script:ConflictingParameterNames
    }
} 

$module | Import-Module

Get-Names
"---"
# this works like InModuleScope,
# it runs the code in the module scope
& ($module) {
    $ConflictingParameterNames = "f"
    Get-Names
}
"---"
Get-Names
```

renehernandez:
@FISHMANPET Are you able to finish the pull request? Let me know, otherwise I can take it from here

FISHMANPET:
@renehernandez yeah still poking away at this, if I get to the point where I can't go any farther or I don't have time I'll let you know, but I'm about to work on the latest feedback.

FISHMANPET:
I moved the code to set the ConflictingParameters variable into Pester.psm1 and it will get called once when Pester is Invoked (it didn't look like anything else was really being called on module import so it didn't seem appropriate to put it there but if someone points out "where" in Pester.psm1 it would go I can move it.

The tests that failed are only for PS2 and related to parameterless mocks, I pulled from master and it looks like the last commit to master also failed those same 5 tests for PS2, so I'm guessing that's not my fault

Functions/Mock.ps1
FISHMANPET: I wasn't able to get this to work, as far as I could tell tests for conflicting where already being run in the Pester module context (using InModuleScope) but setting the variable unqualified still didn't work, so I've left the Get command in to return the variable to make testing easier.

Pester.psm1
```diff
@@ -1515,6 +1516,15 @@ function Assert-VerifiableMocks {
 
 }
 
+function Set-ConflictingParameterNames {
```
renehernandez: @FISHMANPET Could you move the function back to the Mock.ps1 file. The less we can pollute the Pester.psm1 file the better

Pester.psm1
```diff
@@ -1077,6 +1077,7 @@ function Invoke-Pester {
 
         $script:mockTable = @{}
         Remove-MockFunctionsAndAliases
+        Set-ConflictingParameterNames
```
renehernandez: @FISHMANPET Having this here means that it will get executed every time tests are run, which is what we are trying to avoid.

Just move the function invocation as the last line before the export-modulemember section in Pester.psm1 (line 1531)

FISHMANPET:
And now I've moved the `Set-ConflictingParameterNames` function back into Mock.ps1 and the call to the end of Pester.psm1 so it only gets called at module load time.

renehernandez:
@FISHMANPET I merged from master into your branch to get the fixes for PS 2

renehernandez:
@nohwnd This is ready to be merged.

@FISHMANPET Thank you for you contributions 👍 !!

nohwnd:
Is there a technical reason why the $script variable is set from the insides of the function, instead of using $script:var = Get-* ? E.g. is it easier to test?

renehernandez:
> 
> 
> Is there a technical reason why the $script variable is set from the insides of the function, instead of using $script:var = Get-* ? E.g. is it easier to test?

There is no technical reason per se for setting the variable inside the function, except that I think that dot sourcing the files into the psm1 module file shouldn't perform any assignment (avoiding change of state) and just load the functions into the files.

But it's definitely something that I am open to suggestions :).

To test and mock the conflicting parameters, we use the `Get-ConflictingParameterNames` function inside the `Repair-ConflictingParameters` function. That way we have a seam we can leverage in the tests

nohwnd:
> dot sourcing into module file shouldn't perform any assignment (avoiding change of state) and just load the functions into the files 

What if I defined the function by using a variable, would that be bad or good?:D `$function:f = { function body }`

(But seriously it does not change the state in this case it just defines a new "constant", same as if you import a function. You could potentially overwrite an existing function by dot sourcing a file, the same way you could overwrite $script:abc variable if it was already existing. In the end having a variable that is imported into the current state using the "import into current state operator" is way more logical to me, than having a function that has no output, but has a side effect of creating a script scoped variable. And which I must not forget to call. Plus other places already do it the first way. )

renehernandez:
@nohwnd Variable assignment is now performed in the Pester.psm1 file

renehernandez:
@nohwnd Any update on this?

PR #1293, opened 2019-04-23
Fix function definition in parameter filter

### LINKED ISSUE #1291: Cannot use ${function:}s for ParameterFilters anymore

Sometimes I need to use a mock input in multiple places, so I will define it as a function. PowerShell allows you to refer to a function using the syntax `${function:MyFunctionName}` where the ScriptBlock would be expected. This worked great in previous versions of Pester.

As of pull request #1240, introduced 2 months ago, I can no longer specify a parameter filter using this syntax.

```
Operating System, Pester version, and PowerShell version:
Pester version     : 4.7.2 C:\Users\e404sib\Documents\WindowsPowerShell\Modules\Pester\4.7.2\Pester.psd1
PowerShell version : 5.1.15063.1746
```

While I have 4.7.2 on my system, the issue is also present in `master` as of today (commit `ead3cbd67e31b2994192b060558395e2313af859`).

## 3. Expected Behavior

The function is mocked with an appropriate parameter filter. Here is a minimal test case.

File `test.psm1`:

```
function Write-TestMessage ([string]$User) {
    Write-Output "Hi $User, this is the original function!"
}
```

File `test.Tests.ps1`:

```
$here = Split-Path -Parent $MyInvocation.MyCommand.Path
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path).Replace('.Tests.ps1', '.psm1')
Import-Module -Force "$here\$sut"

Describe "Demonstration of the parameter filter bug" {
    $MY_USER = "Soren"

    function myTestMock {
        Write-Output "Mock output"
    }
    function myParameterFilter {
        "$User" -match $MY_USER
    }

    # Passes
    Context "The original unmocked function" {
        It "uses the output defined in the module" {
            Write-TestMessage -User $MY_USER | should -Be "Hi $MY_USER, this is the original function!"
        }
    }

    # Passes
    Context "This Mock works" {
        Mock Write-TestMessage ${function:myTestMock} -ParameterFilter {
            "$User" -match $MY_USER
        }

        It "uses the output defined in our mock" {
            Write-TestMessage -User $MY_USER | should -Be "Mock output"
        }
    }

    # Fails
    Context "This Mock throws an exception when created in the Context block" {
        Mock Write-TestMessage ${function:myTestMock} -ParameterFilter ${function:myParameterFilter}

        It "uses the output defined in our mock" {
            Write-TestMessage -User $MY_USER | should -Be "Mock output"
        }
    }
}
```

## 4. Current Behavior

An exception is thrown:

```
Context This Mock throws an exception when created in the Context block
      [-] Error occurred in Context block 0ms
        RuntimeException: You cannot call a method on a null-valued expression.
        at New-BlockWithoutParameterAliases, C:\Workspaces\Git\toolslang_powershell_TfsServerDeploy\pestertmp\Functions\Mock.ps1: line 1664
        at Mock, C:\Workspaces\Git\toolslang_powershell_TfsServerDeploy\pestertmp\Functions\Mock.ps1: line 350
        at <ScriptBlock>, C:\Workspaces\Git\toolslang_powershell_TfsServerDeploy\test.Tests.ps1: line 35
        at DescribeImpl, C:\Workspaces\Git\toolslang_powershell_TfsServerDeploy\pestertmp\Functions\Describe.ps1: line 199
```

## 5. Possible Solution

The error is on this line in master: https://github.com/pester/Pester/blame/master/Functions/Mock.ps1#L1664

I set a breakpoint here and it appears that with a true ScriptBlock, `$Block.Ast.EndBlock` is defined, but with a function there is no `Ast` member; instead, the same information is called `$Block.Body.Ast.EndBlock`. As a result, `$ast` is left null and the attempt to call the `.FindAll()` method on it yields the exception.

I don't know enough about PowerShell internals to know the best way to check for this case.

Off the top of my head, seems that it might be worth double-checking that closures work here as well, if you want to support them. I have not investigated.

## 6. Context

I don't think this is related to issue #1275 since it doesn't involve dynamic parameters, but that's the closest thing I see in open issues.

clcaldwell:
## 1. Fix bug where function definition ( ${function:} ) pass to -ParameterFilter breaks upon AST.

Fix #1291 

This is a small fix to to the 'New-BlockWithoutParameterAliases' function. When a function definition was passed to -ParameterFilter and internally passed to 'New-BlockWithoutParameterAliases', the AST would fail because it was expecting a true scriptblock, so the AST was in a different location than expected.

It now checks to see if the passed scriptblock is a FunctionDefinitionAst or ScriptBlockAst and handles the AST appropriately.

Tests included.

### DIFF
--- a/Functions/Mock.Tests.ps1
+++ b/Functions/Mock.Tests.ps1
@@ -2105,24 +2105,6 @@ if ($PSVersionTable.PSVersion.Major -ge 3) {
                 }
             }
 
-            Context 'Get-Content mock ParamFilter passed as function definition instead of scriptblock' {
-                BeforeAll {
-                    Function ParamFilter {
-                        $Tail -eq 100
-                    }
-                    Mock Get-Content { "default-get-content" }
-                    Mock Get-Content -ParameterFilter ${function:ParamFilter} -MockWith { "aliased-parameter-name" }
-                }
-
-                It "returns mock that matches parameter filter block" {
-                    Get-Content -Path "c:\temp.txt" -Last 100 | Should -Be "aliased-parameter-name"
-                }
-
-                It 'returns default mock' {
-                    Get-Content -Path "c:\temp.txt" | Should -Be "default-get-content"
-                }
-            }
-
             Context "Alias rewriting works when alias and parameter name differ in length" {
 
                 Mock New-Item { return "nic" } -ParameterFilter { $Type -ne $null -and $Type.StartsWith("nic") }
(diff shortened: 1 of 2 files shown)


Functions/Mock.ps1
```diff
@@ -1659,7 +1659,12 @@ function New-BlockWithoutParameterAliases {
     try {
         if ($PSVersionTable.PSVersion.Major -ge 3) {
             $params = $Metadata.Parameters.Values
-            $ast = $Block.Ast.EndBlock
+            if ($block.Ast -is [System.Management.Automation.Language.FunctionDefinitionAst]) {
+                $ast = $block.Ast.Body.EndBlock
+            }
+            elseif ($block.Ast -is [System.Management.Automation.Language.ScriptBlockAst]) {
```
nohwnd: What happens if it is neither? 

Please use same capitalization for `Block` with capital `B` :)

renehernandez: I have been reading about conversions to scriptblocks, but I think this should give us enough coverage. I was worried that .Net delegates could be transformed into scriptblocks with a different AST structure, but it does not seem to be the case

nohwnd: Okay, so if there is no third option then it should throw on `else`?

clcaldwell: Thanks for the quick feedback!

I've changed the casing on to match: $Block .

I also added a throw for the case of it getting through the function param validation [scriptblock] , but not having a type of 'FunctionDefinitionAst' or 'ScriptBlockAst' . This probably won't ever happen - but just in case.

Functions/Mock.ps1
```diff
@@ -1659,7 +1659,15 @@ function New-BlockWithoutParameterAliases {
     try {
         if ($PSVersionTable.PSVersion.Major -ge 3) {
             $params = $Metadata.Parameters.Values
-            $ast = $Block.Ast.EndBlock
+            if ($Block.Ast -is [System.Management.Automation.Language.FunctionDefinitionAst]) {
```
renehernandez: Could this be moved to a custom validation attribute?

If so, we could then apply the attr to both the `Mock` and `Assert-MockCalled` functions to catch problems even sooner

clcaldwell: To be sure we are talking about the same thing, you want a custom validator to set against to set against the $ScriptBlock Param, like this:

    Param(
      [ValidateNotNullOrEmpty()]
      [SetScriptBlockToAST()]
      $ScriptBlock
    )

If I'm not mistaken - for the custom validator, you need to either define this in c# or write a Powershell class. If Pester is still trying to support older version of Powershell - this will have to be c# - unless there is some other way to create a validator or I misunderstood what you meant.

If you want to be able to use this logic elsewhere - I personally would lean towards justing make this logic into it's own function (maybe Get-ScriptBlockAst or similar?). Then, this case would look like:

    Param( [ScriptBlock]$ScriptBlock )
    # blah blah
    $ast  = Get-ScriptBlockAst $ScriptBlock
    $blockText = $ast.Extent.Text

renehernandez: Good point. I tend to get carried away with the features of PS 5+ :).

Actually, I meant moving the if checks only, so they could be use to validate the parameter value. We would still need to do the check here inside the function to properly extract **EndBlock** property if it a `FunctionDefinitionAst` or a `ScriptBlockAst`.

On the other hand future changes could make the validation function and the if statements here get out of sync, so I think we should hold off for now.

clcaldwell: @renehernandez Are you sure? I'm thinking of just a trivial function, where we receive a scriptblock, and then always return a [ScriptBlockAst] . Then, you don't have to verify the type in any of the Mock functions.

    # pseudo
    Function Get-ScriptBlockAst {
      param ( [scriptblock]$Scriptblock )
      
      switch ($scriptblock) {
        $_.ast -is [FunctionDefinitionAst] { $returnblock = $_.body }
        $_.ast -is [ScriptBlockAst] { $returnblock = $_ }
        $_.ast -is [SomeOtherAst] { $returnblock = $_.some.path.body }     
        Default { Throw = "You have a magical block, who defies logic. Cherish him!" }

     Return $returnblock
    } # End Func

@nohwnd @renehernandez Thoughts on this? If you want to do it this way, I can modify the PR to match this.

nohwnd: The conversion function seems like a nice way to do it.

renehernandez: @clcaldwell The new function would help simplify the logic in `New-BlockWithoutParameterAliases` function, which I am all for it, but we would not be able to reuse it within the `ValidationScript` block (the original purpose of my comment) since the function does not return boolean.

clcaldwell: Ah, I see what you are saying now. I'll modify the PR to have this as a function just so this is a bit cleaner - and then let's get this merged if we can, the bug is squashed. 

I'll open a separate ticket to have a discussion on potential refactoring for scriptblock validation/conversion across the board in Pester.

nohwnd: @renehernandez validation script errors suck, it's much nicer to have a guard at the start of the function that returns a more reasonable message instead :)

Functions/Mock.Tests.ps1
```diff
@@ -2216,11 +2216,92 @@ Describe "Mock definition output" {
     }
 }
 
-Describe 'Mocking using ParameterFilter with scriptblock' {
-    $filter = [scriptblock]::Create( ('$Path -eq ''C:\Windows''') )
-    Mock -CommandName 'Test-Path' -ParameterFilter $filter
+Describe 'Mocking using ParameterFilter' {
```
renehernandez: @clcaldwell Some of these tests seem to be very similar. They could be candidate to be refactored using the **TestCases** option

PR #1294, opened 2019-04-23
Use real path to temp path on macOS and GetTempPath() otherwise when constructing TestDrive

SteveL-MSFT:
special case macOS to use /private/tmp which is the real folder path
otherwise, use GetTempPath() so that on Windows it returns the long path

On macOS, the `/tmp` path is a symlink to `/private/tmp`, this causes problems in tests that create a test file and later tries to validate the path:

```powershell
$testPath = Join-Path $TestDrive "foo"
...
$realPath | Should -BeExactly $testPath
```

This fails as $testPath will be something like: `/tmp/foo` but the real file path returned from some cmdlet or .NET API will be `/private/tmp/foo`.

On Windows, the problem is that `$env:TEMP` which is used to construct the TestDrive path uses the 8.1 path syntax.  So on AzDevOps, the user is called `VSSADMINISTRATOR` so `$env:TEMP` is `C:\Users\VSSADM~1\AppData\Local\Temp\` but similar to macOS where a test validates the paths, it will fail because a cmdlet or .NET API will often return the longpath `C:\Users\VSSADMINISTRATOR\AppData\Local\Temp\`.

On Linux, `GetTempPath()` simply returns `/tmp/` which is correct.

### DIFF
--- a/Functions/Environment.Tests.ps1
+++ b/Functions/Environment.Tests.ps1
@@ -92,7 +92,7 @@ InModuleScope -ModuleName Pester {
 
     Describe 'Get-TempDirectory' {
         It 'returns the correct temp directory for Windows' -Skip:((GetPesterOs) -ne 'Windows') {
-            $expected = $env:TEMP = "$($env:SystemDrive)\temp"
+            $expected = [System.IO.Path]::GetTempPath()
 
             $temp = Get-TempDirectory
             $temp | Should -Not -BeNullOrEmpty
--- a/Functions/Environment.Tests.ps1
+++ b/Functions/Environment.Tests.ps1
@@ -91,11 +91,8 @@ InModuleScope -ModuleName Pester {
 
 
     Describe 'Get-TempDirectory' {
-        It 'returns the correct temp directory for Windows' {
-            Mock 'GetPesterOs' {
-                'Windows'
-            }
-            $expected = $env:TEMP = "C:\temp"
+        It 'returns the correct temp directory for Windows' -Skip:((GetPesterOs) -ne 'Windows') {
+            $expected = $env:TEMP = "$($env:SystemDrive)\temp"
 
             $temp = Get-TempDirectory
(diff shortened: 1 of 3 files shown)


SteveL-MSFT:
I'll fix the test failures tomorrow

iSazonov:
Tests use https://github.com/pester/Pester/blob/600abd2bff8713c9f1ef78d6b176ee50c9bec4f1/Functions/Environment.Tests.ps1#L98
that doesn't change `[io.path]::GetTempPath()`

nohwnd:
Interesting, I thought the code was using the `GetTempPath()` already.

iSazonov:
@SteveL-MSFT Do we ready jump to latest Pester version in PowerShell Core repo? I ask because perhaps we need servicing fix for version we use.

SteveL-MSFT:
@iSazonov unless there's a bunch of test failures with latest Pester, I don't see a reason not to move to latest stable version of Pester

iSazonov:
In the case we should make this in PowerShell master branch before we move to .Net Core 3.0.

SteveL-MSFT:
@iSazonov we need to wait for a release first, but yes, we can take it in master branch

Functions/Environment.Tests.ps1
```diff
@@ -106,10 +103,10 @@ InModuleScope -ModuleName Pester {
             Mock 'GetPesterOs' {
```
nohwnd: The name of the test still contains the old path. "returns '/tmp' directory for MacOS"

nohwnd:
If you don't fix the name I will fix it and release on Saturday/Sunday.

Functions/Environment.Tests.ps1
SteveL-MSFT: Will fix

SteveL-MSFT:
@nohwnd I believe your comment has been addressed

PR #1298, opened 2019-05-01
Add list of all built-in readonly and constant variables

nohwnd:
Listed all built-in readonly and constant variables that are defined in PowerShell 5 and 6 and added them to conflicting names list.

Related #1292 #1290


renehernandez:
@nohwnd So do we still need/want the changes from the related pull request?

nohwnd:
@renehernandez @FISHMANPET imho it's worth to finish it. I mainly wanted to fix the related issue because I am making a release even though my solution is less futureproof.

Issue #1265, opened 2019-03-24
RFC: Better API for `Invoke-Pester`


nohwnd:
I am trying to simplify the interfaces on Invoke-Pester, to create a pit of success, where the usual stuff is simple easy to discover, and advanced stuff is less simple and less easy to discover. The way Pester v4 does that (to some extent) is by hiding hashtables behind parameters, this gives you a lot of freedom, but hardly a clean strongly typed interface. What I would like is a simple interface to be easy to start with, and the configurations to be explicit so I don't have to remember structure of hashtables. It would still be nice to have hashtables, so they can be used as literals in the code, modified and passed around easily.

## Simple

What I am assuming is that a typical person has a folder with tests, and want's to do some basic filtering on them. They also want to run them locally to develop, and run them on a build server easily. To me the simplest non restrictive parameter set is:

```powershell
Invoke-Pester 
 -Path <string[]>
 -ExcludePath <string[]>
 -Name <string[]>
 -ExcludeName <string[]>
 -Tag <string[]>
 -ExcludeTag <string[]>
 -PassThru <switch>
 -Parallel <switch>
 -Watch <switch>
 -Ci <switch>
 -Output <Normal, Short, None>
```
This would cover 99% of what I do. 😁 

This minimal api would then be typically used like this:

```powershell
Invoke-Pester
Invoke-Pester -Watch
Invoke-Pester -Path C:\projects\mymodule
Invoke-Pester -Path . -Ci
Invoke-Pester -Path . -Ci -PassThru | Export-NunitXML -Path "TestResults.xml"
Invoke-Pester -Tag "Integration*" -Ci -PassThru | 
    Export-NunitXML -Path "TestResults.xml"

Invoke-Pester -Path "*.Integration.Tests.ps1" -Ci -PassThru | 
    Export-NunitXML -Path "TestResults.xml"
```
`Invoke-Pester`
List of paths to run `*.Tests.ps1` from. When the path is a folder it will run all `*.Tests.ps1` files in it and its subfolders. When the path is a wildcarded file like `c:\tests\*.ps1` it will run all `*.Tests.ps1` in that folder, if that path is a full path, it will run the path, even if it is not a `.Tests.ps1` file.
Defaults to the current directory "." (Just like in v4).
`-Path <string[]>`

Excludes paths from the test run, it uses -like wildcards to match the paths, and automatically replaces '\' to '/' to make it work cross-plaform easily.
`-ExcludePath <string[]>`

[...]

PR #1307, opened 2019-05-05
Simple interface for Pester

nohwnd:
Invoke-Pester 
- [x] `-Path <string[]>`
- [x] `-ExcludePath <string[]>`
- [x] `-ScriptBlock <scriptblock[]>`
- [x] ` -Tag <string[]>`
- [x] `-ExcludeTag <string[]>`
- [x]  remove `-PassThru <switch>` 
- [x] `-CI <switch>`
- [x] `-Output <Normal, None>`

Most of the work here was to actually make sure that the NUnitXml output and Code Coverage output are comparable to the one in v4. We need some acceptance tests for this maybe :) 

Some more improvements are here as well. Like printing multiple errors.


nohwnd:
Related #1265

PR #1311, opened 2019-05-17
Mocking dynamic params

### LINKED ISSUE #1308: Cannot execute mock for Get-PackageSource cmdlet

Executing a mock of `Get-PackageSource` fails with *ParameterBindingException: Parameter set cannot be resolved using the specified named parameters.*

Example tests cases:

```powershell
Import-Module PackageManagement

Describe 'Mocked Get-PackageSource' {
    It 'should invoke our Mock' {
        Mock Get-PackageSource { 'Hello World' }

        Get-PackageSource | Should -Be 'Hello World'
    }
}

Describe 'Original Get-PackageSource' {
    It 'should produce valid results' {
        Get-PackageSource | Should -Not -BeNullOrEmpty
    }
}
```

The issue stems from the detection of the dynamic parameters for cmdlets when invoking the mock function, which is performed in `Get-DynamicParametersForCmdlet` function. Particularly, during the iteration of the [foreach loop](https://github.com/pester/Pester/blob/master/Functions/Mock.ps1#L1437), all the dynamic parameters are copied and returned to be used in the mock. And this fails because it is returning dynamic parameters that do not belong to the same **ParameterSet**.

```
Pester version     : master
PowerShell version : 5.1.17134.590
OS version         : Microsoft Windows NT 10.0.17134.0
```

## 3. Expected Behavior

Both tests should pass

## 4.Current Behavior

Tests with mock execution fails

## 5. Possible Solution

Detect which **ParameterSet** is selected at this point by the command invocation and only load/return the dynamic parameters that are part of it.

## 6. Context

This was derived from investigation into issue #339.

renehernandez:

Fix #1308

### DIFF
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -224,10 +224,6 @@ about_Mocking
     if (-not $mock) {
         $metadata = $null
         $originalMetadata = $null
-        $cmdletBinding = ''
-        $paramBlock = ''
-        $dynamicParamBlock = ''
-        $dynamicParamScriptBlock = $null
 
         if ($contextInfo.Command.psobject.Properties['ScriptBlock'] -or $contextInfo.Command.CommandType -eq 'Cmdlet') {
             $metadata = [System.Management.Automation.CommandMetaData]$contextInfo.Command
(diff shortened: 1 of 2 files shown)


renehernandez:
@nohwnd Will investigate the failures

renehernandez:
@nohwnd  There are 59 tests failing on PS 2 currently while all other build types are passing. Some of the failures are totally unrelated to my changes.

Any suggestions?

Functions/Mock.ps1
```diff
@@ -248,104 +244,12 @@ about_Mocking
...
-            }
-
-            end
-            {
-                $($newContent -replace '#BLOCK#', 'End' -replace '#INPUT#')
-            }
-"@
-
-        $mockScript = [scriptblock]::Create($code)
+        $mockBlocks = New-CodeBlocksForMock -ContextInfo $contextInfo -Metadata $metadata
```
renehernandez: The generation of the mock function body was moved to the new `New-CodeBlocksForMock` function.

It helps to have a cleaner and smaller `Mock` function by delegating the scriptblocks generation to the new function

Functions/Mock.ps1
```diff
@@ -1423,16 +1484,15 @@ function Get-DynamicParametersForCmdlet {
...
 
-        if ($null -eq $Parameters) {
-            $paramsArg = @()
-        }
-        else {
-            $paramsArg = @($Parameters)
-        }
+        $paramsArg = $InputArgs | ForEach-Object { $_.ToString() }
 
         $command = $ExecutionContext.InvokeCommand.GetCommand($CmdletName, [System.Management.Automation.CommandTypes]::Cmdlet, $paramsArg)
```
renehernandez: By passing all the input received from the `$args` variable instead of just the `$PSBoundParameters` it fixes the issue of restricting the parameter set based on the dynamic parameters themselves, since the `$args` variable will contain all the dynamic parameters specified that we want the original function to resolve

renehernandez:
@dlwyatt @nohwnd  One thing to notice is that the fix is currently applicable only if running in powershell version '5.0.10586.122' or higher.

I still need to try to devise a solution, if possible, for older versions of powershell. @dlwyatt do you have any suggestion for that particular part of the code?

renehernandez:
@dlwyatt @nohwnd  One thing to notice is that the fix  is currently applicable only if running in powershell version '5.0.10586.122' or higher.

I still need to try to devise a solution, if possible, for older versions of powershell. @dlwyatt do you have any suggestion for that particular part of the code? Starting at [line 1455](https://github.com/pester/Pester/blob/master/Functions/Mock.ps1#L1455) how could we use the $args information to determine the dynamic parameter sets?

Functions/Mock.ps1
```diff
@@ -1357,6 +1261,155 @@ function Get-DynamicParamBlock {
...
+                }
+            }
+            end {
+                `$receivedInput = @(`$input)
+
+                if (`$receivedInput.Count -gt 0) {
+                    `$receivedInput | Internal_$functionName @args
+                }
+                else {
+                    Internal_$functionName @args
```
renehernandez: @nohwnd I think I found part of the issue with PowerShell 2. It seems to be that splatting the `$args` variable to the new internal mock function does not work properly in PS 2.

I inserted a ```Write-Host `$(`$PSBoundParameters | Out-String)``` in the dynamicParam block to print the static parameters resolution using:

```
Mock Set-Item { return "I am not Set-Item" }
Set-Item "mypath" -value "value"
```

The invocation prints to console:
- PS 5.1 : 
Key   Value
---   -----
Value value
Path  {mypath}

- PS 2:
Key                                                                 Value
---                                                                 -----
Path                                                                {mypath}
Value                                                               -value

So, in PS 2 it is trying to map to the *-value* string as the input for the 
**-Value**. I suppose this is due to **-Value** being a positional parameter

Functions/Mock.ps1
```diff
@@ -1357,6 +1261,155 @@ function Get-DynamicParamBlock {
...
+                `$script:receivedArgs = `$args
+                function Internal_$functionName {
+                    $functionBody
+                }
+            }
+            end {
+                `$receivedInput = @(`$input)
+
+                if (`$receivedInput.Count -gt 0) {
+                    `$receivedInput | Internal_$functionName @args
```
renehernandez: Same issue as below

Functions/Mock.ps1
renehernandez: Based on this, we may want to consider merging this for pester v5 instead of master.

Thoughts?

renehernandez:
@nohwnd Do you agree with my last suggestion? Of changing the pull request against v5 and doing the review there?

nohwnd:
@renehernandez sorry, been out of the loop, preparations for psconfeu are my main priority now :) sure you can target v5, I have no problem with that :)

PR #1317, opened 2019-05-30
Cmdlet dynamic param

### LINKED ISSUE #1308: Cannot execute mock for Get-PackageSource cmdlet

renehernandez:

Fix #1308

### DIFF
--- a/Functions/Mock.ps1
+++ b/Functions/Mock.ps1
@@ -235,6 +235,9 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
         }
     }
     end {
+        `$internalFunction = & `$MyInvocation.MyCommand.Mock.Get_Item -Path "Function:Internal_$functionName"
+        & `$MyInvocation.MyCommand.Mock.Add_Member -InputObject `$internalFunction -MemberType NoteProperty -Name Mock -Value `$MyInvocation.MyCommand.Mock
+
         `$receivedInput = @(`$input)
 
         if (`$receivedInput.Count -gt 0) {
@@ -325,6 +328,8 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
         Get_Variable             = $SafeCommands["Get-Variable"]
         Invoke_Mock              = $InvokeMockCallBack
         Get_MockDynamicParameter = $SafeCommands["Get-MockDynamicParameter"]
+        Add_Member               = $SafeCommands["Add-Member"]
+        Get_Item               = $SafeCommands["Get-Item"]
         # returning empty scriptblock when we should not write debug to avoid patching it in mock prototype
         Write_PesterDebugMessage = if ($PesterDebugPreference.WriteDebugMessages) { { param($Message) & $SafeCommands["Write-PesterDebugMessage"] -Scope Mock -Message $Message } } else { {} }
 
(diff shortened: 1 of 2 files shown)


nohwnd:
I can review this after next week, right now all my time is spent on presentation for psconfeu.

Functions/Mock.ps1
```diff
@@ -226,6 +226,29 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
     }
 "@
 
+    $functionName = [Guid]::NewGuid().Guid
+    $code = @"
```
renehernandez: This is the essence of the change. By introducing a new level of indirection on the function definition we are able to leverage the `$args` values unfiltered to determine the expected dynamic parameters

Functions/Mock.ps1
```diff
@@ -743,14 +768,14 @@ function Invoke-MockInternal {
                         Write-PesterDebugMessage -Scope Mock "Original command is Set-Variable, patching the call."
                     }
                     if ($MockCallState['BeginBoundParameters'].Keys -notcontains "Scope") {
-                        $MockCallState['BeginBoundParameters'].Add( "Scope", 2)
+                        $MockCallState['BeginBoundParameters'].Add( "Scope", 3)
```
renehernandez: This is due to the actual function definition being wrapped in an internal function (to leverage `$args` content)

Functions/Mock.ps1
```diff
@@ -1249,16 +1282,15 @@ function Get-DynamicParametersForCmdlet {
         return
     }
 
+    if ($null -eq $InputArgs -or $InputArgs.Count -eq 0) {
```
renehernandez: if the `InputsArgs` is null or empty there is no point on trying to find dynamic parameters since the command invocation didn't specify them

nohwnd:
Me reviewing this code with very limited knowledge of dynamic params.

Screenshot 2019-06-19 at 19 03 53

nohwnd:
On a serious note, this is building and passing the new tests. So I merge it?:)

renehernandez:
Related to this, I think we need to figure out a way to express on which version of PowerShell a certain fix or feature is available

nohwnd:
@renehernandez do you have any ideas around that? I can put it in the change log if you think it has some value.

PR #1335, opened 2019-07-22
Compound assertion

### LINKED ISSUE #1325: Compound assertions as default & default PassThru on assertions
Writing tests with multiple assertions is not-recommended because it gives you just partial information on failure. The way to do this correctly is to write multiple It blocks, each having just a single assertion. Or we could allow assertions to tell the framework that there was an error, but not actually fail. This combined with outputting the asserted object, and maybe some extra assertions, would allow for chained assertions that only fail when the test is done, not at the first failed assertion:

```
Describe "d1" { 
    It "i1" {
        $user = Get-User

        $user | Should -NotBeNullOrEmpty -ErrorAction Stop

        $user | 
            Should -HaveProperty Name -Value "Jakub" |
            Should -HaveProperty Age  -Value 30 
    }
}
```

So here the first assertion would behave like the normal Pester assertion, it would fail the whole test, because it uses -ErrorAction Stop. The next two assertions would inform the framework that there is a new error and the test would fail in the end, but both of the assertions would run, not just the first one to fail.

I think this leads to a very readable tests, and would be a huge improvement over the current approach.

This issues drives the implementation for compound assertions from issue #1319. Further feature discussions are encouraged to take place here

renehernandez:

Fix #1325


renehernandez:
@nohwnd I am opening the pull request as a draft to get feedback earlier in the implementation phase

new-runtimepoc/Pester.Runtime.psm1
```diff
@@ -1388,8 +1394,9 @@ function Invoke-ScriptBlock {
                     }
 
                     if ($______parameters.EnableWriteDebug) { &$______parameters.WriteDebug "Running scriptblock { $($______parameters.ScriptBlock) }" }
+
                     $______parameters.CurrentlyExecutingScriptBlock = $______parameters.ScriptBlock
-                    . $______parameters.ScriptBlock @______innerSplat
+                    $______testResult = . $______parameters.ScriptBlock @______innerSplat
```
nohwnd: Am I reading this correctly that you are outputting an error record object from the test to the standard output and adding it to the error record? If yes then I don't think this is the way to go, and this is also not the place to implement this.

1) Why is it not the way to go? 

I commented a bit on it [here already](https://github.com/pester/Pester/issues/1325#issuecomment-510102303) but ultimately the user test code is free to output whatever it wants, and because in Pester v4 we are simply redirecting the output to $null there might be a lot of tests outputting data and this could be a breaking change for them. A better way would be for the internals to work as they do now (so exception is caught and added to the list of ErrorRecords - this will keep Should -ErrorAction Stop work, and also all assertions that simply throw exceptions, and any code that throws exceptions.) And then additionally Should (without -ErrorAction Stop) would add error record to the ErrorRecord collection via `Get-CurrentTest`. This frees the standard output of Should from outputting errors, because we need to output data, so we can chain assertions. And it will also prevent us from trying to detect if the output is an error record, because we will be simply looking at ErrorRecord collection, and if that is empty then it went fine.

2) Why is this not the place to implement it? 

This function is responsible for invoking the scriptblocks for Block and Test with the correct scoping, but should not be concerned with the type of the executed block. As you can see in 1408-1424 you are sometimes missing the test context, and that hints that this code does not really belong there. Instead this should be processed further up. The function is purposefully written in a way that the standard output goes to the standard output, and that exceptions bubble up.

Functions/Assertions/Should.ps1
```diff
@@ -112,36 +166,70 @@ function Invoke-LegacyAssertion($assertionEntry, $shouldArgs, $valueToTest, $fil
...
-        #extract data to return if there are any on the object
-        $data = $testResult.psObject.Properties.Item('Data')
-        if ($data) {
-            $data.Value
+    try {
+        # Write-Host "Bound Parameters: $($PSBoundParameters | Out-String)"
+        $testResult = & $AssertionEntry.Test -ActualValue $ValueToTest -Negate:$Negate -CallerSessionState $CallerSessionState @BoundParameters
+        # Write-Host "Test result: $testResult"
+        if (-not $testResult.Succeeded) {
+            New-ShouldErrorRecord -Message $testResult.FailureMessage -File $file -Line $lineNumber -LineText $lineText
```
nohwnd: This is imho where there "magic" happens. With `if ($shouldThrow) { throw $shouldErrorRecord } else { $currentTest.ErrorRecords.Add($shouldErrorRecord) }`. And that is about it. At least that is how I see it with my naive view right now :)

new-runtimepoc/Pester.Runtime.psm1
renehernandez: I think I got a better sense now of what you were trying to convey in the linked comment :).

nohwnd: Great, if you have any other questions, or if you'd like me to provide a draft of the changes, so you can better see what I meant or anything like that then let me know :)

renehernandez:
@nohwnd The pull request is ready for review

nohwnd:
@renehernandez can't do it right now, but will try to get to it within week. Working on some presentations. Sorry :)

renehernandez:
@nohwnd let me know when can we review this pull request

Dependencies/Axiom/Verify-AssertionFailed.ps1
```diff
@@ -4,24 +4,37 @@ function Verify-AssertionFailed {
...
     $err = $null
+    $assertionExceptionThrown = $false
     try {
         $null = & $ScriptBlock
     }
     catch [Exception] {
         $assertionExceptionThrown = ($_.FullyQualifiedErrorId -eq 'PesterAssertionFailed')
         $err = $_
-        $err
+        Write-Output $err
```
nohwnd: What is the reason to use `Write-Output` here?

Dependencies/Axiom/Verify-AssertionFailed.ps1
```diff
@@ -4,24 +4,37 @@ function Verify-AssertionFailed {
...
+        Get-CurrentTest
     }
 
+    if (-not $assertionExceptionThrown -and $test.ErrorRecord.Count -gt 0) {
+        $pesterAssertionErrors = $test.ErrorRecord | Where-Object { $_.FullyQualifiedErrorId -eq 'PesterAssertionFailed' }
+        $assertionExceptionThrown = $null -ne $pesterAssertionErrors
+        Write-Output $test.ErrorRecord
+    }
+
+    $test.ErrorRecord.Clear()
```
nohwnd: Why are you modifying Pester's own state?

Functions/Assertions/HaveParameter.Tests.ps1
```diff
@@ -2,87 +2,92 @@ Set-StrictMode -Version Latest
...
-                        if (-not (Test-Path $_)) {
-                            $errorItem = [System.Management.Automation.ErrorRecord]::new(
-                                ([System.ArgumentException]"Path not found"),
-                                'ParameterValue.FileNotFound',
-                                [System.Management.Automation.ErrorCategory]::ObjectNotFound,
-                                $_
-                            )
-                            $errorItem.ErrorDetails = "Invalid path '$_'."
-                            $PSCmdlet.ThrowTerminatingError($errorItem)
+    $script:functionsBlock = {
```
nohwnd: Why is this necessary? The functions seem to be in scope and you don't need to import them again in `BeforeAll`. Are you changing this to make the code cleaner? Because this change is really hard to review and unless it is totally necessary I'd skip it.

Functions/Assertions/Should.ps1
```diff
@@ -53,17 +53,18 @@ function Should {
 
     begin {
         $inputArray = [System.Collections.Generic.List[PSObject]]@()
+        $lineNumber = $MyInvocation.ScriptLineNumber
```
nohwnd: Why did those definitions need to move to begin? I don't see them used until `end` where they were defined before.

Functions/Assertions/Should.ps1
```diff
@@ -53,17 +53,18 @@ function Should {
...
+        $lineText = $MyInvocation.Line.TrimEnd("$([System.Environment]::NewLine)")
+        $file = $MyInvocation.ScriptName
     }
 
     process {
         $inputArray.Add($ActualValue)
+
+        $ActualValue
```
nohwnd: Returning the input value is probably not what we always want, is it? For example when doing `Should -Throw` will that assertion return the exception that was caught and not the input values?

Functions/Assertions/Should.ps1
```diff
@@ -157,3 +198,19 @@ function Format-Because ([string] $Because) {
 
     " because $($bcs -replace 'because\s'),"
 }
+
+function Invoke-LegacyAssertion($assertionEntry, $shouldArgs, $valueToTest, $file, $lineNumber, $lineText) {
```
nohwnd: If this is not needed please delete it.

new-runtimepoc/Pester.Runtime.psm1
```diff
@@ -597,9 +597,9 @@ function Invoke-TestItem {
                 -NoNewScope
 
             $Test.FrameworkData.Runtime.ExecutionStep = 'Finished'
-            $Test.Passed = $result.Success
+            $Test.ErrorRecord += ($result.ErrorRecord | Where-Object { $_.FullyQualifiedErrorId -notlike '*PesterAssertionFailed*' })
```
nohwnd: Why is this filter needed? Any error in the test should mean that the test failed, not just errors from assertions. Also don't use cmdlets :)

new-runtimepoc/Pester.Runtime.psm1
```diff
@@ -597,9 +597,9 @@ function Invoke-TestItem {
...
+            $Test.Passed = $Test.ErrorRecord.Count -eq 0
```
nohwnd: Imho the `result` should determine whether or not the test was successful. And this piece of code should just copy the values from the result to the `Test` structure.

new-runtimepoc/Pester.Runtime.psm1
```diff
@@ -856,6 +856,7 @@ function New-TestObject {
         Executed          = $false
         ExecutedAt        = $null
         Passed            = $false
+        Result            = $null
```
nohwnd: Where is this used and what it means? I can't find any usage, and since the runtime only supports passed / skipped / failed, this is not needed (yet) because all the states can be determined from the data that are on the object (eg. -not ShouldRun = skipped, ShouldRun -and Executed -and Passed = success).

new-runtimepoc/Pester.Runtime.psm1
```diff
@@ -1057,9 +1058,12 @@ function Run-Test {
     [CmdletBinding()]
     param (
         [Parameter(Mandatory = $true)]
-        [PSTypeName("DiscoveredBlock")][PSObject[]] $Block,
```
nohwnd: Rest of the changes in this file are just unnecessary formatting changes, please revert :)

Dependencies/Axiom/Verify-AssertionFailed.ps1
renehernandez: There is no change of behavior there, it was just a matter of making more clear the intent that the `$err` variable is to be sent as output to the pipeline.

I can revert that, but I think that is something that we should discuss as part of our coding practices. It could be in a separate issue of course.

Dependencies/Axiom/Verify-AssertionFailed.ps1
renehernandez: In order to verify that each assertion is producing a corresponding error record in case of failure, we need to send the error records to the pipeline within Verify-AssertionFailed. Once we do that then, in the Test results generation stage it would double the error records results since they were being received from the pipeline and where already stored in the Test context

renehernandez: You can see examples of this as part of the new tests added in Functions/Assertions/Should.Tests.ps1

Functions/Assertions/Should.ps1
renehernandez: You want me to delete the legacy assertion function? I am down to that :)

Functions/Assertions/Should.ps1
renehernandez: Good point. But should we consider `Should -Throw` a one-off scenario? Because the idea is to pass the original input to subsequent Should cmdlets, which in this case would be a `scriptblock`.

Or do you prefer to include a check for `Should -Throw`? I believe this will require us to hightlight it somehow in the documentation

new-runtimepoc/Pester.Runtime.psm1
renehernandez: It is actually doing the opposite :). It is adding the errors that are not from assertions, the assertions ones are already captured as part of the `Invoke-Assertion` function in Should.ps1

I will move away from Where-Object to `foreach`

new-runtimepoc/Pester.Runtime.psm1
renehernandez: Within the `Invoke-ScriptBlock` function, the Success property is set based on the ErrorRecord count from the recursive invocation of the *wrapperScriptBlock* object. The issue is that that ErrorRecord property is only being filled when the test throw an exception, which is not the only valid case anymore, now assertions will add an existing ErrorRecord to the current test without throwing an exception, unless `-ErrorAction Stop` is specified

new-runtimepoc/Pester.Runtime.psm1
renehernandez: Will remove. I think it is a leftover from when I was trying some other things

Functions/Assertions/HaveParameter.Tests.ps1
renehernandez: Without the dot sourcing inside the Describe block, `Get-Command` complains that it cannot find the functions

renehernandez: @nohwnd Logs from the build showing the issue: [HaveParameter tests](https://nohwnd.visualstudio.com/Pester/_build/results?buildId=449&view=logs&jobId=986454e2-4a53-5b4f-19e5-426b1a182c9d&taskId=22f2c122-d6a2-5866-e1bb-4f38fae86508&lineStart=2825&lineEnd=2826&colStart=1&colEnd=1)

Functions/Assertions/Should.ps1
nohwnd: The scriptblock does not matter on the output, at least not as much as the exception. Getting the exception from the failed scriptblock is way more practical and interesting than getting the scriptblock. So yes Throw is an exception. Should Throw should return the error record, and Should -Not Throw should return nothing. IMO

Functions/Assertions/Should.ps1
nohwnd: yup

new-runtimepoc/Pester.Runtime.psm1
nohwnd: Okay. Would that change the order of errors, though? Why are we capturing the errors in two different collections? Would this work with -ErrorAction Stop? It seems a bit overcomplicated. But we can refine later I guess.

new-runtimepoc/Pester.Runtime.psm1
nohwnd: Then the determination of success should move somewhere where it still makes sense and can be done in one step. Ideally. I am not sure what the possibilites are here :)

Functions/Assertions/HaveParameter.Tests.ps1
renehernandez: @nohwnd It may be related to [this issue](https://github.com/PowerShell/PowerShell/issues/8022) in PowerShell. Since the blocks discovery and execution is now moved to the internal *Pester.Runtime* module, `Get-Command` seems to be impacted

Functions/Assertions/Should.ps1
renehernandez: Reverted

Functions/Assertions/Should.ps1
renehernandez: Done

new-runtimepoc/Pester.Runtime.psm1
renehernandez: Done

new-runtimepoc/Pester.Runtime.psm1
renehernandez: Done

Functions/Assertions/HaveParameter.Tests.ps1
renehernandez: @nohwnd  How do you want to proceed here? The workaround mentioned there is what I was already doing, dot-sourcing the functions so they area reachable. Another alternative would be to define the functions in the global scope

Functions/Assertions/Should.ps1
renehernandez: This would be another change of behavior. Currently the error record is sent in the pipeline if the `PassThru` switch is specified. Do you want me to remove the PassThru parameter and return the error record everytime?

Functions/Assertions/HaveParameter.Tests.ps1
nohwnd: Yeah true, I forgot. Just keep it :)

Functions/Assertions/Should.ps1
nohwnd: Yes please, otherwise the assertion chaining would really suck. You'd have to specify -PassThru all the time.

Dependencies/Axiom/Verify-AssertionFailed.ps1
renehernandez: @nohwnd What is your take on this?

new-runtimepoc/Pester.Runtime.psm1
renehernandez: As far as I know, there were no order of errors before, in version 4.x you only get one *ErrorRecord* per It, so it is a breaking change anyways. This works with `-ErrorAction Stop`, but in that case you may not get a full list of errors, only the errors up to the point of the assertion with ErrorAction specified. The new order, first all assertion errors and then whatever error happened to be captured during execution

Any suggestions about this?

new-runtimepoc/Pester.Runtime.psm1
renehernandez: Uhm, I am not sure how to proceed here. We can either extract it to a separate function now, or wait until a more general use case appears

renehernandez:
@nohwnd Do you have any suggestion about the pending comments? Otherwise, we could move with the code as is, and refactor as the need arises

fourpastmidnight:
I'm not really sure what this PR is trying to accomplish. If I read the test examples in this PR correctly, it appears that what you want is the ability to state all of your assertions up-front, using a pipeline, all the while collecting the results, and then take the appropriate action: failing or passing the test.

I would submit that:

1. This is overly complex
2. It's strictly not necessary.
3. It may even be a unit testing anti-pattern

As we all know, the goal of a unit test is to verify behavior of the SUT. I'm in the camp of a unit test should contain one logical assertion (and that assertion _may_ consist of one or more physical assertion statements/expressions). So, I may have something like this:

```powershell
$result.Parent | Should -Not -Be Null
$result.Children | Should -HaveCount $expectedValue
```

OK, so some people don't like this because when the unit tests run, if `$result.Parent` _is_ null, then the test fails right then and there on that first assertion. So then they go and fix that issue, only to have the test fail again, only this time on the second assertion. Repeat this a few dozen times over a few dozen tests, and you get frustrated and you start screaming that unit tests aren't worth it because they're too expensive to maintain.

But there is a solution to this! Create a "mini-DSL" for your tests. It doesn't need to be fancy. For example, you may want to start with something _really_ simple in this case. Let's say you have a number of places in your tests that require you to check that the result's parent is not null and that it has the expected children count. You could just create a PowerShell function for that (bear with me, this example is _extremely_ trivial, but serves to illustrate my point):

```powershell
Function Has-ParentAndExpectedChildrenCount {
    param(
        [Parameter(ValueFromPipeline)]
        $Result,

        [int]$ExpectedCount
    )

    $TestResult = [PSCustomObject]@{
        Success = $True
        Messages = @()
    }

    if (!$result.Parent) {
        $TestResult.Success = $False
        $Messages += 'The object does not have a parent'
    }

    if (!@($result.Children).Length) {
        $TestResult.Success = $False
        $Messages += "The object does not have the expected number of children. Expected: $ExpectedCount    Actual: $(@($result.Children).Length)"
    }

    $TestResult
}

It 'Should have a parent and the expected number of children' {
    $TestResult = $Result | Has-ParentAndExpectedChildrenCount
    $TestResult | Should -BeTrue -Because $TestResult.Messages
}
```

Now, that _should_ work. Let's say, however that you have many many such places, and you may have many, many such "compound assertions" in your test suite. Well then you could take this concept one step further and create your own assertion operators! See `Add-AssertionOperator` in `Pester.psm1` to give you more of a "standard look and feel" for what you want to assert in your test suite. In the documentation for that function is even an example of how to create such an operator. So now, using the function I created above, let's turn it into a custom assertion operator:

```powershell
function Has-ParentAndExpectedChildren {
    param (
        $ActualValue,
        [int]$ExpectedCount
    )

    Process {
        $Success = $True
        $Messages = [string[]]@()

        if (!$result.Parent) {
            $Success = $False
            $Messages += @'
The object does not have a parent
Expected: the object has a non-null parent
 but got: $null
'@
        }

        if (!@($result.Children).Length) {
            $Success = $False
            $Messages += @"
The object does not have the expected number of children.
Expected: $ExpectedCount 
 but got: $(@($ActualValue.Children).Length)
"@
        }
    }

    End {
        $FailureMessage = ''
        if (!$Success) {
            $FailureMessage = "The following conditions were not met:`r`n$(Messages -join "`n")"
        }

        return [PSCustomObject]@{
            Succeeded = $Success
            FailureMessage = $FailureMessage
        }
}

Add-AssertionOperator -Name HaveParentAndExpectedChildrenCount -Test $function:Has-ParentAndExpectedChildren

Describe 'Foo' {
    Context 'Bar' {

        It 'Should have a parent and the expected number of children' {
            $Result | Should -HaveParentAndExpectedChildrenCount 4
        }
    }
}
```

As you can see, using the built-in facilities of the existing test framework allows you to accomplish the same thing: create "compound assertions" via the use of custom assertion operators. Plus, you have the benefit of creating functions and operator names for them which makes sense in the context of your application's domain, thereby making the tests much more understandable and readable.

nohwnd:
@fourpastmidnight what you described is pretty much what the PR is trying to achieve. This is a new feature in Pester v5 so we can chain assertions and use them to test more complex objects. 

This is very practical because the test gives more complete information about the failures, and you can still force the assertion to fail fast by specifying -ErrorAction Stop. Also if you write one assertion per test, and the assertion is at the bottom of the test, nothing will change for you. Just the internals error reporting mechanism will change, the Should assertion will tell It block that it failed, instead of throwing an exception. Of course any exception will still fail the test, so all misbehaving code, as well as custom assertions that throw exceptions will still work.

fourpastmidnight:
@nohwnd Gotcha. I suppose where a feature like this shines, that I hadn't fully thought through until now, is in cases where you don't have a sufficiently complex domain/test usite to justify creating a "mini-DSL" as suggested. This could be a real time-saver in those situations. I guess the old adage applies: "With great power, comes great responsibility." 😄

renehernandez:
@nohwnd When can we tackle this PR?

nohwnd:
@renehernandez I did not forget about this, when I have time I hack on the #1307 which is nearing donness, and contains a lot of changes, including stuff that publishes multiple errors nicely. So when that is merged to the main branch this PR is next. Please hold on for few more days 🙏

nohwnd:
Checked the code yesterday, changed the implementation to require no change to the runtime. I now understand that my idea was valid but very hard to implement without knowing how exactly the new internals work. And it still feels a bit like a hack without the runtime providing an explicit service to add the errors to the current execution, but overall it should be cleaner because we don’t lose the order of errors, and don’t need to collect them afterwards, or make runtime aware of what it is executing. 

I also moved some of the tests to use the minimal test framework (P) it makes the high level tests easier to read and debug.

Thanks for the initial implementation, it helped a lot ☺️

renehernandez:
@nohwnd So, are you going to finish the PR?

renehernandez:
Reading, your comments and changes, are we moving away from chained assertions? Will it be possible to have multiple assertions in the same `It` block?

nohwnd:
@renehernandez Yes, I will finish it. I just needed a break yesterday 🙂

>Reading, your comments and changes, are we moving away from chained assertions? Will it be possible to have multiple assertions in the same It block?

By chained assertions I mean just this: 

```powershell 
$user | 
    Should -HaveProperty Name -Value "Jakub" |
    Should -HaveProperty Age  -Value 30
```
That is not much easier than doing this, in fact this seems easier, and the user will get similar contextual information in the error / stack trace: 
```
$user.Name | Should -Be "Jakub"
$user.Age | Should -Be 30
```
This approach is also already working and does not require `Should` to output stuff into the pipeline, which would be then captured in `StandardOutput` property on the result object. 

So multiple assertions will remain possible and will not throw by default, I just won't implement the case where we are chaining `Should`s by pipeline.

Issue #1342, opened 2019-08-06
Log the version of Pester that is used to run tests at the top of the output


KirkMunro:
## General summary of the issue

When Pester tests pass on one machine and fail on another, users must be able to easily identify the version of Pester used in both cases.

## Detailed request

Since Pester is software like any other software, it can contain bugs. For this reason, all Pester output should log the version of Pester used to generate that output at the top of the output. That way, when there are discrepancies (the tests work on my machine), the Pester version can be checked to make sure that something that appears off isn't due to a bug that may have been fixed already in Pester.

## Possible Solution

Please consider adding the version of Pester used to run tests to the top of the output for the reasons outlined above.

## Context

I'm facing a scenario where a test that runs successfully locally fails in ADO CI builds, and I don't know why. I'd like to compare the version of Pester they use with the one I use and make sure I'm using the same version to test locally so that I can verify whether or not the issue has to do with Pester.

nohwnd:
```
    ____            __           
   / __ \___  _____/ /____  _____
  / /_/ / _ \/ ___/ __/ _ \/ ___/
 / ____/  __(__  ) /_/  __/ /    
/_/    \___/____/\__/\___/_/                                  
Pester v4.8.0
```

So it is finally time to have some ascii art banner? 🙌

nohwnd:
I think it should be enough to edit the [start message](https://github.com/pester/Pester/blob/ead3cbd67e31b2994192b060558395e2313af859/Functions/Output.ps1#L3) to include the version of the module (the ascii art is unnecessary imho). Will you send a PR?

KirkMunro:
Sure, I'll send a PR for that. Just give me a bit to finish what's in front of me now.

KirkMunro:
And I like the ASCII art. Unnecessary, sure, but retro is cool.

nohwnd:
Should be implemented in the v5 as well. Not sure if simple forward merge will do the trick. So I am keeping this open even though it is implemented for v4.

martin9700:
Think it would make sense to also include it in the Passthru object.

fflaten:
Still sure about the ASCII-art in v5? 😅🤘

nohwnd:
Imho we should output the version in the detailed mode, which we don't do now. The ascii art had surprisingly negative feedback, so let's skip that.

nohwnd:
[...]

PR #1343, opened 2019-08-07
Add header to Pester output that shows the version

KirkMunro:

Fix issue #1342.

Told you I think ASCII art and most things retro are 😎 .

### DIFF
--- a/Functions/Output.ps1
+++ b/Functions/Output.ps1
@@ -1,5 +1,13 @@
 $Script:ReportStrings = DATA {
     @{
+        HeaderMessage     = @'
+    ____            __
+   / __ \___  _____/ /____  _____
+  / /_/ / _ \/ ___/ __/ _ \/ ___/
+ / ____/  __(__  ) /_/  __/ /
+/_/    \___/____/\__/\___/_/
+Pester v{0}
+'@
         StartMessage      = "Executing all tests in '{0}'"
         FilterMessage     = ' matching test name {0}'
         TagMessage        = ' with Tags {0}'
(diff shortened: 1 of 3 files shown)


renehernandez:
@KirkMunro It is failing to find the new HeaderMessage property in powershell 2

KirkMunro:
FYI, the reason I did this as a new `HeaderMessage` string is because I thought that Gherkin and anything else that customizes the `StartMessage` wouldn't need to be changed as well (I assumed their string tables would only contain items that they wanted to change, not that they'd have to contain all strings. Turns out that was an incorrect assumption.

Maybe the header message, which does not contain anything that needs to be localized, should simply be defined inline without an entry in the string table at all. If you want that changed, let me know and I'll update it.

nohwnd:
Looks good, and I also checked it visually in the output. Merging.

Greg-Smulko:
@KirkMunro , the version number can be really useful, thanks for doing it. 

Just a quick question, I assume it's not possible as for now to disable the ASCII art logo? You know, I'm not against ASCII art overall, but in this case, it just pollutes our build log enormously...

KirkMunro:
@Greg-Smulko There's no built-in support for that right now, no. The ASCII art isn't a requirement though. Seemed appropriate/fun at the time, but if it's just causing pollution and getting in the way of business, maybe we could either remove it altogether or have a switch to turn the logo off (i.e. -NoLogo, like PowerShell has). Thoughts @nohwnd?

nohwnd:
@KirkMunro agree the logo was fun, but I saw way to many people complain about it. Let's revert that. :)

PR #1347, opened 2019-08-13
Add JUnitXml to OutputFormat

### LINKED ISSUE #1346: Add JUnitXml as OutputFormat
Some CI systems (Like GitLab) don't support NUnitXml but they do support JUnitXml. I'm working on adding JUnitXml as an additional OutputFormat to Pester [here](https://github.com/bgelens/Pester/tree/junitxml). Once it is done, I'll submit a PR

For the time being, I created a [sample function](https://gist.github.com/bgelens/03907b83dd9b2cf7104f284a08f1b2fc) that converts the PassThru object into JUnitXml. The output is working but is not checked against a schema yet. 

I'm planning to make the output compliant with this [schema](https://github.com/junit-team/junit5/blob/master/platform-tests/src/test/resources/jenkins-junit.xsd)

bgelens:

Fixes #1346 

This PR will be in draft so progress on the JUnitXml implementation can be tracked / discussed. Once done, it will be promoted to a generic PR

### DIFF
--- a/Functions/TestResults.ps1
+++ b/Functions/TestResults.ps1
@@ -33,9 +33,9 @@ function Export-PesterResults {
         [string] $Format
     )
 
-    switch ($Format) {
-        'NUnitXml' {
-            Export-NUnitReport -PesterState $PesterState -Path $Path
+    switch -Wildcard ($Format) {
+        '*Xml' {
+            Export-XmlReport -PesterState $PesterState -Path $Path -Format $Format
         }
 
         default {
@@ -43,13 +43,16 @@ function Export-PesterResults {
         }
     }
 }
-function Export-NUnitReport {
+function Export-XmlReport {
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true)]
         $PesterState,
 
         [parameter(Mandatory = $true)]
-        [String]$Path
+        [String]$Path,
+
+        [parameter(Mandatory = $true)]
+        [string] $Format
     )
 
     #the xmlwriter create method can resolve relatives paths by itself. but its current directory might
@@ -69,7 +72,15 @@ function Export-NUnitReport {
         $xmlFile = [IO.File]::Create($Path)
         $xmlWriter = [Xml.XmlWriter]::Create($xmlFile, $settings)
 
-        Write-NUnitReport -XmlWriter $xmlWriter -PesterState $PesterState
+        switch ($Format) {
+            'NUnitXml' {
+                Write-NUnitReport -XmlWriter $xmlWriter -PesterState $PesterState
+            }
+
+            'JUnitXml' {
+                Write-JUnitReport -XmlWriter $xmlWriter -PesterState $PesterState
+            }
+        }
 
         $xmlWriter.Flush()
         $xmlFile.Flush()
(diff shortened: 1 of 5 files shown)


bgelens:
@nohwnd The reports generated with the current code changes are correctly interpreted in GitLab (I've seen AzDevOps also supports JUnit so I'm going to test there a bit as well).

Could you do an initial review pass and tell me what you are not happy with :) I've copied the NUnit routine as much as possible.

Functions/TestResults.ps1
```diff
@@ -33,23 +33,26 @@ function Export-PesterResults {
...
+function Export-XmlReport {
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true)]
         $PesterState,
 
         [parameter(Mandatory = $true)]
-        [String]$Path
+        [String]$Path,
+
+        [parameter(Mandatory = $true)]
```
nohwnd: Would be nice to use `ValidateRange` with the supported formats.

Functions/TestResults.ps1
```diff
@@ -146,7 +157,7 @@ function Write-NUnitEnvironmentInformation($PesterState, [System.Xml.XmlWriter]
     $XmlWriter.WriteStartElement('environment')
 
     $environment = Get-RunTimeEnvironment
-    foreach ($keyValuePair in $environment.GetEnumerator()) {
+    foreach ($keyValuePair in $environment.GetEnumerator() | Where-Object -FilterScript {$_.Name -ne 'junit-version'}) {
```
nohwnd: This might fail when user mocks `Where-Object`. It's not a big issue here, because at this point all Mocks should be removed, but pipeline is slow, and keeping track of where we can and cannot use cmdlets is difficult, so new code should rather not use cmdlets :) use `if  {$_.Name -eq 'junit-version'} { continue }` instead please. If there is no other option and you need to use cmdlet please use it via SafeCommands hashtable.

bgelens:
I've addressed your comments.

I think what is needed now before we consider to move this forward is:

* [ ] Add tests
* [ ] Test on other CI systems than GitLab
  * [ ] Azure DevOps
  * [ ] ...?

Functions/TestResults.ps1
```diff
@@ -222,6 +238,131 @@ function Write-NUnitTestSuiteElements($Node, [System.Xml.XmlWriter] $XmlWriter,
...
+    $XmlWriter.WriteAttributeString('disabled', $PesterState.PendingCount + $PesterState.InconclusiveCount + $PesterState.SkippedCount)
+    $XmlWriter.WriteAttributeString('time', ($PesterState.Time.TotalSeconds.ToString('0.000', [System.Globalization.CultureInfo]::InvariantCulture)))
+}
+
+function Write-JUnitTestSuiteElements($Node, [System.Xml.XmlWriter] $XmlWriter, [string] $Package, [uint16] $Id) {
+    $XmlWriter.WriteStartElement('testsuite')
+
+    Write-JUnitTestSuiteAttributes -Action $Node -XmlWriter $XmlWriter -Package $Package -Id $Id
+
+    if ($Node.Actions.Type -eq 'TestCase') {
```
renehernandez: You could refactor this as follows:

```
$actions = $Node.Actions.Actions

if ($Node.Actions.Type -eq 'TestCase') {
    $actions = $Node.Actions
}

foreach ($t in $actions) {
    Write-JUnitTestCaseElements -Action $t -XmlWriter $XmlWriter
}
```

That way you reduce code duplication

Functions/TestResults.ps1
```diff
@@ -222,6 +238,131 @@ function Write-NUnitTestSuiteElements($Node, [System.Xml.XmlWriter] $XmlWriter,
...
+        Failed {
+            'failure'
+        }
+
+        default {
+            'skipped'
+        }
+    }
+
+    $XmlWriter.WriteAttributeString('status', $Action.Result)
```
renehernandez: Should not this use `$status` instead of `$Action.Result`?

bgelens: I've updated the code a bit to make is more clear what it is used for. 

I'm using the original status for the attribute as its value is not predefined in the schema (keeping terms like "Inconclusive" as part of the junit report). Next, a skipped or failure type adds extra elements which value is predefined in the schema and widgets like the one from GitLab explicitly search for)

bgelens:
Azure Pipelines result:

image

image

I think it is looking good :)

bgelens:
Adding the test file name to class= will make Azure DevOps aware (filter allows to check the box of the .tests.ps1 file) of where the test originated from so I'll update this. I'll also will create the same testname as with NUnitXml `Describe.Context.TestName`

renehernandez:
@bgelens It looks good

bgelens:
Generating junit based on these tests: [simple.tests.ps1.txt](https://github.com/pester/Pester/files/3512290/simple.tests.ps1.txt)

Result in this xml: [junit.xml.txt](https://github.com/pester/Pester/files/3512291/junit.xml.txt)

In Azure DevOps this now looks like:

image

Where the name of the test is the path Describe.Context.It or Describe.It / Context.It or It

I think it is good enough unless someone with actual JUnit experience has some comments :)

I'll work on some tests in the next few days and will remove the draft once that is done

bgelens:
@nohwnd you offered to help with the tests right :)

I started adding test by duplicating and adjusting what was already there for nunit.

I apparently don't know enough about PesterState to imagine how I would get PSv2 tests to pass. I get that a property is missing but afaict, `Enter-TestGroup` calls `New-TestGroup` which should add the `TotalCount` property.

I need to imagine for PSv2 as I'm working on Mac with PSv7.

Can you take a look? Help out :)

bgelens:
hey @nohwnd, Any idea why the CI is not triggered?

bgelens:
Turns out I had a merge conflict as master was changed but this wasn't obviously presented.
I rebased and apparently still need to fix the psv2 issues :|

bgelens:
I think I fixed the issues by handling things a little bit differently. I'm going to do some validation with reports. If it looks ok, I'm going to remove the draft status

bgelens:
I've validated the results to be correct in both GitLab and Azure DevOps. Ready for review!

nohwnd:
@bgelens I know I said I will help with the tests, but I did not and I am sorry. But on the other hand you did it all by yourself without anyone "holding your hand".  🥳🕺🍰 It's merged and will be released soon :)

PR #1353, opened 2019-08-21
Add support for logging the reason if a test is skipped or marked pending

JamesWTruher:
If you create a nunit log, the reason for skipping or a pending test is not persisted, but the schema seems to allow this, and we would like to keep the information in our logs.

The nunit schema allows for a reason to be associated with a test which is skipped or marked pending (or inconclusive). While the current behavior provides a way to emit a reason for skipping or pending, that is only written to the console and not persisted. This PR ensures that those reasons (if provided) will make it into the nunit log.


nohwnd:
Merged thanks. Will try to do release till the end of the week. :)

PR #1369, opened 2019-09-19
WIP Build module

### LINKED ISSUE #1365: Reduce Pester's module import time by adding a build process to the module that merges the dot sourced functions into Pester.psm1
Whilst Pester is great, it suffers from a very long import time (around 3 seconds on a modern 4 core i5 laptop), even using the latest version of PowerShell 7.
This makes it very painful when (re-)running just a few tests locally in a new PowerShell process (which many test frameworks do in order to start from a clean state) or just when running it the first time in a new Editor session.
A common solution to this is to add a build process to the module that essentially merges all the dependent (dot sourced) `.ps1` files into the main `.psm1` file as PowerShell can process one large file more efficiently. I've done a PoC where I've just merged all the `.ps1` files inside the `Functions` folder into it and could halve the import time. I've seen there is a bit of complexity  (like e.g. how to handle the `Gherkin.ps1xml`) but I'm sure I could work it out.

Would you be happy to accept a PR that effectively merges all the `.ps1` (excluding .`tests.ps1`) from the `Functions` and `Dependencies` folder (for the latter, I'll have to first merge the individual ps1 files into the main psm1) into Pester.psm1 for the purpose of deployment and creates a local folder that could be used to publish Pester then (i.e. all files will still stay where they are in source control, it's just a pre-publish step and I'll try to work with you that signing still works). I'd appreciate some help should I find that some tests are broken with this build process to investigate and potentially fix the problem in source or test code.

For the CI build: I would first run the tests as it is now and then have a 2nd task where the module is build and tests are re-run again for a true end to end testing experience.

I already have a working prototype in my branch here and there are only 2-4 test failures when using the built module in CI, some help with investigating those tests would be appreciated, somehow the number of error traces went up to 11, which is the cause of the test failures.
https://github.com/bergmeister/Pester/tree/buildModule
https://ci.appveyor.com/project/bergmeister/pester/builds/27506705

bergmeister:
Fixes #1365

Adds a `build.ps1` file to merge all dot sourced dependencies into `Pester.psm1` for faster import speed.

Outstanding work:
- Investigate failing tests: On PowerShell 5/6 there are 2 failing tests due to the error stack trace suddenly being 11 lines long somehow, on PowerShell 3/4 there are 4 failures in total
- PowerShell 2: The` Invoke-Pester` cmdlet does not get exported
- Instead of merging into the `Pester.psm1` in source control, create everything in a new folder
- Add some unit tests for `build.ps1` to test the internal functions

Question: What about the Travis? I cannot update their build to see if there would be breaking changes there? The PS v2 build is broken because there is no `-Raw` parameter on `Get-Content`, I really do not understand the need to support deprecated technology.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,0 +1,16 @@
+$PowerShellScriptsToBeMerged = Get-ChildItem -Path (Join-Path $PSScriptRoot Functions) -Filter '*.ps1' -Exclude '*.Tests.ps1' -Recurse
+$mergedContent = ''
+foreach ($powerShellScriptToBeMerged in $PowerShellScriptsToBeMerged) {
+    $scriptContentToBeMerged = Get-Content -Path $powerShellScriptToBeMerged.FullName -Raw
+    $mergedContent += @"
+#region $($powerShellScriptToBeMerged.BaseName)
+$scriptContentToBeMerged
+#endregion
+
+"@
+}
+
+$mainModulePath = Join-Path $PSScriptRoot 'Pester.psm1'
+$contentOfMainModule = Get-Content -Path $mainModulePath
+$contentOfMainModule = $contentOfMainModule.Replace('# Functions', $mergedContent)
+Set-Content -Path $mainModulePath -Value $contentOfMainModule
--- a/build.ps1
+++ b/build.ps1
@@ -1,4 +1,4 @@
-$PowerShellScriptsToBeMerged = Get-ChildItem -Path (Join-Path $PSScriptRoot Functions) -Filter '*.ps1' -Exclude '*.Tests.ps1' -Recurse
+$PowerShellScriptsToBeMerged = Get-ChildItem -Path (Join-Path $PSScriptRoot Functions) -Filter '*.ps1' -Exclude '*.Tests.ps1', '*.ps1xml' -Recurse
 $mergedContent = ''
 foreach ($powerShellScriptToBeMerged in $PowerShellScriptsToBeMerged) {
     $scriptContentToBeMerged = Get-Content -Path $powerShellScriptToBeMerged.FullName -Raw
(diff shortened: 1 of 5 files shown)


nohwnd:
Travis - I don't know we will have to believe :) 
PowerShell 2 - unfortunately PowerShel 2 should stay supported until Pester 4 starts being in maintenance mode. How about `System.IO.File.ReadAllText` .NET method?

build.ps1
```diff
@@ -0,0 +1,82 @@
...
+    #endregion
+
+"@)
+        }
+    }
+    $mainModuleContent
+}
+
+$dependenciesFolder = Join-Path $PSScriptRoot 'Dependencies'
+$axiomsCode = Get-ContentOfMergedFolder -DirectoryPath (Join-Path $dependenciesFolder 'Axiom')
```
nohwnd: Axioms should be just for tests, it should not be needed to merge them in Pester code.

build.ps1
```diff
@@ -0,0 +1,82 @@
...
+$typeClassImport = 'Import-Module $PSScriptRoot\..\TypeClass\TypeClass.psm1 -DisableNameChecking'
+if (-not $formatCode.Contains($typeClassImport)) {
+    throw "Expected the following string to be present in Format.psm1 for replacement '$typeClassImport'"
+}
+$formatCode = $formatCode.Replace($typeClassImport, '')
+
+$dependenciesCode = "$axiomsCode$([System.Environment]::NewLine)$formatCode$([System.Environment]::NewLine)$typeClassCode"
+Edit-Region -Path $mainModulePath  -RegionName 'Dependencies' -ReplacementText $dependenciesCode
+
+# pwsh -c 'measure-command { ipmo C:\Users\christoph.bergmeiste\git\Pester\Pester.psd1 }'
```
nohwnd: This should be deleted before merge.

nohwnd:
The failures for the stack output should go back to this line where the "full error" variable is set to `$false` (default value) to prevent the full stack to be printed. 

https://github.com/bergmeister/Pester/blob/buildModule/Functions/Output.Tests.ps1#L197

The thing is that it does not seem to resolve to it correctly and so the longer stack trace is produced. Just a guess though.

bergmeister:
Yes, you're right, I must've not seen the `.travis.yml`, I've made the change here and only the 2 known test failures happen there.  What is a bit odd is that the tests take longer on some PS version (3 and 6). I've managed to get the `build.ps1` 'working' now for PS v2 but the module import later does not export the `Invoke-Pester` cmdlet somehow, that needs more debugging, I suppose the build script does not do what it is supposed to do on this version, will need to investigate more and potentially upload the merged module of that build so see what is going on.
Thanks for the first investigation about why the stack traces might have increased, I will have a look at it a bit later.

nohwnd:
Forgot that this is here, and did it myself. Sorry.

Issue #292, opened 2015-03-17
Add "Context" / "It" level granularity to -TestName parameter


vors:
Current granularity allows only `Describe` block level granularity.

One important dev scenario: 

There is a pester suite with one `Describe` block, that runs long time (minutes). I have a failure in one `It`. I want to be able modify my production code, run / debug test (without modifying test code) as fast as possible (seconds). I'm only interested in one `It`.

Currently, I have to modify test code: comment all but one tests.

vors:
Proposed syntax: `Invoke-Pester -TestName "DescribeBlockName[\ContextBlockName[\ItBlockName]]"`

nohwnd:
@vors That is an interesting idea, I also miss it sometimes myself. 
I have some questions:
1) Describe and context names are usually pretty verbose, copy pasting them in console is a bit clumsy. Is there another syntax you considered?
2) I suppose the before each, after each and other setup code, including all the code placed in describes should run. How would you implement this? As a simple `if` on the names?
3) should wildcards be supported?

dlwyatt:
We already support wildcards in the Describe portion, so may as well keep that behavior.  Then you could do something like `-TestName '*PartialDescribeName*\*PartialContextName*\*Whatever*'`.  We'd just split the string on the backslashes and take it from there.

It would be nice to be able to just specify the '*Whatever*' part and have Pester figure out that there's a deep nested test that needs to be run, but since we're executing the code as it's being read (rather than parsing ahead of time), that's not really an option at the moment.

nohwnd:
Should `\\*correcly*` run all the tests that include "correctly" in its name?

dlwyatt:
That, or `'*\*\*correctly*'`.  However, I see that as being of pretty limited use right now.  Most of the time spent is out in the Describe and Context blocks anyway, and they'd still be executed, even if there were no matching It blocks inside them.

vors:
@nohwnd 
1) I think it's fine: at least way better then existing alternative.
2) yes, this is exactly what I had in mind: simple `if` on name.
3) Dave already covered that, I personally don't see value in wildcards, except `'*\*\*correctly*'`

One think that I missed: what to do if `Describe` block name has `\` inside. I think its' a corner case and should not stop us from implementing, but I don't have a good answer from the top of my head.

nohwnd:
[...]

Issue #721, opened 2017-01-31
Comments needed: Allowing a single test to run


nohwnd:
As mentioned in #657 and possibly quite a few other places, Pester test case filtering is limited only to tags on top-level describes. I would like to see this to change to allow for running just a single test case, which would allow us to:
- run just the current test
- run just failed tests
- run just tests that have a given tag (trait)
- run just tests from selected scope, even child scopes

As in most other frameworks.

To make this work we would need to change the way, Pester works a bit, but more importantly we would need to change the way test are written. 

To allow running any code on demand we would need to control all of it's sorroundings. No code would be allowed outside of pester controlled blocks, and dot sourcing would be replaced with a custom function that would dot source on demand. In a sense `It` would become a first class citizen. Pretty much as a method is first class citizen in any C# testing framework. Limiting where code is placed is quite hard, but we possibly do not need a strong limit, just a set of recommendations.

I went through my tests and most of them already follow this schema, with occasional guard mock here and there, or some other common setup for multiple test cases, in general the tests look something like this: 

```powershell
$here = Split-Path -Parent $MyInvocation.MyCommand.Path
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.'
. "$here\$sut"

Describe "Remove-InactiveADUser" {
    #guard mock to avoid calling the real command by accident
    Mock Remove-ADUser

    It "Calls Remove-ADUser with correct parameters when user has been inactive for more than a year" {
        # -- Arrange
        Mock Get-ADUser { Name = "Jakub"; LastLogon = [datetime]::Now.AddMonths(-13) }
        Mock Remove-ADUser { } -ParameterFilter { $Name = "Jakub" }

        # -- Act
        Get-ADUser | Remove-InactiveADUser

        # -- Assert
        Assert-MockCalled Remove-ADUser { } -ParameterFilter { $Name = "Jakub" } -Exactly 1
    }
}
```

Which would transform to something like this: 

```powershell
Add-Dependency Remove-InactiveADUser.ps1

Describe "Remove-InactiveADUser" {
    BeforeAll {
        #guard mock to avoid calling the real command by accident
        Mock Remove-ADUser
    }

    It "Calls Remove-ADUser with correct parameters when user has been inactive for more than a year" {
        # -- Arrange
        Mock Get-ADUser { Name = "Jakub"; LastLogon = [datetime]::Now.AddMonths(-13) }
[...]

PR #1374, opened 2019-09-24
Filter what Context and It blocks run

splatteredbits:
Pester wants test fixtures to have one (or just a few) `Describe` blocks with many `It` blocks representing specific tests. Unfortunately, the `TestName` parameter only applies to `Describe` blocks. If you have all your `It` blocks in a single `Describe` (as most everybody and Pester examples do), then you can't run individual tests. If these `It` blocks take a long time and one of them is failing, it can be very painful to debug them. You can comment out the tests you don't want to run, or re-arrange, or add `return` statements. Or, you use a pattern where you have a single `It` block in each `Describe` block.

As a developer, I'd like parameters on `Invoke-Pester` that allow me to filter `It` blocks so I don't have to go through all the steps I currently have to just to run a single `It` block.

This PR adds `ContextFilter` and `ItFilter` parameters to `Invoke-Pester` that allows this level of control.

This should address issue #292, although without the special formatting to the `TestName` parameter described in that issue.

I would also like to rename the `TestName` parameter to `DescribeFilter` (with a backwards-compatible alias) along with renaming internal parameters and Pester state property names. That's a big refactor and didn't want to do it without permission.

Also, maybe in Pester 5, we can make `ItFilter` the second  positional parameter instead of `TestName` because it just makes more sense to filter `It` blocks my default than `Describe` blocks.

### DIFF
--- a/Functions/Context.Tests.ps1
+++ b/Functions/Context.Tests.ps1
@@ -20,3 +20,61 @@ Describe 'Testing Context' {
         }| should -Throw  'No test fixture is provided. (Have you put the open curly brace on the next line?)'
     }
 }
+
+Describe 'Filtering Context' {
+    It 'should only run filtered contexts' {
+        InModuleScope 'Pester' {
+            Mock -CommandName 'DescribeImpl' -ModuleName 'Pester' -ParameterFilter { $CommandUsed -eq 'Context' }
+
+            $currentFilter = $Pester.ContextFilter
+            $Pester.ContextFilter = @( 'one', 'tw*' )
+            try
+            {
+                Context 'One' {
+                }
+
+                Context 'Two' {
+                }
+
+                Context 'Three' {
+                }
+
+                Assert-MockCalled -CommandName 'DescribeImpl' -ModuleName 'Pester' -Times 2 -Exactly
+                Assert-MockCalled -CommandName 'DescribeImpl' -ModuleName 'Pester' -Times 1 -ParameterFilter { $Name -eq 'One' }
+                Assert-MockCalled -CommandName 'DescribeImpl' -ModuleName 'Pester' -Times 1 -ParameterFilter { $Name -eq 'Two' }
+                Assert-MockCalled -CommandName 'DescribeImpl' -ModuleName 'Pester' -Times 0 -ParameterFilter { $Name -eq 'Three' }
+            }
+            finally
+            {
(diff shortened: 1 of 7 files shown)


nohwnd:
Thanks for this PR. I see that my comment from last week was not posted. Not sure what I did wrong. 

This can theoretically work, but there are a lot of limitations and edge cases, and things that the user needs to understand for this implementation to work in his favor. I was achieving this in v5 and it took a lot of figuring out to provide a good solution that is easy to understand and has less edge cases. 

I'd like to finish v5 soon and release it and I don't think it makes sense to merge this into v4 as it would add parameters that don't have behavior that is easy to understand and require a lot from the user to actually do what they are supposed to do. 

I am not sure if I am describing this in a way that can be understood, so please ask questions.

splatteredbits:
I'm curious to know more about the limitations and edge cases. It seems pretty straight-forward to me. 

Getting this functionality into Pester is critical for our team. 

Do you have documentation on how the v5 functionality works?

Also, I disagree that these parameters are not easy to understand. `ItFilter` seems pretty obvious. 

I could be convinced to drop the `ContextFilter`. Contexts behave weirdly.

nohwnd:
> I'm curious to know more about the limitations and edge cases. It seems pretty straight-forward to me.

The filter promises to run just the single test, but it will run the test and every setup and teardown in every file that is included in the run. It is the same situation for Describe/Context. To enable the parameter to work correctly the test author would need to put all code in Pester controlled blocks so the execution of the code can be postponed till appropriate time, or skipped entirely (as in v5). 

The thread you recently commented in [sums up](https://github.com/pester/Pester/issues/721) my thinking about this in a more complete manner I hope. 

Another limitation is that once you add anything dynamic to the name of the test you won't be able to filter it anymore because the strings won't match and you will have to resort to using wildcards. A better way to solve this is for example using the position of the opening curly brace of the test because that won't change during runtime. 

Pretty much I am reluctant to merge this because the promises the parameter names give are hard to keep, the limitations of the functionality is hard to sum up, and I am pretty sure that when v4 will be put to maintenance mode those parameters would be a source of confusion for the years to come. ☹️

ErikUmhoe:
Any plans to get this merged in?

nohwnd:
Not merging this because of the reasons outlined above. Thanks for posting the PR.

PR #1382, opened 2019-10-07
Handle when exceptions have no error messages.

rasmusjelsgaard:
Pester crashes on exceptions with null error messages. This PR fixes the problem by adding an additional check in the code that "unwinds" the stack trace and error messages.

I do not believe there is an open issue for this problem on github.


nohwnd:
Thanks for the PR. Looks good to me. Are you done or still work in progress?:)

rasmusjelsgaard:
It is done. Thanks for the review.

Issue #1157, opened 2018-12-12
Document array comparing capabilities of Should


nohwnd:
As mentioned in #804 there is no documentation about Should and how it compares arrays. Add it or wait for Should refactoring to add it.

joshcorr:
Would be happy to add this documentation, if it hasn't been addressed already. Does it need to go into `about_Should.help.txt`

nohwnd:
Yes there, and wiki. 🙂

PR #1396, opened 2019-12-07
Add an Example of Should Be with an Array

joshcorr:

Adding documentation about existing support for array comparison in the `-Be` operator.
This PR fixes the `about_Should.help.txt` as requested in #1157. 
This also includes a pester test of the specific example added, showing that arrays with similar values will not pass assertion if the values are in different indexes .

If this PR is accepted the [corresponding Wiki page](https://github.com/pester/Pester/wiki/Should#be) will be updated for this change to the Help.

### DIFF
--- a/Functions/Assertions/Be.Tests.ps1
+++ b/Functions/Assertions/Be.Tests.ps1
@@ -36,6 +36,15 @@ InModuleScope Pester {
             $array | Should -EQ $arrayWithCaps
         }
 
+        It 'Compares arrays with similar values in different order' {
+            [int32[]]$array = (1..10)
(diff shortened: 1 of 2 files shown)


joshcorr:
Looks like a test failed, but I don't think it is related to the code that I included. Please let me know if anything needs to change or I need to kick off another run of the pipeline.

nohwnd:
Merged. There was probably some update on the AppVeyor host or something, because I've seen this exact test fail in other PRs as well. Thank you for your contribution. 🙂

joshcorr:
Awesome! Happy to contribute finally to a module I use all the time. I went ahead and updated the wiki with the help text merged in this PR.

PR #1400, opened 2019-12-18
Fixed incorrect variable used for hooks with tags in Invoke-GherkinHook.

jtomkiew-mng:

There was an incorrect variable used in `Invoke-GherkinHook` function (in `Gherkin.ps1`) for accessing hook's script - instead of an actual hook object (`$GherkinHook`), a hook's name was given (`$hook`).

This only happened for hooks with tags.

This has now been fixed and a unit test was added.

I don't believe there was any issue reported for this.


nohwnd:
Thanks, merged!

PR #1421, opened 2020-01-11
Throw on should failure in mock parameter filter

### LINKED ISSUE #1409: Collect Should errors in Mock
When running in the context of Mock we should collect errors and fail the MockWith or ParameterFilter, instead of writing the errors to the parent It block.

nohwnd:
Forces should to fail on assertion failure when in mock param filter irrespective of Should preference to keep scenario that throws in ParameterFilters working.

Partial fix #1409


nohwnd:
Fixes #1409

nohwnd:
Close reopen to trigger new CI build

PR #1426, opened 2020-01-17
Handle the case when the failure message contains an escape sequence

JamesWTruher:
…which may occur with PowerShell 7

If the failure message contains an escape sequence (which can happen with PowerShell 7, and the results are going to be serialized to an Nunit file, you may see something like this:
```
MethodInvocationException: /Users/jimtru/src/github/forks/JamesWTruher/Pester/Functions/TestResults.ps1
Line |
 591 |             $xmlWriter.WriteElementString('message', $TestResult.FailureMessage)
     |             ^ Exception calling "WriteElementString" with "2" argument(s): "', hexadecimal value 0x1B, is an invalid character."

```

and the xml file created is partial. The escape in the message is not valid for serialization

This escapes the escape character ([char]27) to `&27;`

### DIFF
--- a/Functions/TestResults.Tests.ps1
+++ b/Functions/TestResults.Tests.ps1
@@ -45,7 +45,7 @@ InModuleScope Pester {
                 $TestResults = New-PesterState -Path TestDrive:\
                 $testResults.EnterTestGroup('Mocked Describe', 'Describe')
                 $time = [TimeSpan]25000000 #2.5 seconds
-                $TestResults.AddTestResult("Failed testcase", 'Failed', $time, "Assert failed: ""Expected: `e[34mTest`e[0m. But was: Testing""", 'at line: 28 in  C:\Pester\Result.Tests.ps1')
+                $TestResults.AddTestResult("Failed testcase", 'Failed', $time, "Assert failed: ""Expected: ``[34mTest``[0m. But was: Testing""", 'at line: 28 in  C:\Pester\Result.Tests.ps1')
 
                 #export and validate the file
                 [String]$testFile = "$TestDrive{0}Results{0}Tests.xml" -f [System.IO.Path]::DirectorySeparatorChar
--- a/Functions/TestResults.Tests.ps1
+++ b/Functions/TestResults.Tests.ps1
@@ -40,6 +40,25 @@ InModuleScope Pester {
                 $xmlTestCase.failure.'stack-trace' | Should -Be 'at line: 28 in  C:\Pester\Result.Tests.ps1'
             }
 
+            It "should write a failed test result when the failure message has an escape character" {
+                #create state
+                $TestResults = New-PesterState -Path TestDrive:\
+                $testResults.EnterTestGroup('Mocked Describe', 'Describe')
+                $time = [TimeSpan]25000000 #2.5 seconds
+                $TestResults.AddTestResult("Failed testcase", 'Failed', $time, "Assert failed: ""Expected: `e[34mTest`e[0m. But was: Testing""", 'at line: 28 in  C:\Pester\Result.Tests.ps1')
+
+                #export and validate the file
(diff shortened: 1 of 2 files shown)


Functions/TestResults.ps1
```diff
@@ -588,7 +588,7 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
             $XmlWriter.WriteAttributeString('result', 'Failure')
             $XmlWriter.WriteAttributeString('executed', 'True')
             $XmlWriter.WriteStartElement('failure')
-            $xmlWriter.WriteElementString('message', $TestResult.FailureMessage)
+            $xmlWriter.WriteElementString('message', $TestResult.FailureMessage.Replace("`e",'&27;'))
```
nohwnd: `` `e`` escape sequence is a recent addition to PowerShell, prior version 6 it produces `e` (as escaping any other non-special letter) which will replace incorrectly. Please use a different way of providing the grave accent, for example a non-expandable string `` '`' ``.

nohwnd: I just realized you meant `Escape` as in escape key, not escape sequence character (grave accent `` ` ``),  fixing.

Functions/TestResults.Tests.ps1
```diff
@@ -40,6 +40,26 @@ InModuleScope Pester {
                 $xmlTestCase.failure.'stack-trace' | Should -Be 'at line: 28 in  C:\Pester\Result.Tests.ps1'
             }
 
+            It "should write a failed test result when the failure message has an escape character" {
+                #create state
+                $TestResults = New-PesterState -Path TestDrive:\
+                $testResults.EnterTestGroup('Mocked Describe', 'Describe')
+                $time = [TimeSpan]25000000 #2.5 seconds
+                $escape = [string][char]27 # escape, also `e in PowerShell 6+
```
nohwnd: @JamesWTruher not 100% sure how to repro the error with escapes on my PowerShell 7, please verify the changes. 

The cast to string here is unnecessary, but I kept it anyway to keep it same as in the code file. (There it is needed to force use of `Replace(string, string)` instead of `Replace(char, char)` which fails.

JamesWTruher: i'll grant you that my test is quite artificial, but here's the way to see it:
``` powershell
PS> Get-Content /tmp/t1.tests.ps1
describe "a simple test" {
    it "can handle an error with an escape sequence" {
       throw  "`e[31m < < , > > `e[0m"
    }
}
```
when run as follows, you can see the error:
``` powershell
[Pester|master↑0↓0•0+0?0] PS> invoke-pester /tmp/t1.tests.ps1 -OutputFile /tmp/xx.xml -OutputFormat NUnitXml
    ____            __
   / __ \___  _____/ /____  _____
  / /_/ / _ \/ ___/ __/ _ \/ ___/
 / ____/  __(__  ) /_/  __/ /
/_/    \___/____/\__/\___/_/
Pester v4.9.0
Executing all tests in '/tmp/t1.tests.ps1'

Executing script /tmp/t1.tests.ps1

  Describing a simple test
    [-] can handle an error with an escape sequence 1ms
      RuntimeException:  < < , > > 
      at <ScriptBlock>, /tmp/t1.tests.ps1: line 3
Tests completed in 24ms
Tests Passed: 0, Failed: 1, Skipped: 0, Pending: 0, Inconclusive: 0 
MethodInvocationException: /Users/jimtru/src/github/forks/JamesWTruher/Pester/Functions/TestResults.ps1
Line |
 591 |             $xmlWriter.WriteElementString('message', $TestResult.FailureMessage)
     |             ^ Exception calling "WriteElementString" with "2" argument(s): "', hexadecimal value 0x1B, is an invalid character."
```

JamesWTruher: I probably should have put that in the test, but it seemed more information than might have been needed.

PR #1428, opened 2020-01-19
Remove logo from header because it is noisy

nohwnd:


Greg-Smulko:
In #1343 the logo was added in 3 files, but here it's removed from a single file only. Is it on purpose @nohwnd ?

nohwnd:
@Greg-Smulko  Nope, my bad, will fix :)

PR #1457, opened 2020-02-19
sessionstate variable is not being passed correctly in Context.ps1

halvarsson:
I recently upgraded to the latest version of Pester and while debugging some of my tests were failing with below

New-PesterState : Cannot process argument transformation on parameter 'SessionState'. Cannot convert the "SessionState" value of type "System.String" to type "System.Management.Automation.SessionState".
At C:\Program Files\WindowsPowerShell\Modules\pester\4.10.1\Functions\Context.ps1:88 char:128
 ... '] .) -TestNameFilter $null -TagFilter @() -SessionState SessionState

                                                            
It seems that a bug was introduced with Session state hints (#1169) in the change to Context.ps1 (   https://github.com/pester/Pester/commit/d25ff9f34b1fb160df096cc7101efdd8a06cbd43#diff-d4fc7a10c374cf6c159f443304feb3e4)


nohwnd:
Thanks! I guess this only reproduces when you use Context as top level block and run interactively? That is why no-one noticed for such a long while. 😊

halvarsson:
Thanks for the quick merge :) yep exactly, I was building a context/it block builder function for some data structures when I ran into it

Issue #1129, opened 2018-10-22
Pester Gherkin processor assumes too much when locating Step files


fourpastmidnight:
## 1. General summary of the issue

When the Gherkin feature of Pester attempts to find step definitions for the scenarios in feature files, it makes a bad assumption about the possible location(s) of `*.step.ps1` files containing step definitions.

## 2. Describe Your Environment

Pester version : 4.4.2 C:\Program Files\WindowsPowerShell\Modules\Pester\4.4.2\Pester.psd1
PowerShell version : 5.1.17134.165
OS version : Microsoft Windows NT 10.0.17134.0

## 3. Expected Behavior

In _The Cucumber Book_, on page 79 it talks about how one might go about organizing feature files:

> ### Subfolders
> This is the easiest way to organize your features. You may find yourself torn as to how to choose a catgeory, though....\
> Of course, this is a decision for you and your team to take, but we can offer a little bit of advice. We've had the most success using subfolders to represent different high-level tasks that a user might try to do....
>
>     features/
>         reading_reports/
>         report_building/
>         user_administration/

My module will be made up of many "regions" of functionality--and so my module will contain various folders under the `Public` folder of the module to organize these regions of functionality. So, in the same way, I want to organize my Feature files.

With Cucumber, all step definition files are expected to be located under the `features` folder. _The Cucumber Book_ recommends a folder named `step_definitions`. I simply use `steps`, as do many others. In any event, on page 130, they go on to suggest a method of organizing step definitions:

> ### Organizing Our Step Definitions
> We've managed to get this far with a single file of step-definitions called `steps.rb`, and for a project of this size, we might well stick with a single file for a while longer. As the number of step definitions starts to grow, we'll want to split up the files so that the code is more cohesive. We've found that our favorite way to organize step definition files is to organize them with one file per domain entity. So, in our example, we'd have three files:
>
>     features/step_definitions/account_steps.rb
>     features/step_definitions/teller_steps.rb
>     features/step_definitions/cash_slot_steps.rb

The reason why there's no further folder organization under the `step_definitions` directory is because in Cucumber, all step definitions are global to the test suite. (See the sidebar on page 149 in _The Cucumber Book_.)

So, following along with the previous section, and this section, the file system would be organized as follows:

```
Specs/
    features/
        reading_reports/
            *.feature
        report_building/
            *.feature
        step_definitions/
            account_steps.rb
            teller_steps.rb
            cash_slot_steps.rb
        user_administration/
            *.feature
```

[...]

PR #1458, opened 2020-02-24
Adding Parameter -StepPath to Gherkin.ps1

Mayneg:
This will allow a user to to run step files in path specified by the user. This has has been discussed here (https://github.com/pester/Pester/issues/1129) and i have used the code from user cgnuechtel


nohwnd:
sorry but won't merge, I am sure you've found another way to have those changes since this is years old.

PR #1460, opened 2020-02-29
Add advanced tag filtering options

### LINKED ISSUE #608: Invoke-Pester for tests, where all of numerus tags are present
`Invoke-Pester -Tag @('A', 'B')` runs any tests tagged with "A" or "B".

```powershell
Describe 'foo1' -Tag 'A' {
    It 'should be ok' {
        1 | should be 1
    }
}

Describe 'foo2' -Tag 'B' {
    It 'should be ok' {
        1 | should be 1
    }
}

Describe 'foo3' -Tag @('A', 'B') {
    It 'should fail' {
        1 | should be 1
    }
}

```

For example, here all 3 tests would be executed.

We have a situation, where it's desirable to run tests **when all tags** are present.
So in this case, only "foo3" would be run.

Changing the meaning of `-Tag` is obviously not an option.
I'd like to propose to add a new parameter, maybe `-TagAll` that enable this scenario.

At the same time, we may rename `-Tag` into `-TagAny` and add alias `-Tag` to this parameter.
Also, maybe aliasing `-ExcludeTag` with `-TagExclude` will make exportability of all these 3 related parameters better.

For the context, I'm working on https://github.com/PowerShell/PowerShell/pull/2280 where I'm adding a new tag "RequireAdminOnWindows" and this tag is orthogonal to our current used tags.

hbuckle:

Adds a new parameter AdvancedTag to Describe to specify tags as a hashtable and a corresponding TagFilter parameter to Invoke-Pester that allows for granular filtering of tags using a scriptblock

Fix #608 

### TODO
- [ ] Documentation

### DIFF
--- a/Functions/Describe.Tests.ps1
+++ b/Functions/Describe.Tests.ps1
@@ -207,6 +207,28 @@ InModuleScope Pester {
             }
         }
 
-        # Testing nested Describe is probably not necessary here; that's covered by PesterState.Tests.ps1 and $pester.EnterDescribe().
+        Context 'Advanced Tags Filter' {
+            $testBlock = { MockMe }
+            $cases = @(
+                @{ AdvancedFilter = { $tags.key1 -eq "value1" }; AdvancedTag = @{ key1 = "value1"; key2 = "value2" }; Runs = 1; Because = 'key1 == value1' }
+                @{ AdvancedFilter = { $tags.key1 -eq "value1" -and $tags.key2 -eq "value2" }; AdvancedTag = @{ key1 = "value1"; key2 = "value2" }; Runs = 1; Because = 'key1 == value1 AND key2 == value2' }
+                @{ AdvancedFilter = { $tags.key1 -ne "nothing" -and $tags.key2 -eq "value2" }; AdvancedTag = @{ key1 = "value1"; key2 = "value2" }; Runs = 1; Because = 'key1 != nothing AND key2 == value2' }
+                @{ AdvancedFilter = { $tags.key1 -ne "value1" }; AdvancedTag = @{ key1 = "value1"; key2 = "value2" }; Runs = 0; Because = 'key1 != value1' }
+            )
+            foreach ($case in $cases) {
+                $testState = New-PesterState -Path $TestDrive -AdvancedTagFilter $case.AdvancedFilter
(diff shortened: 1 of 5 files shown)


hbuckle:
I can't seem to get the PowerShell 2 run to pass and I've no way of testing it locally, any ideas?

nohwnd:
@hbuckle thanks for the PR but I won't be merging this into v4 I am sorry. The tagging in v4 barely works for the basic cases, and adding yet another concept to tagging is confusing, and would be hard for me to support going forward. 

In Pester v5 the tagging works much better, and while it is not as flexible as what you provided here right now, it can be easily extended to do more, by allowing the filter to take a scriptblock. 

I am definitely interested in why you are using  hashtable of tags together with a scriptblock. Why tags are not an array of strings? 🙂  What advantage it brings? Maybe there is some idea that I never considered. 🙂

hbuckle:
@nohwnd thanks for the feedback - hashtable and scriptblock seemed much more natural for writing filters than a string array. For example we have tests that need to run at different points during a release and for different environments. So a simplified example could look like

```powershell
Describe "Linting" { } -Tag @{
  when         = "before_plan"
  environments = @("ci")
}

Describe "Policy" { } -Tag @{
  when         = "before_apply"
  environments = @("ci", "staging", "production")
}

Describe "Integration" { } -Tag @{
  when         = "after_apply"
  environments = @("staging")
}

Describe "Smoke" { } -Tag @{
  when         = "after_apply"
  environments = @("production")
}

Invoke-Pester -Filter {
  $tags.when -eq "before_plan" -and
  $tags.environment -contains "$(Release.EnvironmentName)"
}

Invoke-Pester -Filter {
  $tags.when -eq "before_apply" -and
  $tags.environment -contains "$(Release.EnvironmentName)"
}
```

nohwnd:
That definitely looks interesting, I saw other requests to be able to add more metadata to a test. I think this could be merged with that and added to v5. I would just need to think a bit more about how it works in relation to filtering, testcases, addinge metadata to parent blocks, and being able to print the info to the screen. 🙂

nohwnd:
Closing this for the reasons above,

nohwnd:
Need to look into this if it can be retargetted for v5, or if the functionality needs to be written from scratch there.

nohwnd:
we will need to make this from scratch. thanks for the code  :)

PR #1463, opened 2020-03-09
Added parameter for Alias in HaveParameter

lipkau:

Added `-Alias` to `| Should -HaveParameter`

This will check if the list of aliases of the parameter contain the string the test is trying to assert

### DIFF
--- a/Functions/Assertions/HaveParameter.Tests.ps1
+++ b/Functions/Assertions/HaveParameter.Tests.ps1
@@ -6,6 +6,7 @@ InModuleScope Pester {
         function Invoke-DummyFunction {
             param(
                 [Parameter(Mandatory = $true)]
+                [Alias('First')]
                 $MandatoryParam,
 
                 [ValidateNotNullOrEmpty()]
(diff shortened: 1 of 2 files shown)


nohwnd:
Is the formatting here automatic? It has space on the end, but not on the start.

lipkau:
yes, it is automatic, @nohwnd .
I normally rely on the project `.vscode/settings.json` to define the formatting style the vscode powershelle extension should use.
Which parts of the formatting should I change? can you comment the lines?

Functions/Assertions/HaveParameter.Tests.ps1
```diff
@@ -85,30 +87,30 @@ InModuleScope Pester {
     Describe "Should -HaveParameter" {
 
         It "passes if the parameter <ParameterName> exists" -TestCases @(
-            @{ParameterName = "MandatoryParam"}
-            @{ParameterName = "ParamWithNotNullOrEmptyValidation"}
-            @{ParameterName = "ParamWithScriptValidation"}
+            @{ParameterName = "MandatoryParam" }
```
nohwnd, line 90: I was talking about this, but the formatter does not add the space on the start line. And the many cases below.

nohwnd: But running the automatic formatter on my side adds a lot of undesired changes. Maybe it is because I am running on preview IDK. Merging as is.

PR #1470, opened 2020-03-28
Clean result object

nohwnd:


nohwnd:
Tests passed, but the upload failed due to 500. Merging.

PR #1494, opened 2020-04-13
Fixes #1028 - Code Coverage for proxy functions/steppablePipelines

### LINKED ISSUE #1028: Code Coverage for proxy functions/steppablePipelines

In proxy functions, a steppable pipeline is created from a scriptblock:
```
$scriptCmd = {& $wrappedCmd @PSBoundParameters }
$steppablePipeline = $scriptCmd.GetSteppablePipeline($myInvocation.CommandOrigin)
```

Each pipeline phase (begin/process/end) is then called on this object, thus actually executing the scriptblock.

However, when executing Pester with the CodeCoverage option, the scriptblock is reported as not being executed:
```
Missed commands:

File        Function      Line Command
----        --------      ---- -------
Module.psm1 Select-String   71 $scriptCmd = {& $wrappedCmd @PSBoundParameters }
Module.psm1 Select-String   71 & $wrappedCmd @PSBoundParameters
```

```powershell
Pester version     : 3.4.0 C:\Program Files\WindowsPowerShell\Modules\pester\3.4.0\Pester.psm1
PowerShell version : 5.1.16299.251
```

## 3. Expected Behavior

Pester should check the pipeline is executed (`$steppablePipeline`'s Begin, Process and End methods should be called) and if so, mark the code in the scriptblock as executed.

## 4. Current Behavior

Pester doesn't recognise steppable pipelines execution and marks the code as not having been executed.

## 5. Possible Solution

Not sure how code coverage is implemented in Pester, so I'll not venture into a possible solution and its corner cases.

## 6. Context

This issue is for statistic purpose only, having two lines by function not being reported as covered isn't a big deal, but that's still erroneous-ish.

leojackson:
Pester generates all breakpoints using a combination of line number and column number. PowerShell does not trigger breakpoints on the proxy function constructs described in #1028 when initiated using both line and column number, but it does when using line number only. This appears to be a bug within PowerShell itself, as the behavior can be replicated outside of Pester like so:
```powershell
# Create a new proxy function for Write-Output in a file in the current directory
$ProxyMetadata = New-Object System.Management.Automation.CommandMetaData (Get-Command Write-Output)
@"
Function Write-FancyTestOutput {
$([System.Management.Automation.ProxyCommand]::Create($ProxyMetadata))
}
"@ | Out-File $PWD\Write-FancyTestOutput.ps1

# Get AST for the new function
$ast = [System.Management.Automation.Language.Parser]::ParseFile( `
    "$PWD\Write-FancyTestOutput.ps1", `
    [ref]$null, `
    [ref]$null `
)

# Initialize line number var
$LineNumber = 0

# Use FindAll to loop through all commands, look for the
# '& $wrappedCmd @PSBoundParameters' command, and add a
# breakpoint at line & column for that command -- this is
# what Pester currently does
$ast.FindAll( `
    {$args[0] -is [System.Management.Automation.Language.CommandBaseAst]}, `
    $true `
) | Where-Object {
    $_.Extent.Text -like '*& $wrappedCmd @PSBoundParameters*'
} | ForEach-Object {
    If($LineNumber -le 0) {
        $LineNumber = $_.Extent.StartLineNumber
    }
    Set-PSBreakpoint -Script "$PWD\Write-FancyTestOutput.ps1" `
                     -Action { } `
                     -Line $_.Extent.StartLineNumber `
                     -Column $_.Extent.StartColumnNumber
}

. $PWD\Write-FancyTestOutput.ps1
Write-FancyTestOutput -InputObject "Hello"   # will not trigger any breakpoints

# Add another breakpoint with no column specified for the same line
If($LineNumber -gt 0) {
    Set-PSBreakpoint -Script "$PWD\Write-FancyTestOutput.ps1" `
                     -Action { } `
                     -Line $LineNumber
}

Write-FancyTestOutput -InputObject "Hello again"   # Triggers the last breakpoint
Get-PSBreakpoint | fl
```
...which results in this output:
```text
Id       : 0
Script   : C:\Temp\Write-FancyTestOutput.ps1
Line     : 22
Column   : 22
Enabled  : True
HitCount : 0
Action   :

Id       : 1
Script   : C:\Temp\Write-FancyTestOutput.ps1
Line     : 22
Column   : 23
Enabled  : True
HitCount : 0
Action   :

Id       : 2
Script   : C:\Temp\Write-FancyTestOutput.ps1
Line     : 22
Column   : 0
Enabled  : True
HitCount : 1
Action   :
```
Rather than trying to rewrite Pester to create breakpoints without column specifications, this PR includes proxy functions using steppable pipelines in the `IsIgnoredCommand` private function. I feel this is appropriate because in proxy command situations, the steppable pipeline constituents are already covered by breakpoints.

It does this be validating two criteria:

* The extent text of the command matches `'^{?& \$wrappedCmd @PSBoundParameters ?}?$'`, which matches the specific constructs created when auto-generating a proxy command using `[System.Management.Automation.ProxyCommand]::Create()`.
* The topmost parent of the command is like `'*$steppablePipeline.Begin($PSCmdlet)*$steppablePipeline.Process($_)*$steppablePipeline.End()*'`, which matches when all three of the steppable pipeline steps have been defined in the function or file where this proxy command is defined.

Since there is no out-of-the-box mechanism to recursively iterate through nested AST objects to get to the topmost parent, this PR also includes a new private function, `Get-AstTopParent`, that does exactly that.

Fix #1028

### DIFF
--- a/Functions/Coverage.ps1
+++ b/Functions/Coverage.ps1
@@ -306,10 +306,7 @@ Function Get-AstTopParent {
         return $Ast
     }
     elseif ($MaxDepth -le 0) {
-        $params = @{
-            Message = "Max depth reached, skipping this check"
-        }
-        & $SafeCommands['Write-Warning'] @params
+        & $SafeCommands['Write-Verbose'] "Max depth reached, moving on"
         return $null
     }
     else {
--- a/Functions/Coverage.ps1
+++ b/Functions/Coverage.ps1
@@ -296,6 +296,26 @@ function New-CoverageBreakpoint {
     }
 }
 
+Function Get-AstTopParent {
+    param(
+        [System.Management.Automation.Language.Ast] $Ast,
+        [int] $MaxDepth = 30
+    )
+    If ([string]::IsNullOrEmpty($Ast.Parent)) {
+        $Ast
+    }
+    ElseIf ($MaxDepth -le 0) {
+        $params = @{
+            Message = "Max depth reached, skipping this check"
+        }
+        & $SafeCommands['Write-Warning'] @params
+    }
+    Else {
+        $MaxDepth--
+        Get-AstTopParent -Ast $Ast.Parent -MaxDepth $MaxDepth
+    }
+}
+
 function IsIgnoredCommand {
     param ([System.Management.Automation.Language.Ast] $Command)
 
@@ -320,6 +340,15 @@ function IsIgnoredCommand {
             # configuration, we'll ignore it so it doesn't clutter up the coverage analysis with useless junk.
             return $true
         }
+    }
+
+    if ($Command.Extent.Text -match '^{?& \$wrappedCmd @PSBoundParameters ?}?$' -and
+        (Get-AstTopParent -Ast $Command) -like '*$steppablePipeline.Begin($PSCmdlet)*$steppablePipeline.Process($_)*$steppablePipeline.End()*' ) {
+        # Fix for proxy function wrapped pipeline command. PowerShell does not increment the hit count when
+        # these functions are executed using the steppable pipeline; further, these checks are redundant, as
+        # all steppable pipeline constituents already get breakpoints set. This checks to ensure the top parent
+        # node of the command contains all three constituents of the steppable pipeline before ignoring it.
+        return $true
     }
 
     if (IsClosingLoopCondition -Command $Command) {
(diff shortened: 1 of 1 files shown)


nohwnd:
Thanks, this will take some time for me to review, I am currently busy with Pester v5.

PR #1500, opened 2020-04-21
Debug mocks

### LINKED ISSUE #1297: Add possibility to use debugging and breakpoints in mocks

When creating Pester tests, you can use debugging and breakpoints in Pester file. However, it is not possible to set breakpoints or step through mocks. It would be very helpful to add the possibility to do this to Pester, especially when you have complex mocks which you are trying to troubleshoot.

Windows 10  and Windows Server 2012 / 2016 / 2019

PowerShell 5.1

Pester 4.7.3

## 3. Expected Behavior

Set breakpoints to mocks, so you can troubleshoot them.

## 4.Current Behavior

Breakpoints don't work in mocks

## 5. Possible Solution

Implement debugging for mocks

## 6. Context

The current behavior makes debugging mocks harder, especially when you have very complex mocks.

nohwnd:
Remove rewriting the scriptblock, and instead define the variables that would otherwise be bound in above scope. 

This allows both Mock and ParameterFilter to be debugged. And allows usage of aliases in the ParameterFilter. 

This prevents usage of conflicting parameters names direcly in the ParameterFilter, a variable prefixed with _ needs to be used instead. This is a small sacrifice to make, and there is a "warning" in the log. 

image

Fixes #1297

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,3 +1,5 @@
+param ([switch]$Debug)
+
 $ErrorActionPreference = 'Stop'
 Get-Module Pester | Remove-Module
 if (Test-Path "$PSScriptRoot/bin") {
@@ -26,6 +28,8 @@ $script = @(
 
 $sb = [System.Text.StringBuilder]""
 foreach ($s in $script) {
+    $lineNumber = 1
+    $hereString = $false
     foreach ($f in Get-ChildItem $s -File) {
         $lines = Get-Content $f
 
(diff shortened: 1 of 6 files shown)


nohwnd:
@renehernandez please give it a quick  look, it should pass all tests (it does on my machine). Sorry for the formatting changes.

nohwnd:
I think I missed debugging the `Should -Invoke` parameter filter, I need to write some tests to see if I can hit the breakpoints in the respective places.

build.ps1
```diff
@@ -38,7 +42,23 @@ foreach ($s in $script) {
             }
 
             if (-not $noBuild) {
```
renehernandez, line 44: Double negations always throws me off, would it be simple instead:

```
if ($Build) {
....
}
```

renehernandez:
@nohwnd Looks good to me. The comment is more of a stylish thing than actual change request, so feel free to dismiss it if you don't consider it valid enough

nohwnd:
@renehernandez awesome. I used -nobuild because that is similar to --no-build on dotnet test, that's the whole reason.

But back to the real point of this PR. I am glad you like it. I was not sure how to exactly handle dynamic parameters, but as long as the changes pass our tests I think this is good to go. I must say I really like it together with the logging. It is super easy to see what is happening with the mock. I just need to work on an easier way to enable it, but that can wait for a bit.

build.ps1
nohwnd: Keeping nobuild because it is familiar to me from dotnet test

PR #1520, opened 2020-05-04
[v5] Repair Set-ItResult

JustinGrote:
## Summary

Resolves Set-ItResult not working correctly

## To-Do Checklist Prior to Merge

- [x] Fixes improper hashtable format
- [x] Treat Pending and Inconclusive as "skipped" for now
- [x] Fix Because Message in output

### DIFF
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -532,6 +532,10 @@ function Get-WriteScreenPlugin ($Verbosity) {
             $margin = $ReportStrings.Margin * ($level)
             $error_margin = $margin + $ReportStrings.Margin
             $out = $_test.ExpandedName
+            if ($_test.ErrorRecord.FullyQualifiedErrorId -eq 'PesterTestSkipped') {
+                $skippedMessage = [String]$_Test.ErrorRecord
+                [String]$out += " $skippedMessage"
+            }
         }
         elseif ('Minimal' -eq $PesterPreference.Output.Verbosity.Value) {
             $level = 0
@@ -578,11 +582,6 @@ function Get-WriteScreenPlugin ($Verbosity) {
             Skipped {
                 if ($PesterPreference.Output.Verbosity.Value -in 'Normal', 'Detailed', 'Diagnostic') {
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped "$margin[!] $out" -NoNewLine
-                    & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped " is skipped" -NoNewLine
-                    $because = $_test.Data.Because
-                    if ($because) {
-                        & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped ", because $because" -NoNewLine
-                    }
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.SkippedTime " $humanTime"
                 }
                 break
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -577,9 +577,12 @@ function Get-WriteScreenPlugin ($Verbosity) {
 
             Skipped {
                 if ($PesterPreference.Output.Verbosity.Value -in 'Normal', 'Detailed', 'Diagnostic') {
-                    $because = if ($_test.FailureMessage) { ", because $($_test.FailureMessage)" } else { $null }
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped "$margin[!] $out" -NoNewLine
-                    & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped ", is skipped$because" -NoNewLine
+                    & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped " is skipped" -NoNewLine
+                    $because = $_Test.Data.Because
+                    if ($Because) {
+                        & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped ", because $Because" -NoNewLine
+                    }
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.SkippedTime " $humanTime"
                 }
                 break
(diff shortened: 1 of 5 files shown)


JustinGrote:
So it looks like that, since a lot of Pester types moved to C#, that support for "Pending" tests was dropped and the only options are Failed, Passed, Skipped, and NotRun.

@nohwnd should pending tests fall into the "Skipped" or "NotRun" category?

JustinGrote:
@nohwnd this is ready for review, LGTM!

src/functions/Set-ItResult.ps1
```diff
@@ -117,10 +117,10 @@ function New-PesterErrorRecord {
             }).TrimEnd($([System.Environment]::NewLine))
     }
 
-    & $SafeCommands['New-Object'] Management.Automation.ErrorRecord $exception, $errorID, $errorCategory, $targetObject@{
-        exception     = $exception
-        errorId       = $errorID
-        errorCategory = $errorCategory
-        targetObject  = $targetObject
-    }
+    & $SafeCommands['New-Object'] -TypeName Management.Automation.ErrorRecord -ArgumentList @(
```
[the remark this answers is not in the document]
nohwnd: If you modify the c# code you need to run the build.ps1 with -Clean and it will fail to run if there is a session with the dll already loaded. Then you need to close your powershell session and run it again. Annoying, but that is how it is now, the csharp code is not edited that often.

src/functions/Output.ps1
```diff
@@ -548,8 +548,15 @@ function Get-WriteScreenPlugin ($Verbosity) {
         if ($PesterPreference.Debug.ShowNavigationMarkers.Value) {
             $out += ", $($_test.ScriptBlock.File):$($_Test.ScriptBlock.StartPosition.StartLine)"
         }
-
         $result = $_test.Result
+        #Enable display of pending/inconclusive results as in Pester 4 temporarily
```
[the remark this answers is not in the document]
JustinGrote: I'll adjust this so that they all set to skip and reword the because message instead to continue backwards compatibility with set-itresult

src/functions/Set-ItResult.ps1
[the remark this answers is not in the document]
JustinGrote: @nohwnd I just preserved it as it was but fixed the broken hashtable, I'll look for a related example and implement it

JustinGrote:
> This should happen in another layer, presumably in Invoke-TestItem or where is the place I set FailureMessage. This plugin is only responsible for printing the result object.  

I did it so as not to disturb the nunit results per the "quick fix" approach. I'll move it back.

JustinGrote:
@nohwnd New approach:

1. Set-ItResult now modifies the Test object directly and does a generic throw to stop the flow, I don't see a scenario where you set an ItResult and want more code to run, but the "throw" could be hidden behind an optional -Terminating switch as long as cleanup cases were addressed.
1. The message is stored in the test object Data.Because property.
1. Pending and Inconclusive now generate skipped with an extra note in them
1. Output modified to check for a because message
1. Added Deprecation warnings for -Pending and -Inconclusive (do you want me to remove these?)

src/functions/Output.ps1
```diff
@@ -577,9 +577,12 @@ function Get-WriteScreenPlugin ($Verbosity) {
 
             Skipped {
                 if ($PesterPreference.Output.Verbosity.Value -in 'Normal', 'Detailed', 'Diagnostic') {
-                    $because = if ($_test.FailureMessage) { ", because $($_test.FailureMessage)" } else { $null }
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped "$margin[!] $out" -NoNewLine
-                    & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped ", is skipped$because" -NoNewLine
+                    & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped " is skipped" -NoNewLine
```
nohwnd, line 581: I think you misunderstood me, what I meant by another layer was that this code should only look at FailureMessage, and FailureMessage is imho set in Invoke-TestItem. And in here we should only take the message and print it, if possible. But that is not a huge problem, I can fix that later in 5.1. 

What is problematic is that the Data hashtable is for TestCases data, you should not modify that. Please just revert this to the previous solution, to not make this change unnecessarily difficult.

src/functions/Set-ItResult.ps1
```diff
@@ -67,60 +54,14 @@ function Set-ItResult {
     Assert-DescribeInProgress -CommandName Set-ItResult
 
     $result = $PSCmdlet.ParameterSetName
-    $message = "It result set to $result$(if ($Because) { ", $Because" })"
-    $data = @{
-        Result  = $result
-        Because = $Because
-    }
-    $errorRecord = New-PesterErrorRecord -Result $result  -Message $message -Data $data
```
nohwnd, line 75: this was imho okay, you'd just change the message here, and change the Pester.Factory method to also accept data, and that should be it, throwing a runtime exception makes this a possbile conflict with an exception that a user would throw, and we don't want that. (Sure they can throw our exception type, but that is was more unlikely.)

src/functions/Set-ItResult.ps1
```diff
@@ -67,60 +54,14 @@ function Set-ItResult {
...
-                $MyInvocation.ScriptLineNumber
-            })
-        LineText = $(if ($LineText -ne $null) {
-                $LineText
-            }
-            else {
-                $MyInvocation.Line
-            }).TrimEnd($([System.Environment]::NewLine))
+    #TODO: Remove in Pester 6
+    if ($result -in 'Inconclusive','Pending') {
```
nohwnd, line 59: Yes please get rid if the warnings, these would trigger million times. We can add them back in when I make up my mind about the faith of Pending and Inconclusive.

nohwnd:
Thanks, added my comments. Thanks for exploring this with me, it helps. 🙂 If this started to feel like a chore, let me know I can finish it.

src/functions/Output.ps1
JustinGrote: OK, I fixed this by overloading your CreateErrorRecord to allow me to specify the error type for PesterTestSkipped (which you specified elsewhere in the code in a -contains but it really should be an enum)

src/functions/Set-ItResult.ps1
JustinGrote: It now throws using the Pester.Factory method

src/functions/Set-ItResult.ps1
JustinGrote: Removed.

JustinGrote:
@nohwnd OK it uses the Pester Factory approach you suggested, with the message formatting all moved to Set-ItResult.

image

PR #1529, opened 2020-05-08
Fix TestDrive and TestRegistry teardowns when already teared down

### LINKED ISSUE #1524: TestDrive error when calling Pester inside nested Context blocks
Latest v5 prerelease:

Test script:
```powershell
Describe "Testing" {
    Context 'foo' { 
        Context 'bar' {
            It "runs Pester" {
                Invoke-Pester -Path $PSScriptRoot/test.ps1
            }
        }
    }
}
```

nohwnd:
Teardown TestDrive and TestRegistry only when they still exist, but don't fail when someone else already removed them, possibly a child Invoke-Pester running inside of this test.

Fixes #1524

### DIFF
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -191,6 +191,10 @@ function PostProcess-RspecTestRun ($TestRun) {
         }
 
         ## summarize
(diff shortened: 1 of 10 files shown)


mgreenegit:
:shipit:

PR #1532, opened 2020-05-11
Fix path resolve

### LINKED ISSUE #1526: V5 (rc8 and prior): Invoke-Pester throws when using -CI switch together with an array of Paths (happening only on PowerShell 5.1)
Full local bug repro is here: https://github.com/bergmeister/PesterV5BugRepro

Using .Net 4.8.0 on Windows 10 1909

cc @nohwnd 

This is where it happens in CI:
https://dev.azure.com/powershell/psscriptanalyzer/_build/results?buildId=51793&view=logs&j=873b4b40-df3a-56d8-66d1-ef5810ea11cb&t=84435a71-1ef3-583f-b877-ff8d25a10f11&l=22
https://github.com/PowerShell/PSScriptAnalyzer/pull/1444/checks?check_run_id=642234994

nohwnd:
Fix #1526

### DIFF
--- a/src/functions/Coverage.ps1
+++ b/src/functions/Coverage.ps1
@@ -75,9 +75,9 @@ function New-CoverageInfo {
 function Get-CoverageInfoFromDictionary {
     param ([System.Collections.IDictionary] $Dictionary)
 
-    [string] $path = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Path', 'p'
-    if ([string]::IsNullOrEmpty($path)) {
-        throw "Coverage value '$Dictionary' is missing required Path key."
+    $path = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Path', 'p'
+    if ($null -eq $path -or 0 -ge @($path).Count) {
+        throw "Coverage value '$($Dictionary | & $script:SafeCommands['Out-String'])' is missing required Path key."
     }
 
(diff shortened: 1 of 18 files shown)


src/functions/Coverage.ps1
```diff
@@ -75,9 +75,9 @@ function New-CoverageInfo {
 function Get-CoverageInfoFromDictionary {
     param ([System.Collections.IDictionary] $Dictionary)
 
-    [string] $path = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Path', 'p'
-    if ([string]::IsNullOrEmpty($path)) {
-        throw "Coverage value '$Dictionary' is missing required Path key."
+    $path = Get-DictionaryValueFromFirstKeyFound -Dictionary $Dictionary -Key 'Path', 'p'
+    if ($null -eq $path -or 0 -ge @($path).Count) {
```
bergmeister, line 79: Would this be more readable whilst still being the same?
```suggestion
    if ($null -eq $path -or @($path).Count -eq 0) {
```

nohwnd: It is not totally the same, and I do (or should do) these `none` and `any` checks always this way. But yes it would be more readable, but not more reliable.

bergmeister: How is it different? As far as I know PowerShell always populates .Count and is therefore always >=0

PR #1535, opened 2020-05-11
Feature: Allow Output Status Character Customization

JustinGrote:

Adds properties to the PesterConfiguration object and modifies appropriate output items to allow for status symbol customization, such as displaying "PASS/FAIL" or displaying emojis.

Capture


nohwnd:
Why is this worth to have three new options and all the future maintenance?

nohwnd:
while this is fun, I won't merge because I want to avoid adding a lot of options to a plugin that should be in the future replaceable as described in #1555, we can revisit this later.

PR #1536, opened 2020-05-11
Try Powershell Standard

JustinGrote:
Adjusts build to only publish a .NET standard version of the module that works downlevel to PS3+ rather than separate net452 and netstandard2.0 builds

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,7 +16,6 @@ stages:
         pool:
           name: Default
           demands:
-            - net452
             - netcoreapp3.1
         timeoutInMinutes: 3
         steps:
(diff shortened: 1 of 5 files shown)


nohwnd:
It did not work on the older versions right? netstandard2.0 is from net462, or rather net472, but powershell 3 is net451 (or 452 I don't remember). So the the reference assemblies target 3-5 and netstandards targets 6-7.

JustinGrote:
@nohwnd It did work with PowershellStandard 3
https://github.com/PowerShell/PowerShellStandard

But I ran into a few edge cases that I decided I didn't want to mess with right now :)

nohwnd:
Interesting. What were the edgecases?

JustinGrote:
@nohwnd I don't recall specifically but some stuff didn't pass on PSv3 (it loaded though) and I didn't have time to spin up an old box with PSv3 on it to troubleshoot :)

PR #1549, opened 2020-05-16
Make StringArrayOption property perform object[] to string[] transformation automatically in a similar way how PowerShell would do it

### LINKED ISSUE #1548: v5: Run.Path configuration object does not offer the same compatiblity as -Path parameter
Using `rc9` on any version of PowerShell (and presumably any platform, I am using Windows though):
```pwsh
Import-Module Pester
$configuration = [PesterConfiguration]::Default
$configuration.Run.Path = Join-Path $PWD 'foo' # works
$configuration.Run.Path = "$PWD\foo", "$PWD\bar" #works
$configuration.Run.Path = (Join-Path $PWD 'foo'), (Join-Path $PWD 'bar') # does not work
```

The last statement throws an error:
```
SetValueInvocationException: Exception setting "Path": "Cannot convert value "System.Object[]" to type "Pester.StringArrayOption". Error: "Unable to cast object of type 'System.Management.Automation.PSObject' to type 'System.String'.""
```

I consider this a bug because when passing `(Join-Path $PWD 'foo'), (Join-Path $PWD 'bar')`  to the `-Path` parameter, it works.
It seems to be because the `-Path` parameter is of type `[string[]]`, therefore a way to make this work would be casting to it:
```pwsh
$configuration.Run.Path = [string[]] ((Join-Path $PWD 'foo'), (Join-Path $PWD 'bar'))
```
What is the benefit of the used `Pester.StringArrayOption` type for the Run.Path property?

bergmeister:
Fixes #1548

As described in the issue, there are cases where  the cast fails for a valid object array. Since very object has to implement `.ToString()`, we can do the conversion ourselves. This way the behaviour will be more PowerShell-y.

Can you point me at a good place for a unit test for that? The xunit test project don't seem to be run in CI btw.

### DIFF
--- a/src/csharp/Pester/Configuration.cs
+++ b/src/csharp/Pester/Configuration.cs
@@ -220,7 +220,7 @@ namespace Pester
 
         }
 
-        public StringArrayOption(object[] value) : base("", new string[0], value.Cast<string>().ToArray())
+        public StringArrayOption(object[] value) : base("", new string[0], value.Select(oneValue => oneValue.ToString()).ToArray())
         {
 
         }
--- a/tst/Pester.RSpec.Configuration.ts.ps1
+++ b/tst/Pester.RSpec.Configuration.ts.ps1
@@ -186,7 +186,7 @@ i -PassThru:$PassThru {
         }
         t "StringArrayOption can be assigned an array of Objects that don't directly cast to a string" {
             $config = [PesterConfiguration]::Default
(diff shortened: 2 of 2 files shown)


nohwnd:
Would be nice if you added tests that make sure this won't regress. There are some for the options already (in a .ts.ps1 file).

bergmeister:
Yep, just added a test for the exact case that this PR fixes

PR #1555, opened 2020-05-17
Feat: Add a configuration option to suppress per-file Discovery messages

### LINKED ISSUE #1552: [v5] Consider moving "Discovering" messages to Output Diagnostic Level
Because each test file name is shown in the "Running" section, the discovery section duplicates the information and leads to a "wall of purple" if you have tests in a lot of separate files
image

However I don't want to go all the way to minimal, I want the "wall of green" of passing tests to watch progress and see how the performance of each test went.

Therefore, I think the discovery section on output normal should look like it does on output minimal (just that discovery started and completed, two lines) and then do the normal output. This would also be closer to what v4 users would expect on default which will ease the v4-v5 transition.

Here's a mockup of what I think Output Normal should look like:

image
image

JustinGrote:

Adds Output.SummarizeDiscovery option

Resolves #1552

### Default Behavior
image

### Enabling SummarizeDiscovery
image

### DIFF
--- a/src/csharp/Pester/Configuration.cs
+++ b/src/csharp/Pester/Configuration.cs
@@ -1006,6 +1006,7 @@ namespace Pester
     public class OutputConfiguration : ConfigurationSection
     {
         private StringOption _verbosity;
+        private BoolOption _summarizeDiscovery;
 
         public static OutputConfiguration Default { get { return new OutputConfiguration(); } }
         public static OutputConfiguration ShallowClone(OutputConfiguration configuration)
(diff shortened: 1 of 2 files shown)


JustinGrote:
@nohwnd ready for review

nohwnd:
Thanks, but I won't merge that, I plan to make the whole output replaceable and this option would then have to be implemented by anyone who wishes to customize the output. So unless I decide not to publish the plugins, I want to have as few customization options as possible. :)

PR #1556, opened 2020-05-18
Add alias parameter to -HaveParameter

nohwnd:
Moving #1463 into v5


nohwnd:
FYI @lipkau

lipkau:
awesome!
thank you, @nohwnd

PR #1578, opened 2020-05-28
Add Legacy parameter set

nohwnd:


src/Pester.ps1
```diff
@@ -367,6 +417,223 @@ function Invoke-Pester {
                 }
             }
 
+            if ('Legacy' -eq $PSCmdlet.ParameterSetName) {
```
chrisgardnersage, line 420: I think there needs to be a Write-Warning in here to let users know explicitly that they are using a deprecated parameter set that won't be around forever and they need to move over the new approach.

src/Pester.ps1
```diff
@@ -255,17 +255,26 @@ function Has-Flag {
 function Invoke-Pester {
     [CmdletBinding(DefaultParameterSetName = 'Simple')]
     param(
-        [Parameter(Position = 0, Mandatory = 0,  ParameterSetName = "Simple")]
+        [Parameter(Position = 0, Mandatory = 0, ParameterSetName = "Simple")]
+        [Parameter(Position = 0, Mandatory = 0, ParameterSetName = "Legacy")]  # deprecated
+        [Alias("Script")] # deprecated
```
bergmeister, line 260: If the comments said `Legacy` or `Legacy ParameterSet` instead it would be easier to infer that it refers to the `Legacy` parameter set

src/Pester.ps1
nohwnd: Added it, was thinking about the same thing. Linked to release notes that don't exist yet, (obviously).

src/Pester.ps1
nohwnd: done

src/Pester.ps1
```diff
@@ -367,6 +417,224 @@ function Invoke-Pester {
...
             }
 
+            if ('Legacy' -eq $PSCmdlet.ParameterSetName) {
+                Write-Warning "You are using Legacy parameter set that adapts Pester 5 syntax to Pester 4 syntax. This parameter set is deprecated, and does not work 100%, please refer to https://github.com/pester/Pester/releases/tag/5.0.1#legacy-parameter-set for more information."
+                # populate config from parameters and remove them so we
+                # don't inherit them to child functions by accident
+
+                $Configuration = [PesterConfiguration]::Default
+
+                # if ($PSBoundParameters.ContainsKey('Path')) {
```
bergmeister, line 427: left-over code?

nohwnd: yup. cleaned.

PR #1595, opened 2020-06-07
(doc): Add Legacy v4 parameter mapping to config object in README.md

plastikfan:
 As requested by @nohwnd (in issue #1218), this pull request describes the mapping of v4 Legacy parameters to the configuration object. I also noticed that the *PassThru* parameter was missing from the Simple interface, so that mapping was added to that table.

There are also minor 1 line markdown linter adjustments.


nohwnd:
Thanks!

Issue #132, opened 2014-05-06
Don't use Mandatory Parameters. Instead throw if not provided


ghostsquad:
Mandatory parameters force a prompt if that parameter is not provided, which makes situations including when tests are NOT running hard to debug. Instead functions like "Describe" should throw an exception if specific mandatory parameters are not provided.

A recent project of mine had a malformed test. If you can find the bug, +extra credit.

``` powershell
Describe "Stuff" 
{
    It "should do stuff" {
        $true | Should Be $false
    }
}
```

yecril71pl:
Please reopen, we have a regression in ``SHOULD -ERRORID:X`` (without ``-THROW``).

ghostsquad:
@yecril71pl probably just file a new issue...

PR #1610, opened 2020-06-18
Make switches not mandatory

### LINKED ISSUE #1607: prompt on SHOULD -ERRORID:X (without -THROW)

The invocation  ``SHOULD -ERRORID:X`` prompts for parameter ``THROW``.

<DL ><DT >Pester version     : <DD >5.0.2 <CODE >C:\Users\User\Documents\PowerShell\Modules\Pester\5.0.2\Pester.psd1</CODE >
<DT >PowerShell version : <DD >7.0.2
<DT >OS version         : <DD >Microsoft Windows NT 10.0.18363.0</DD >

## 3. Expected Behavior

<DL ><DT >OperationStopped: <DD >Input is not a ScriptBlock</DL >

## 4.Current Behavior

cmdlet Should at command pipeline position 1
Supply values for the following parameters:
<DL ><DT >Throw:</DL >

## 5. Possible Solution

Regression for #132 .
## 6. Context

The PowerShell team requires exception tests to specify ``ERRORID``.

yecril71pl:
The command ``SHOULD -ERRORID:X`` should work all right without an explicit ``-THROW``.  Restores #132.  Fixes #1607.

Should switches should not be mandatory per #132.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
   "files.insertFinalNewline": true,
 
   "editor.tabSize": 4,
-  "editor.formatOnSave": false
+  "editor.formatOnSave": false,
+  "powershell.codeFormatting.addWhitespaceAroundPipe": true
 }
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -156,8 +156,7 @@ function Add-AssertionDynamicParameterSet {
 
     $attribute = New-Object Management.Automation.ParameterAttribute
     $attribute.ParameterSetName = $AssertionEntry.Name
-    $attribute.Mandatory = $true
-
+    
     $attributeCollection = New-Object Collections.ObjectModel.Collection[Attribute]
     $null = $attributeCollection.Add($attribute)
     if (-not ([string]::IsNullOrWhiteSpace($AssertionEntry.Alias))) {
--- a/tst/functions/assertions/PesterThrow.Tests.ps1
+++ b/tst/functions/assertions/PesterThrow.Tests.ps1
@@ -23,7 +23,7 @@ InPesterModuleScope {
             }
 
             It "returns error when -PassThru is specified" {
-                $err = { throw } | Should -Throw -PassThru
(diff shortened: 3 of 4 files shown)


nohwnd:
That build fail on 6_2 is me doing something wrong when PS4 and PS6.2 run on the same agent. Please add tests for this change as well. Hopefully it will pass on your next push.

nohwnd:
This does not do what #132 describes, it should throw when switch is missing instead of not adding it to the parameter set. The switch param should be the main way to disambiguate the parameter set to be used.

Also please don't change the existing tests, instead add new ones for the new behavior.

yecril71pl:
The switch is still added to the parameter set.
The switch is not missing, it is pulled by using `-ERRORID`.  It is implicitly present.

nohwnd:
Let me correct myself. 

> This does not do what #132 describes, it should throw when switch is missing instead of ~not adding it to the parameter set~ just not making it mandatory.

yecril71pl:
I would like to remove the requirement for an explicit switch.  I agree this is a separate issue.  Could I have your opinion on that?  My POV is that `-ERRORID` naturally implies `-THROW` and this excessive verbosity makes the test specification harder to read.

nohwnd:
> Also, as I said, the switch is not missing, it is present because it is pulled by the parameter set.

Yes because -ErrorId is not ambiguous, but this there are other parameters that might be ambiguous, or some might be added and they will become ambiguous. That is why I don't want to set the precedence of not specifying the main assertion switch. 

> this excessive verbosity makes the test specification harder to read.

I don't agree with this. Reading `{ Get-Something } | Should -Throw -ErrorId "SpecificError"`, is in my opinion much easier to understand, than `{ Get-Something } | Should -ErrorId "SpecificError"`.

yecril71pl:
I need your advice on how to detect that a switch parameter is implicit.

nohwnd:
Reverted the change in existing tests and added new test. Let's hope this won't break many existing tests.

PR #1630, opened 2020-07-17
fix formatting for Nunit parameterrized tests

### LINKED ISSUE #1573: v5: ConvertTo-NUnitReport fails consistently
`Invoke-Pester | ConvertTo-NUnitReport` fails consistently every time.

```output
Pester version     : 5.0.0 C:\Program Files\WindowsPowerShell\Modules\Pester\5.0.0\Pester.psd1
PowerShell version : 5.1.18362.752
```

Issue is exactly the same on Linux (Which I had to upgrade, since Pester 4.10 does not work on Linux at all).

```output
Pester version     : 5.0.0 /home/jhp/.local/share/powershell/Modules/Pester/5.0.0/Pester.psd1
PowerShell version : 7.0.0
OS version         : Unix 4.4.0.18362
```
## 3. Expected Behavior

`Invoke-Pester | ConvertTo-NUnitReport` should return an XmlDocument without throwing an error.

## 4.Current Behavior

Currently, `Invoke-Pester | ConvertTo-NUnitReport` throws an error, which depending on the CI configuration results in total failure of the pipeline. Sometimes even regardless of the ErrorAction.
```output
You cannot call a method on a null-valued expression.
At C:\Program Files\WindowsPowerShell\Modules\Pester\5.0.0\Pester.psm1:12947 char:5
+     $XmlWriter.WriteAttributeString('date', $Result.ExecutedAt.ToStri ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull
```

Terminating error in AppVeyor release pipeline on Linux, note the ErrorAction SilentlyContinue
```output
Error: 05/27/2020 12:15:30: 
At /home/appveyor/projects/automatedlab/.build/psake.ps1:82 char:61 + … ojectRoot\Tests | ConvertTo-NUnitReport -ErrorAction SilentlyContinue +                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [<<==>>] Exception: You cannot call a method on a null-valued expression.
```
## 6. Context

Trying to export test results, since it has become impossible to specify the name of the test result file in v5 otherwise. There are no apparent parameters or other documented ways of exporting the NUnit XML file with a custom name. If there are, please point me to the documentation.

### LINKED ISSUE #1612: NUnit output does not show <field> replacement values
NUnit output does not show `<field>` replacement values inside the xml file output.

Test Example:
```
 it "Hard Disk should be at least 15% free, and found <FreeSpacePercent>% free on <DriveLetter>" -TestCases $ht {
            param
            (
                $FreeSpacePercent,
                $DriveLetter
            )
            Add-LogData -Group 'System Information' -Name "Hard Drive" -Key "$($DriveLetter) Free" -Value "$($FreeSpacePercent)%"
            $FreeSpacePercent | Should -BeGreaterOrEqual 15 -Because "we expected the hard drive to have at least 15% free"
        }
```
Pester version     : 5.0.2 C:\Program Files\WindowsPowerShell\Modules\Pester\5.0.2\Pester.psd1
PowerShell version : 5.1.14409.1018
OS version         : Microsoft Windows NT 6.3.9600.0

## 3. Expected Behavior
Test Case Descriptions that is expected:
- Hard Disk should be at least 15% free, and found 11% free on C:

NUnit File Output:
```
                <results>
                  <test-case description="Hard Disk should be at least 15% free, and found 81% free on E:" name="All Environment Tests.Hard Disk should be at least 15% free, and found 81% free on E: (42944186880,34644951040,&quot;81&quot;,42813358080,&quot;VMware Virtual disk SCSI Disk Device&quot;,33040,&quot;E:&quot;,32.27,42813292544,&quot;\\.\PHYSICALDRIVE1&quot;,40,&quot;Disk #1, Partition #0&quot;,&quot;New Volume&quot;)" time="0.0193" asserts="0" success="True" result="Success" executed="True" />
                  <test-case description="Hard Disk should be at least 15% free, and found 11%  free on C:" name="All Environment Tests.Hard Disk should be at least 15% free, and found 11% free on C: (42944186880,4611911680,&quot;11&quot;,42580574208,&quot;VMware Virtual disk SCSI Disk Device&quot;,4398,&quot;C:&quot;,4.3,42580570112,&quot;\\.\PHYSICALDRIVE0&quot;,40,&quot;Disk #0, Partition #1&quot;,&quot;&quot;)" time="0.0195" asserts="0" success="False" result="Failure" executed="True">
                    <failure>
                      <message>Expected the actual value to be greater than or equal to 15, because we expected the hard drive to have at least 15% free, but got '11'.
at $FreeSpacePercent | Should -BeGreaterOrEqual 15 -Because "we expected the hard drive to have at least 15% free", C:\JX\PowerShell Tests\HATS\Modules\ValidationFunctions\ValidationFunctions.psm1:3989</message>
                      <stack-trace>at &lt;ScriptBlock&gt;, C:\JX\PowerShell Tests\HATS\Modules\ValidationFunctions\ValidationFunctions.psm1:3989</stack-trace>
                    </failure>
                  </test-case>
                </results>

```

## 4.Current Behavior
Test Case Description that is wrong:
- Hard Disk should be at least 15% free, and found &lt;FreeSpacePercent&gt;% free on &lt;DriveLetter&gt;

NUnit File Output:
```
                <results>
                  <test-case description="Hard Disk should be at least 15% free, and found &lt;FreeSpacePercent&gt;% free on &lt;DriveLetter&gt;" name="All Environment Tests.Hard Disk should be at least 15% free, and found &lt;FreeSpacePercent&gt;% free on &lt;DriveLetter&gt;(42944186880,34644951040,&quot;81&quot;,42813358080,&quot;VMware Virtual disk SCSI Disk Device&quot;,33040,&quot;E:&quot;,32.27,42813292544,&quot;\\.\PHYSICALDRIVE1&quot;,40,&quot;Disk #1, Partition #0&quot;,&quot;New Volume&quot;)" time="0.0193" asserts="0" success="True" result="Success" executed="True" />
                  <test-case description="Hard Disk should be at least 15% free, and found &lt;FreeSpacePercent&gt;% free on &lt;DriveLetter&gt;" name="All Environment Tests.Hard Disk should be at least 15% free, and found &lt;FreeSpacePercent&gt;% free on &lt;DriveLetter&gt;(42944186880,4611911680,&quot;11&quot;,42580574208,&quot;VMware Virtual disk SCSI Disk Device&quot;,4398,&quot;C:&quot;,4.3,42580570112,&quot;\\.\PHYSICALDRIVE0&quot;,40,&quot;Disk #0, Partition #1&quot;,&quot;&quot;)" time="0.0195" asserts="0" success="False" result="Failure" executed="True">
                    <failure>
                      <message>Expected the actual value to be greater than or equal to 15, because we expected the hard drive to have at least 15% free, but got '11'.
at $FreeSpacePercent | Should -BeGreaterOrEqual 15 -Because "we expected the hard drive to have at least 15% free", C:\JX\PowerShell Tests\HATS\Modules\ValidationFunctions\ValidationFunctions.psm1:3989</message>
                      <stack-trace>at &lt;ScriptBlock&gt;, C:\JX\PowerShell Tests\HATS\Modules\ValidationFunctions\ValidationFunctions.psm1:3989</stack-trace>
                    </failure>
                  </test-case>
                </results>

```
PowerShell Text output is correct:
image

## 5. Possible Solution
Allow `<field>` replacement in the NUnit xml output file.

## 6. Context
We use the NUnit xml to gather the tests off of the remote machine and bring it back for support analysis. 
This bug reduces the efficiency of support.

alexbuit:
# 1. General summary of the pull request

Fix #1573 
Fix #1612

It will replace the Parameter tokens in the test names with the parameter values if provided. If no replacement has taken place, the current behavior will take place (put all parameters cast into [string] after the test name.

The current behavior breaks on test builds we are running with parametersets containing a huge array of hash tables, as they will produce test names with "System.Collections.Hashtable System.Collections.Hashtable ..." at the end, making both the result file very big and break the parser eventually. (apart from the fact it makes the results unreadable)

### DIFF
--- a/src/functions/TestResults.ps1
+++ b/src/functions/TestResults.ps1
@@ -592,15 +592,7 @@ function Write-NUnitTestCaseElement($TestResult, [System.Xml.XmlWriter] $XmlWrit
 }
 
 function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlWriter, [string] $ParameterizedSuiteName, [string] $Path) {
-    $testName = $TestResult.Path -join '.'
-
-    foreach ($testParameterName in $TestResult.Data.Keys){
-        $matchString = [regex]::Escape("<$testParameterName>")
-        if ($testName -match $matchString){
-            $testName = $testName -replace $matchString, [string]$TestResult.Data[$testParameterName]
-            $paramsUsedInTestName = $true
-        }
-    }
+    $testName = $TestResult.ExpandedPath
 
     # todo: this comparison would fail if the test name would contain $(Get-Date) or something similar that changes all the time
     if ($testName -eq $ParameterizedSuiteName) {
@@ -632,7 +624,7 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
         }
     }
 
-    $XmlWriter.WriteAttributeString('description', $TestResult.Name)
+    $XmlWriter.WriteAttributeString('description', $TestResult.ExpandedName)
 
     $XmlWriter.WriteAttributeString('name', $testName)
     $XmlWriter.WriteAttributeString('time', (Convert-TimeSpan $TestResult.Duration))
--- a/src/functions/TestResults.ps1
+++ b/src/functions/TestResults.ps1
@@ -594,31 +594,43 @@ function Write-NUnitTestCaseElement($TestResult, [System.Xml.XmlWriter] $XmlWrit
 function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlWriter, [string] $ParameterizedSuiteName, [string] $Path) {
     $testName = $TestResult.Path -join '.'
 
+    foreach ($testParameterName in $TestResult.Data.Keys){
+        $matchString = [regex]::Escape("<$testParameterName>")
+        if ($testName -match $matchString){
+            $testName = $testName -replace $matchString, [string]$TestResult.Data.$testParameterName
+            $paramsUsedInTestName = $true
+        }
+    }
+
     # todo: this comparison would fail if the test name would contain $(Get-Date) or something similar that changes all the time
     if ($testName -eq $ParameterizedSuiteName) {
         $paramString = ''
         if ($null -ne $TestResult.Data) {
-            $params = @(
-                foreach ($value in $TestResult.Data.Values) {
-                    if ($null -eq $value) {
-                        'null'
-                    }
-                    elseif ($value -is [string]) {
-                        '"{0}"' -f $value
-                    }
-                    else {
-                        #do not use .ToString() it uses the current culture settings
-                        #and we need to use en-US culture, which [string] or .ToString([Globalization.CultureInfo]'en-us') uses
-                        [string]$value
-                    }
-                }
-            )
+            $paramsUsedInTestName =$false
 
-            $paramString = "($($params -join ','))"
+            if (-not $paramsUsedInTestName) {
+                $params = @(
+                    foreach ($value in $TestResult.Data.Values) {
+                        if ($null -eq $value) {
+                            'null'
+                        }
+                        elseif ($value -is [string]) {
+                            '"{0}"' -f $value
+                        }
+                        else {
+                            #do not use .ToString() it uses the current culture settings
+                            #and we need to use en-US culture, which [string] or .ToString([Globalization.CultureInfo]'en-us') uses
+                            [string]$value
+                        }
+                    }
+                )
+
+                $paramString = "($($params -join ','))"
+        
+                $testName = "$testName$paramString"
+            }
         }
     }
-
-    $testName = "$testName$paramString"
 
     $XmlWriter.WriteAttributeString('description', $TestResult.Name)
 
@@ -719,6 +731,7 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
         }
     }
 }
+
 function Get-RunTimeEnvironment() {
     # based on what we found during startup, use the appropriate cmdlet
     $computerName = $env:ComputerName
--- a/src/functions/TestResults.ps1
+++ b/src/functions/TestResults.ps1
@@ -618,7 +618,6 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
                 )
 
                 $paramString = "($($params -join ','))"
-        
                 $testName = "$testName$paramString"
             }
         }
--- a/src/functions/TestResults.ps1
+++ b/src/functions/TestResults.ps1
@@ -597,7 +597,7 @@ function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlW
     foreach ($testParameterName in $TestResult.Data.Keys){
         $matchString = [regex]::Escape("<$testParameterName>")
         if ($testName -match $matchString){
-            $testName = $testName -replace $matchString, [string]$TestResult.Data.$testParameterName
+            $testName = $testName -replace $matchString, [string]$TestResult.Data[$testParameterName]
             $paramsUsedInTestName = $true
         }
     }
--- a/src/Pester.Runtime.psm1
+++ b/src/Pester.Runtime.psm1
@@ -448,7 +448,7 @@ function Invoke-TestItem {
 
         $Test.ExpandedName = & $state.ExpandName -Name $Test.Name -Data $Test.Data
 
-        $test.ExpandedPath = & $state.ExpandName -Name ($Test.Path -join '.') -Data $Test.Data
+        $test.ExpandedPath = "$($Test.Block.Path -join '.').$($Test.ExpandedName)"
 
         $block = $Test.Block
         if ($PesterPreference.Debug.WriteDebugMessages.Value) {
--- a/src/Pester.Runtime.psm1
+++ b/src/Pester.Runtime.psm1
@@ -448,6 +448,8 @@ function Invoke-TestItem {
 
         $Test.ExpandedName = & $state.ExpandName -Name $Test.Name -Data $Test.Data
 
+        $test.ExpandedPath = & $state.ExpandName -Name ($Test.Path -join '.') -Data $Test.Data
+
         $block = $Test.Block
         if ($PesterPreference.Debug.WriteDebugMessages.Value) {
             Write-PesterDebugMessage -Scope Runtime "Running test '$($Test.Name)'."
--- a/src/csharp/Pester/Test.cs
+++ b/src/csharp/Pester/Test.cs
@@ -30,6 +30,7 @@ namespace Pester
         public List<string> Path { get; set; }
         public IDictionary Data { get; set; }
         public string ExpandedName { get; set; }
+        public string ExpandedPath { get; set; }
 
         public string Result { get; set; }
         public List<object> ErrorRecord { get; set; }
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -574,7 +574,7 @@ function Get-WriteScreenPlugin ($Verbosity) {
             $level = 0
             $margin = ''
             $error_margin = $ReportStrings.Margin
-            $out = "$($_test.Block.Path -join '.').$($_test.ExpandedName)"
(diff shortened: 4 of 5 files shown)


src/functions/TestResults.ps1
```diff
@@ -594,32 +594,44 @@ function Write-NUnitTestCaseElement($TestResult, [System.Xml.XmlWriter] $XmlWrit
 function Write-NUnitTestCaseAttributes($TestResult, [System.Xml.XmlWriter] $XmlWriter, [string] $ParameterizedSuiteName, [string] $Path) {
     $testName = $TestResult.Path -join '.'
 
+    foreach ($testParameterName in $TestResult.Data.Keys){
```
nohwnd, line 597: I think there already is a property on the test object that has the expanded name, I think it's called DisplayName

alexbuit: If I run the following, I do not see the parameterized SomeItem and the current problem we face is that the nunit output that is now generated actually blows up our bamboo build (as it uses the description field to name the tests in it's output, and having test names that are a 23.528 characters blows up completely (and also in a very unpleasant way, the build actually gets stuck without error) Although the latter is a bug in Bamboo, we are causing it by submitting test results in a unclean way, so I would rather fix the input then the parser. (I kept the about of hashtables in the example below limited to 11, but you can imagine what it would look like when applying this to our production build running on way bigger data sets.

`
$testparam = @(
    @{
        SomeItem = 'bla'
        SomeItemArray = 0..10 | ForEach-Object { @{SomeKey = "bla_$_"} }
    }
)
Describe 'a' {
    It 'great <SomeItem>' {
        Param (
            [string]$SomeItem,
            [hashtable[]]$SomeItemArray
        )
        $SomeItemArray[0].SomeKey | Should -be 'bla_0'
        $SomeItem | Should -be 'bla'
    } -TestCases $testParam
}
`
Results in:
`
<test-results xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="nunit_schema_2.5.xsd" name="Pester" total="1" errors="0" failures="0" not-run="0" inconclusive="0" ignored="0" skipped="0" invalid="0" date="2020-07-20" time="10:34:57">
  <environment clr-version="4.0.30319.42000" user-domain="DESKTOP-9UMGMK5" cwd="C:\repo\Pester\src\functions" platform="Microsoft Windows 10 Pro|C:\WINDOWS|\Device\Harddisk0\Partition4" machine-name="DESKTOP-9UMGMK5" nunit-version="2.5.8.0" os-version="10.0.18363" user="alexb" />
  <culture-info current-culture="en-NL" current-uiculture="en-US" />
  <test-suite type="TestFixture" name="Pester" executed="True" result="Success" success="True" time="0.0655" asserts="0" description="Pester">
    <results>
      <test-suite type="TestFixture" name="C:\temp\bla2.tests.ps1" executed="True" result="Success" success="True" time="0.0655" asserts="0" description="C:\temp\bla2.tests.ps1">
        <results>
          <test-suite type="TestFixture" name="a" executed="True" result="Success" success="True" time="0.0394" asserts="0" description="a">
            <results>
              <test-suite type="ParameterizedTest" name="a.great &lt;SomeItem&gt;" executed="True" result="Success" success="True" time="0.0173" asserts="0" description="great &lt;SomeItem&gt;">
                <results>
                  <test-case description="great &lt;SomeItem&gt;" name="a.great &lt;SomeItem&gt;(&quot;bla&quot;,System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collections.Hashtable System.Collecti
ons.Hashtable System.Collections.Hashtable System.Collections.Hashtable)" time="0.0173" asserts="0" success="True" result="Success" executed="True" />
                </results>
              </test-suite>
            </results>
          </test-suite>
        </results>
      </test-suite>
    </results>
  </test-suite>
</test-results>
`

nohwnd: Okay, I just want to make sure that the output to screen matches the output to file, so we shoukd only convert the name to DisplayName once, and use it in all the places where it is reported. Maybe the display name converter needs fixing instead?

alexbuit: committed a new version where the $test.ExpandedName and (new generated property) $test.ExpandedPath are used, is this the direction you were thinking?

alexbuit: Looks like one of the builds had a bad timing issue:
https://nohwnd.visualstudio.com/Pester/_build/results?buildId=984&view=logs&j=ff75994e-ad5b-5649-588c-530e2c306b35&t=bb64baaf-a3d6-5598-2b44-c9bedcc4901a&l=1670
Reading the logs it should just have taken only a few seconds, but it took way longer, about 7 seconds, (also it's discovery phase, which took 11 seconds)

I will push the code with a minor change to re-trigger the builds to verify if a second run will fix the issue...

alexbuit: To me it feels like we should drop the entire part where the parameters are beeing parsed into the description, but I did not yet do that, as that will change behaviour for tests that do not have parameter tags in their test names

alexbuit: what might be a good alternative there would be implementing what is written in PR https://github.com/pester/Pester/pull/1089
instead.

nohwnd: > To me it feels like we should drop the entire part where the parameters are beeing parsed into the description, but I did not yet do that, as that will change behaviour for tests that do not have parameter tags in their test names

which part would be that? and what would be the impact?

nohwnd:
The code looks good, is it ready to merge or is there more work to be done?

src/functions/TestResults.ps1
alexbuit: I left the code below in the NUnit output generation, but it will only kick in when there are no <Value> tags in the testname, removing this would mean parameterized tests that do not have <value> tags will all just be named exactly the same.

The alternative is in PR #1089 where the parameters are passed in as a seperate XML property into the NUnit file, thus the information is still available, but it will not change the testname.

starting from this line:
https://github.com/pester/Pester/pull/1630/files#diff-9d06394664e89fba6b48c05ff5001458R597
```powershell
 # todo: this comparison would fail if the test name would contain $(Get-Date) or something similar that changes all the time
    if ($testName -eq $ParameterizedSuiteName) {
        $paramString = ''
        if ($null -ne $TestResult.Data) {
            $params = @(
                foreach ($value in $TestResult.Data.Values) {
                    if ($null -eq $value) {
                        'null'
            $paramsUsedInTestName =$false

            if (-not $paramsUsedInTestName) {
                $params = @(
                    foreach ($value in $TestResult.Data.Values) {
                        if ($null -eq $value) {
                            'null'
                        }
                        elseif ($value -is [string]) {
                            '"{0}"' -f $value
                        }
                        else {
                            #do not use .ToString() it uses the current culture settings
                            #and we need to use en-US culture, which [string] or .ToString([Globalization.CultureInfo]'en-us') uses
                            [string]$value
                        }
                    }
                    elseif ($value -is [string]) {
                        '"{0}"' -f $value
                    }
                    else {
                        #do not use .ToString() it uses the current culture settings
                        #and we need to use en-US culture, which [string] or .ToString([Globalization.CultureInfo]'en-us') uses
                        [string]$value
                    }
                }
            )
                )

            $paramString = "($($params -join ','))"
                $paramString = "($($params -join ','))"
                $testName = "$testName$paramString"
            }
        }
    }
```

bielawb:
I spoke with @alexbuit before he went on vacation and got confirmation the code is ready.  Thanks!

src/functions/TestResults.ps1
alexbuit: We can merge this PR as is, I will create a seperate PR to get this idea clarified, maybe that will be easier

alexbuit: I still need to write tests for that change, but the draft looks like this (compared to this PR's version)
https://github.com/alexbuit/Pester/compare/nunitTestcases...alexbuit:NunitTestParameters?expand=1

nohwnd: Thanks, please do create another PR for that :)

PR #1639, opened 2020-07-27
Nunit test parameters

alexbuit:

This PR will move the parameters of a parameterized test to a child property in NUnit called properties, instead of adding it to the name.

The values of the parameters are formatted according to their type. Out-String is used to allow hashtables amongst others to become readable aswell.


alexbuit:
@nohwnd As promised, the PR for putting the parameters in Nunit output as a child property. I used the idea of https://github.com/pester/Pester/pull/1089 so all praise should go to @rdbartram for spinning up this idea

PR #1664, opened 2020-08-24
Fix line-based filter for tests with multiline testcases

### LINKED ISSUE #1659: VSCode cannot run tests with testcases
Code lens in vscode cannot run tests when TestCases are provided because we are using the opening brace position as the line filter, but  test cases will move it further down, not matching the filter. To avoid difficult parsing let's change that to using     `$MyInvocation.ScriptLineNumber` instead of  `$ScriptBlock.StartPosition.StartLine`, which will give us the position of the It (and other) commands directly. 

The VSCoide code lens sends a line filter to identify which test to run. The line filter is simply a filepath + line in that file. In pester we identify the line by the position of the opening curly brace of the scriptblock, but in vscode the lens will take the position of the It command. This is not a problem for test1 where the numbers are the same. But is problem from test2 where we have test cases and the It command and curly brace are not on the same line. So to fix this we would use the position of the It command in Pester, which fixes VSCode and also makes identifying the correct line easier for users who use the line filter manually.

```powershell
function _it_ ($Name, [ScriptBlock] $Test, $TestCases){    
    "It position: $($MyInvocation.ScriptLineNumber)"
    "{ position:  $($Test.StartPosition.StartLine)"
}

_it_ "test1" { 
    1 -eq 2
}

_it_ "test2" -TestCases @(
    @{Value = 1}
    @{Value = 2}
) -Test {
    1 -eq 2
}
```
More info: https://github.com/PowerShell/vscode-powershell/issues/2880

fflaten:

"Run Test" in VS Code doesn't work for running a single test which contains multiple lines to define testcases. The issue is caused by differences in how VS Code would identify the line number for a test compared to Pester's logic. Pester uses the start of scriptblock as the line number, while VS Code uses the line of `It`. VS Code's approach is preferred going forward as it's more consistent with other scenarios like describe, context and tests without testcases as well as being easier for an end user. 

```
It "Test with multiline testcases" -TestCases @(
    @{Value = 1}
    @{Value = 2}
    ) {
        1 | Should -Be 1
    }
}

# Pester's expectation of linefilter for the test above
# Before fix: filepath:4 (start of scriptblock)
# After fix: filepath:1 (line of "It")
```

This PR changes the linefilter-check to use a new "StartLine" property in Pester-objects of type Block (Describe/Context) and Test (It) which stores the location of the function name. This should solve the issue mentioned above.

Fix #1659 
Original issue: https://github.com/PowerShell/vscode-powershell/issues/2880

- [x] Extend Block/Test.cs
- [x] Update line filter
- [x] Add test for multiline testcases
- [x] Fix failing tests
- [x] Update affected output functions: Get-WriteScreenPlugin

### DIFF
--- a/src/Pester.Runtime.psm1
+++ b/src/Pester.Runtime.psm1
@@ -369,10 +369,10 @@ function New-Test {
         [String] $Name,
         [Parameter(Mandatory = $true, Position = 1)]
         [ScriptBlock] $ScriptBlock,
+        [int] $StartLine,
         [String[]] $Tag = @(),
         [System.Collections.IDictionary] $Data = @{ },
         [String] $Id,
-        [int] $StartLine,
         [Switch] $Focus,
         [Switch] $Skip
     )
@@ -2348,12 +2348,12 @@ function New-ParametrizedTest () {
         [String] $Name,
         [Parameter(Mandatory = $true, Position = 1)]
         [ScriptBlock] $ScriptBlock,
+        [int] $StartLine,
         [String[]] $Tag = @(),
         # do not use [hashtable[]] because that throws away the order if user uses [ordered] hashtable
         [System.Collections.IDictionary[]] $Data = @{ },
         [Switch] $Focus,
-        [Switch] $Skip,
-        [int] $StartLine
+        [Switch] $Skip
     )
 
     # we don't need to switch the timer, all the code that runs during discovery is "overhead"
@@ -2365,7 +2365,7 @@ function New-ParametrizedTest () {
     $id = $ScriptBlock.StartPosition.StartLine
     foreach ($d in $Data) {
         #    $innerId = if (-not $hasExternalId) { $null } else { "$Id-$(($counter++))" }
-        New-Test -Id $id -Name $Name -Tag $Tag -ScriptBlock $ScriptBlock -Data $d -Focus:$Focus -Skip:$Skip -StartLine $StartLine
+        New-Test -Id $id -Name $Name -Tag $Tag -ScriptBlock $ScriptBlock -StartLine $StartLine -Data $d -Focus:$Focus -Skip:$Skip
     }
 }
 
--- a/src/Pester.Runtime.psm1
+++ b/src/Pester.Runtime.psm1
@@ -154,6 +154,7 @@ function New-Block {
         [String] $Name,
         [Parameter(Mandatory = $true)]
         [ScriptBlock] $ScriptBlock,
+        [int] $StartLine,
         [String[]] $Tag = @(),
         [HashTable] $FrameworkData = @{ },
         [Switch] $Focus,
@@ -184,6 +185,7 @@ function New-Block {
     $block.Path = $Path
     $block.Tag = $Tag
     $block.ScriptBlock = $ScriptBlock
+    $block.StartLine = $StartLine
     $block.FrameworkData = $FrameworkData
     $block.Focus = $Focus
     $block.Id = $Id
--- a/src/Pester.Runtime.psm1
+++ b/src/Pester.Runtime.psm1
@@ -372,6 +372,7 @@ function New-Test {
         [String] $Name,
         [Parameter(Mandatory = $true, Position = 1)]
         [ScriptBlock] $ScriptBlock,
+        [Parameter(Mandatory = $true)]
         [int] $StartLine,
         [String[]] $Tag = @(),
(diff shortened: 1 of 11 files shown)


fflaten:
Would this be considered a breaking fix considering someone might have scripts using line filter based on the current behaviour?

nohwnd:
@fflaten I don't think that it is breaking change, I doubt anyone was using it apart from VSCode and our own tests.

fflaten:
Haven't been able to check out the last task yet, but came up with an alternative. 

Instead of specifying StartLine as a parameter to `New-Test/New-Block` like `-StartLine $MyInvocation.ScriptLineNumber`, we could use `Get-PSCallStack` inside those functions to dynamically get the first external method-call. This would probably work with future methods as well as doesn't require a change to `It/Context/Describe`.

Thoughts?

```
# DEMO
function New-Test {
...

#$script:PSCommandPath = Pester.psm1 location
$caller = (& $SafeCommands["Get-PSCallStack"] | & $SafeCommands["Where-Object"] -FilterScript { $_.InvocationInfo.PSCommandPath -ne $script:PSCommandPath })[0].InvocationInfo
#Write-Host "New-Test called from command '$($caller.MyCommand)' in $($caller.ScriptName):$($caller.ScriptLineNumber). Line '$($caller.Line)'."

$test = [Pester.Test]::New()
$test.StartLine = $caller.ScriptLineNumber
...
```

nohwnd:
Get-PSCallstack is very slow, please don't use it.

nohwnd:
Converted it to a Draft PR, I always forget to remove the WIP: prefix. Please let me know when it's ready for review & merge.

fflaten:
Cool. What about the last part in my previous comment. Do we need to revert https://github.com/pester/Pester/pull/1664/commits/134f0887a51c6324030d55122836ac8c2a40890d to stay backward compatible with third-party implementations of Block/Test or is it probably not needed?

That's the only thing outstanding AFAIK. Not sure why the PR fails some checks. Feels like the PS4 host is failing random PTests.

nohwnd:
You don't have to revert it, I don't think there are any such implementations.

src/Pester.Runtime.psm1
```diff
@@ -154,6 +154,7 @@ function New-Block {
         [String] $Name,
         [Parameter(Mandatory = $true)]
         [ScriptBlock] $ScriptBlock,
+        [int] $StartLine,
```
nohwnd, line 157:
```suggestion
        [Parameter(Mandatory = $true)]
        [int] $StartLine,
```

src/Pester.Runtime.psm1
```diff
@@ -369,6 +371,7 @@ function New-Test {
         [String] $Name,
         [Parameter(Mandatory = $true, Position = 1)]
         [ScriptBlock] $ScriptBlock,
+        [int] $StartLine,
```
nohwnd, line 374:
```suggestion
        [Parameter(Mandatory = $true)]
        [int] $StartLine,
```

nohwnd:
Just two suggestions, to make the parameter mandatory, to prevent us from errors, and to let possible external implementations (but there are probably none), that that this is needed.

fflaten:
> Just two suggestions, to make the parameter mandatory, to prevent us from errors, and to let possible external implementations (but there are probably none), that that this is needed.

I actually avoided that due to all the tests using New-Block/Test directly 🙂

nohwnd:
Duh!

nohwnd:
Tried it with this code on VSCode and it works perfect: 

```powershell
Describe "d1" `
{
    Context "c1" `
    {
        It "User with guard" `
        {
            1 | Should -Be 2
        }
    }
}
```

One more tiny suggestion added in comments.

src/Pester.Runtime.psm1
```diff
@@ -184,6 +185,7 @@ function New-Block {
     $block.Path = $Path
     $block.Tag = $Tag
     $block.ScriptBlock = $ScriptBlock
+    $block.StartLine = $StartLine
```
nohwnd, line 188: Maybe the $StartLine should be nullable and fallback to $ScriptBlock.StartPosition.StartLine when the value is not provided? That way I am way less likely to run into this when I implement some other LineFilter and test it in the Runtime, but not RSpec implementation? 

Another option would be to fallback to $MyInvocation of the New-* command, but that seems like a worse thing to do.

PR #1665, opened 2020-08-25
(GH-1655) Add support for "-Not" to "Should -Invoke" tests

### LINKED ISSUE #1655: [v5] Using `-Not` for `Should -Invoke` does not appear to work

Using the new mock assertion syntax, it appears that `-Not` has no effect. For example.

I would expect the line below to assert that the function `Invoke-OtherThing` is never called

`Should -Not -Invoke Invoke-OtherThing`

Windows 10 : 19041
Pester : 5.0.3
PowerShell : 7.0.0

Repro Test File

``` powershell
Describe 'Repro' {
  BeforeAll {
    Function Invoke-OtherThing { 1 }
    Function Invoke-Thing([switch]$CallOtherThing) { if ($CallOtherThing) { Invoke-OtherThing } }
  }

  it 'should call thing when told to' {
    Mock Invoke-OtherThing { 'MockResult'}

    $result = Invoke-Thing -CallOtherThing
    $result | Should -Be 'MockResult'

    Should -Invoke Invoke-OtherThing
  }

  it 'should not call thing when told to' {
    Mock Invoke-OtherThing { 'MockResult'}

    $result = Invoke-Thing
    $result | Should -BeNullOrEmpty

    Should -Not -Invoke Invoke-OtherThing
  }
}
```

Repro output
``` text
C:\Source\somewhere [munge-for-syd-mel +1 ~2 -0 !]> invoke-pester -Output Detailed .\Tests\Scripts\test.ps1

Discovering in C:\Source\somewhere\Tests\Scripts\test.ps1.
Found 2 tests. 94ms
Discovery finished in 203ms.

Running tests from 'C:\Source\somewhere\Tests\Scripts\test.ps1'
Describing Repro
  [+] should call thing when told to 497ms (451ms|46ms)
  [-] should not call thing when told to 51ms (46ms|5ms)
   Expected Invoke-OtherThing to be called at least 1 times but was called 0 times
   at Should -Not -Invoke Invoke-OtherThing, C:\Source\somewhere\Tests\Scripts\test.ps1:22
   at <ScriptBlock>, C:\Source\somewhere\Tests\Scripts\test.ps1:22
Tests completed in 1.38s
Tests Passed: 1, Failed: 1, Skipped: 0 NotRun: 0
```

## 3. Expected Behavior

Using `-Not` should set the number of expected invocations to zero

## 4.Current Behavior

Using `-Not` sets the number of expected invocations to one

glennsarti:
Fixes #1655 

Pester has a C# Solution which requires .Net Framework SDKs and Developer Packs
in order to compile. The targetted frameworks can be found in
`src\csharp\Pester\Pester.csproj`.  This commit adds instructions for where to
download these files from.
 
Previously Should -Invoke would ignore a negative expectation i.e. Should -Not
-Invoke.  This commit updates the Should-Invoke and Should-InvokeInternal
functions to pass through the Negation and then change the error raising
behaviour.  In particular, using -Not should have the opposite effect of if
it was not negated. That is, if it raised an error, then the negated expectation
would not raise an error, and vice versa.

This commit also adds tests for using -Not.

Note that using -Not and -ExclusiveFilter is not a valid combination as it
does not behave in an obvious manner.

---

- [x] Add tests for scenarios
- [x] Cleanup Build.md file
- [x] Cleanup commits

### DIFF
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -708,6 +708,70 @@ Describe "When Calling Should -Invoke 0 without exactly" {
     }
 }
 
+Describe "When Calling Should -Not -Invoke without exactly" {
+    BeforeEach {
+        Mock FunctionUnderTest {}
+        FunctionUnderTest "one"
+
+        try {
+            Should -Not -Invoke FunctionUnderTest
+        }
+        Catch {
+            $result = $_
+        }
+    }
+
+    It "Should throw if mock was called once" {
+        $result.Exception.Message | Should -Be "Expected FunctionUnderTest not to be called exactly 1 times"
+    }
+
+    It "Should not throw if mock was not called" {
+        Should -Not -Invoke FunctionUnderTest -ParameterFilter { $param1 -eq "stupid" }
+    }
+}
+
+Describe "When Calling Should -Not -Invoke [Times] without exactly" {
+    BeforeEach {
+        Mock FunctionUnderTest {}
+    }
+
+    It "Should not throw if the mock was called less (<MockCalls>) than the number of times specified (<Times>)" -TestCases @(
+        @{ MockCalls = 3; Times = 15 }
+        @{ MockCalls = 2; Times = 5 }
+        @{ MockCalls = 0; Times = 1 }
+    ) {
+        param($MockCalls, $Times)
+
+        for ($i = 0; $i -lt $MockCalls; $i++) {
+            FunctionUnderTest "one"
+        }
+
+        Should -Not -Invoke FunctionUnderTest $Times
+    }
+
+    It "Should throw if the mock was called (<MockCalls>) at least the number of times specified (<Times>)" -TestCases @(
+        @{ MockCalls = 15; Times = 3 }
+        @{ MockCalls = 3; Times = 3 }
+        @{ MockCalls = 1; Times = 1 }
+        @{ MockCalls = 0; Times = 0 }
+    ) {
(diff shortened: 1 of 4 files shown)


glennsarti:
@nohwnd   Ready for review.

nohwnd:
Looks good. I personally don't see much value in being able to say "was called less than x times", but why not. Please add tests for the IMHO most common case where you don't specify the number of times `Should -Not -Invoke f`, and possibly also one where we don't specify the times parameter but specify exactly `Should -Not -Invoke f -Exactly`, which is unlikely case but should behave correctly.

glennsarti:
>where you don't specify the number of times 

Whoops.  I did have those there at one point.  I will add them in.

tst/functions/Mock.Tests.ps1
```diff
@@ -708,6 +708,70 @@ Describe "When Calling Should -Invoke 0 without exactly" {
     }
 }
 
+Describe "When Calling Should -Not -Invoke without exactly" {
+    BeforeEach {
+        Mock FunctionUnderTest {}
+        FunctionUnderTest "one"
+
+        try {
+            Should -Not -Invoke FunctionUnderTest
```
glennsarti, line 717: Added test without explicit Times

glennsarti:
Added an additional test scenario.  The other was already there (`Should -Not -Invoke f -Exactly`)

Ready for review again.

nohwnd:
Thank you!

PR #1671, opened 2020-09-02
Parametric scripts

### LINKED ISSUE #1485: v5: Parametrized scripts are not implemented
The syntax for parametrizing scripts is not implemented. 

_originally posted by @JustinGrote on [twitter](https://twitter.com/JustinWGrote/status/1247234235447926790?s=20)_

For Pester 5, how am I supposed to run pester tests with param blocks? I don't see any docs on it, in 4 there was the rather obtuse hashtable syntax. Is this not supported in 5 and should I switch to environment variables or something?

I found https://github.com/pester/Pester/blob/v5.0/README.md#advanced-interface

But no answers here. Seems to me that $config.run.Path needs to support a hashtable in addition to string:
```powershell
path = @{
  path = 'path/to/my/test.ps1'
  parameters = @{
    mycustomparam = $true
    verbose = $true
  }
}
```

**TODO:** 

- [x] Single path with single set of data

```powershell
@{
    Path = "C:\temp\abc.Tests.ps1",
    Data = @{
        Name = "Jakub"
        Age = 31
    }
}
```

- [x] Multiple instances of the same path each with it's own data
```powershell
@(
    @{
        Path = "C:\temp\abc.Tests.ps1",
        Data = @{
            Name = "Jakub"
            Age = 31
        }
    }
    @{
        Path = "C:\temp\abc.Tests.ps1",
        Data = @{
            Name = "Thomas"
            Age = 29
        }
    }
)
```

- [x] Single path with multiple sets of data (alternative to the above)

```powershell
@{
    Path = "C:\temp\abc.Tests.ps1",
    Data = @(
        @{
            Name = "Jakub"
            Age = 31
        }
        @{ 
            Name = "Thomas"
            Age = 29
        }
    )
}
```

- [x] Multiple different paths each with it's own data
```powershell
@(
    @{
        Path = "C:\temp\abc.Tests.ps1",
        Data = @{
            Name = "Jakub"
            Age = 31
        }
    }
    @{
        Path = "C:\temp\gef.Tests.ps1",
        Data = @{
            Color = "Blue"
        }
    }
)
```

- [x] Providing path with wildcard should expand the path to multiple containers

- [x] Providing Path that resolves to the same path as Container should skip the paths that already have data, because those files are parametrized and would likely fail

- [ ] API review

### LINKED ISSUE #1558: Reimplementation: Param Blocks in Tests

With Pester 5 it appears support for test parameters and the hashtable syntax for specifying them has been dropped, so a lot of my refactoring has unfortunately been changing these to environment variables.

Params on tests make sense in testing for one-off items, or adjusting the environment such as using the same tests to test a built module vs a source module, or specify a tweak in a generalized test to a more specific environment.

## Potential Implementations

One approach I like a lot is how Invoke-Build reads the params from build scripts and then makes them available as dynamic parameters, so if a build script specifies $BuildPath in its param block, you can do:

`Invoke-Build -BuildPath 'path/to/my/thing'`
Which is very intuitive and dynamic params can easily be sanitized to not allow override of pester parameters (which is easier with the smaller surface area of parameters on `Invoke-Pester` now)

This gets tricky however since multiple tests can have different param blocks, this can be solved by prefixing the parameter with a sanitized name of the test file, or have the test file itself be the parameter that accepts a hashtable of parameters. I considered also just accepting param blocks at the various Describe/Context/It levels and use those names but that seems overly complicated.

Other options is drive it from a .psd1 settings file, or make it part of the PesterConfiguration schema.

Just wondering if there was a best practices approach to apply runtime parameters to Pester (environment variables and global variables are "ew" of course) before I look more into this.

nohwnd:
Add the ability to provide parameters to scripts via `Invoke-Pester`. 

Fix #1485 
Fix #1558


nohwnd:
This is how it would be used, you can provide one or multiple of those objects, and one or multiple sets of data in each objects: 

https://github.com/pester/Pester/pull/1671/files#diff-cde464fba76b676794f7d9560b6443acR571-R575

src/Pester.ps1
```diff
@@ -288,6 +288,9 @@ function Invoke-Pester {
         [Parameter(ParameterSetName = "Legacy")] # Legacy set for v4 compatibility during migration - deprecated
         [Switch] $PassThru,
 
+        [Parameter(ParameterSetName = "Simple")]
```
nohwnd, line 291: Is this good API?

nohwnd:
@JustinGrote review pls :)

PR #1678, opened 2020-09-08
Add fallbackFallback startline

### LINKED ISSUE #1673: $StartLine fallback
Maybe the $StartLine should be nullable and fallback to $ScriptBlock.StartPosition.StartLine when the value is not provided? That way I am way less likely to run into this when I implement some other LineFilter and test it in the Runtime, but not RSpec implementation? 

Another option would be to fallback to $MyInvocation of the New-* command, but that seems like a worse thing to do.

_Originally posted by @nohwnd in https://github.com/pester/Pester/pull/1664#r482717661_

fflaten:
# 1. General summary of the pull request

#1664 changed the way StartLine in test/block-objects where detected to fix linefilter in scenarios where It/Context/Describe has the scriptblock-start on a different line than the function name. This added a requirement to specify a StartLine-parameter when using New-Test/Block directly which might break som custom solutions and makes RSpec-tests unnecessary complicated. 

This PR adds a fallback to New-*Test/Block which ensure StartLine always has a value; either provided as a parameter or equal to the line where New-*Test/Block where called directly. 

Fix #1673


src/Pester.Runtime.psm1
```diff
@@ -175,6 +175,10 @@ function New-Block {
     $block = $null
     $previousBlock = $state.CurrentBlock
 
+    if (-not $PSBoundParameters.ContainsKey('StartLine')) {
+        $StartLine = $ScriptBlock.StartPosition.StartLine
```
nohwnd, line 179: I think this should be `$MyInvocation.ScriptLineNumber` so we use the location of New-Block the same way we would use the location of Describe. If someone implements their own e.g. MyBlock they will get incorrect value from the start, showing them that they did something wrong, instead of mostly correct value which will break only of they move their scriptblock on the next line (which is the bug we were solving by this).

fflaten: I see your point. In that case. We should probably make it the default value for New-ParameterizedTest as well to cover tests using it directly (none using it with line filter atm, but maybe later). Adding a commit now

nohwnd: Perfect!

nohwnd:
Great job! Btw do you have twitter? DM me there if you have, but DMs are open, and I can then mention you  when tweeting releases.

PR #1680, opened 2020-09-08
Add BeforeDiscovery semantic block

nohwnd:
Add BeforeDiscovery block to allow grouping code that needs to run before discovery.


nohwnd:
- [ ] Add some tests

PR #1698, opened 2020-09-30
Remove pipeline support from Should -Invoke

### LINKED ISSUE #1696: Is this a mock bug?
I discovered a (to me) strange behavior of the Mock function. It might not be a bug, but I figured if I could provide a minimal example it might be worth reporting.

```
BeforeAll {
    function a () { "some output" }
    
    function b () { a }
}

Describe "Is this a bug?" {

    It 'works for mock with empty scriptblock' {
        Mock a { }
        b | Should -Invoke a
    }

    It 'does not work if mock has a non-empty scriptblock' {
        Mock a { "mock" }
        b | Should -Invoke a
    }
}

```
Why does this work in the first It clause, but not in the second? The only difference is the mock there has a non-empty scriptclause. Or am I missing something?
 
The second test case fails with the following message:

```
[-] Is this a bug?.does not work if mock has a non-empty scriptblock 11ms (10ms|1ms)
 Expected Mock to be called at least 1 times but was called 0 times
 at b | Should -Invoke a, D:\Powershell\InvokeCountBug.Tests.ps1:16
 at <ScriptBlock>, D:\Powershell\InvokeCountBug.Tests.ps1:16
Tests completed in 185ms
Tests Passed: 1, Failed: 1, Skipped: 0 NotRun: 0
```

I'm using Pester 5.0.4

fflaten:
Updates `Should -Invoke` to always throw on pipeline-input or when `-ActualValue` is provided. Previously accepted a string through which would override `-CommandName`. Since `-CommandName` is mandatory the pipeline only worked in this confusing scenario:

```
"functionB" | Should -Invoke -CommandName "functionA"
# Checks if 'functionB' was called
```

After this PR, valid usage is:

```
Should -Invoke functionA                # + additional parameters for -Invoke
Should -Invoke -CommandName functionA   # + additional parameters for -Invoke
```

Fix #1696 

- [x] Update invoke to always throw
- [x] Add tests

### DIFF
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -901,12 +901,12 @@ Describe "When Calling Should -Not -Invoke -ExclusiveFilter" {
 
 Describe "When Calling Should -Invoke with pipeline-input or -ActualValue" {
     It "Should throw an error on pipeline-input" {
-        $scriptBlock = { "FunctionUnderTest" | Should -Invoke -CommandName "ABC" -Scope Describe }
+        $scriptBlock = { "value" | Should -Invoke -CommandName "ABC" -Scope Describe }
         $scriptBlock | Should -Throw 'Should -Invoke does not take pipeline input or ActualValue.'
     }
 
     It "Should throw an error on ActualInput-value" {
-        $scriptBlock = { Should -Invoke -CommandName "ABC" -ActualValue "FunctionUnderTest" -Scope Describe }
+        $scriptBlock = { Should -Invoke -CommandName "ABC" -ActualValue "value" -Scope Describe }
         $scriptBlock | Should -Throw 'Should -Invoke does not take pipeline input or ActualValue.'
     }
 }
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -899,6 +899,18 @@ Describe "When Calling Should -Not -Invoke -ExclusiveFilter" {
     }
 }
 
+Describe "When Calling Should -Invoke with pipeline-input or -ActualValue" {
+    It "Should throw an error on pipeline-input" {
+        $scriptBlock = { "FunctionUnderTest" | Should -Invoke -CommandName "ABC" -Scope Describe }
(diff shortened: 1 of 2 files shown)


fflaten:
@nohwnd, I noticed that this logic was added in #1299 . Sure we're not breaking any backward compatibility?

tst/functions/Mock.Tests.ps1
```diff
@@ -899,6 +899,18 @@ Describe "When Calling Should -Not -Invoke -ExclusiveFilter" {
     }
 }
 
+Describe "When Calling Should -Invoke with pipeline-input or -ActualValue" {
+    It "Should throw an error on pipeline-input" {
+        $scriptBlock = { "FunctionUnderTest" | Should -Invoke -CommandName "ABC" -Scope Describe }
```
nohwnd, line 904:
```suggestion
        $scriptBlock = { "value" | Should -Invoke -CommandName "ABC" -Scope Describe }
```

I get where you got the "FunctionUnderTest" from, but it is confusing without having that context.

tst/functions/Mock.Tests.ps1
```diff
@@ -899,6 +899,18 @@ Describe "When Calling Should -Not -Invoke -ExclusiveFilter" {
...
+        $scriptBlock | Should -Throw 'Should -Invoke does not take pipeline input or ActualValue.'
+    }
+
+    It "Should throw an error on ActualInput-value" {
+        $scriptBlock = { Should -Invoke -CommandName "ABC" -ActualValue "FunctionUnderTest" -Scope Describe }
```
nohwnd, line 909:
```suggestion
        $scriptBlock = { Should -Invoke -CommandName "ABC" -ActualValue "value" -Scope Describe }
```

tst/functions/Mock.Tests.ps1
fflaten: Agreed. Had the mock there at first, but removed it to avoid unnecessary processing.

nohwnd:
I looked over the examples in readme and could not find any where I actually used that capability.

Issue #1635, opened 2020-07-23
Pester PR Build Pipeline should check for PsScriptAnalyzer errors


asears:
## 1. General summary of the issue

Running PSScriptAnalyzer against the repo throws the following warnings.

```plaintext
RuleName                            Severity     ScriptName Line  Message
--------                            --------     ---------- ----  -------
PSUseDeclaredVarsMoreThanAssignment Warning      test.ps1   14    The variable 'ErrorView' is assigned but never used.
s
PSUseDeclaredVarsMoreThanAssignment Warning      test.ps1   63    The variable 'PesterPreference' is assigned but never used.
s
PSAvoidUsingWriteHost               Warning      test.ps1   40    File 'test.ps1' uses Write-Host. Avoid using Write-Host
                                                                  because it might not work in all hosts, does not work when
                                                                  there is no host, and (prior to PS 5.0) cannot be
                                                                  suppressed, captured, or redirected. Instead, use
                                                                  Write-Output, Write-Verbose, or Write-Information.
PSAvoidUsingWriteHost               Warning      test.ps1   42    File 'test.ps1' uses Write-Host. Avoid using Write-Host
                                                                  because it might not work in all hosts, does not work when
                                                                  there is no host, and (prior to PS 5.0) cannot be
                                                                  suppressed, captured, or redirected. Instead, use
                                                                  Write-Output, Write-Verbose, or Write-Information.
PSAvoidUsingWriteHost               Warning      test.ps1   53    File 'test.ps1' uses Write-Host. Avoid using Write-Host
                                                                  because it might not work in all hosts, does not work when
                                                                  there is no host, and (prior to PS 5.0) cannot be
                                                                  suppressed, captured, or redirected. Instead, use
                                                                  Write-Output, Write-Verbose, or Write-Information.
PSAvoidUsingCmdletAliases           Warning      test.ps1   28    'foreach' is an alias of 'ForEach-Object'. Alias can
                                                                  introduce possible problems and make scripts hard to
                                                                  maintain. Please consider changing alias to its full
                                                                  content.
PSAvoidUsingCmdletAliases           Warning      Get-Planet 15    'foreach' is an alias of 'ForEach-Object'. Alias can
                                                 .ps1             introduce possible problems and make scripts hard to
                                                                  maintain. Please consider changing alias to its full
                                                                  content.
PSAvoidUsingCmdletAliases           Warning      Get-Planet 17    'where' is an alias of 'Where-Object'. Alias can introduce
[...]

PR #1700, opened 2020-10-01
Add support for VSCode problem-tracking for failed assertions

### LINKED ISSUE #1452: Improve VS code problem matcher
Skipping problem matching in the current implementation of VSCode integration. enable vscode marker and options.

### LINKED ISSUE #1577: V5 not honoring WriteVSCodeMarker option
I've specified the WriteVSCodeMarker issue to work with the PowerShell extension's Pester problem matcher.  That causes Pester failures to show up in the `Problems` window and tests get red squiggles.  But this is no longer working with Pester v5.0.0.

```
Pester version     : 5.0.0 C:\Users\Keith\Documents\WindowsPowerShell\Modules\Pester\5.0.0\Pester.psd1
PowerShell version : 5.1.18362.752
```

## 3. Expected Behavior
I expect this one failure to show up in the `Problems` window because the PowerShell extension's problem matcher was able to find the error long with filename and line info.

## 4.Current Behavior

image

## 5. Possible Solution

Part of the reason Dave (IIRC) added the VSCode marker was that we wanted a text format that didn't change between different versions of Pester.  Also, the VSCode problem matcher is regex driven so it can be a bit of a pain to use.  We found it hard to come up with a problem matcher spec when the error filename/line number occurred on varying lines below the initial error line.  Here is the current pester problem matcher:
```json
      {
        "name": "pester",
        "owner": "powershell",
        "fileLocation": [
          "absolute"
        ],
        "severity": "error",
        "pattern": [
          {
            "regexp": "^\\s*(?:\\[-\\]\\s+)(.*?)(?:\\s+\\d+\\.?\\d*\\s*m?s)\\s*$",
            "message": 1
          },
          {
            "regexp": "^\\s+[Aa]t\\s+([^,]+,)?(.+?):(\\s+line\\s+)?(\\d+)(\\s+char:\\d+)?$",
            "file": 2,
            "line": 4
          }
        ]
      }
```

## 6. Context

I configure tasks.json in VSCode to run all my tests and I bind that to a keyboard shortcut (typically `ctrl+shift+t`.  

My Pester test runner task in my various VSCode projects no longer works like it used to.

fflaten:

VSCode's [problem matcher](https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher) used to report failed Pester-tests in previous versions using a regex-pattern and task.json setup included in the vscode-powershell extension. New format for test output introduced in Pester v5 broke this feature for this version. See powershell/vscode-powershell#2931

This PR and a related PR in vscode-powershell will restore the feature by providing a similar output format to previous versions when using tasks in VSCode.

When test-tasks are configured to continue beyond the first failed assertion (configuration option `[PesterConfiguration]@{Should=@{ErrorAction="Continue"}}`), all failed assertions will be get it's own failed-test output. This way VSCode can detect and show a problem-error per failed assertion which will also highlight every failed Should-line in a test for easier troubleshooting.

Fix #1577 
Fix #1452

## Todos

- [x] Enable problem-compliant output when using VSCode task
- [x] Add tests for output to avoid future problems with vscode-powershell problem matcher
- [x] Create related PR in vscode-powershell (powershell/vscode-powershell#2998)
       - Update problemMatcher-pattern to support new `(UserDuration|FrameworkDuration)` timings in v5
- [x] Remove WriteVSCodeMarker option in PesterConfiguration. Not used in this PR and navigation links are set using ShowNavigationMarkers-option.

### DIFF
--- a/tst/Pester.RSpec.Output.ts.ps1
+++ b/tst/Pester.RSpec.Output.ts.ps1
@@ -43,7 +43,7 @@ function Invoke-PesterInProcess ([ScriptBlock] $ScriptBlock, [ScriptBlock] $Setu
 
 # VSCode-powershell Problem Matcher pattern
 # Storing the original pattern below for easier comparison and maintenance
-$titlePattern = ('^\\s*(?:\\[-\\]\\s+)(.*?)(?:\\s+\\d+\\.?\\d*\\s*m?s)\\s*?(?:\\([\\w\\|]+\\))?\\s*?$' -replace '\\\\', '\')
+$titlePattern = ('^\\s*(?:\\[-\\]\\s+)(.*?)(?:\\s+\\d+\\.?\\d*\\s*m?s)(?:\\s+\\(\\d+\\.?\\d*m?s\\|\\d+\\.?\\d*m?s\\))?\\s*$' -replace '\\\\', '\')
 $atPattern = ('^\\s+[Aa]t\\s+([^,]+,)?(.+?):(\\s+line\\s+)?(\\d+)(\\s+char:\\d+)?$' -replace '\\\\', '\')
 
 
--- a/tst/Pester.RSpec.Output.ts.ps1
+++ b/tst/Pester.RSpec.Output.ts.ps1
@@ -68,8 +68,6 @@ i -PassThru:$PassThru {
             $hostSingleError = $output | Select-String -Pattern 'Single error' -Context 0, 1
             $hostSingleError.Line -match $titlePattern | Verify-True
             $hostSingleError.Context.PostContext[0] -match $atPattern | Verify-True
-
-
         }
 
         t "Matches problem pattern with multiple errors" {
--- a/tst/Pester.RSpec.Output.ts.ps1
+++ b/tst/Pester.RSpec.Output.ts.ps1
@@ -1,0 +1,103 @@
+param ([switch] $PassThru)
+
+Get-Module Pester.Runtime, Pester.Utility, P, Pester, Axiom, Stack | Remove-Module
+
+Import-Module $PSScriptRoot\p.psm1 -DisableNameChecking
+Import-Module $PSScriptRoot\axiom\Axiom.psm1 -DisableNameChecking
+
+& "$PSScriptRoot\..\build.ps1"
+Import-Module $PSScriptRoot\..\bin\Pester.psd1
+
+$global:PesterPreference = @{
+    Debug = @{
+        ShowFullErrors         = $true
+        WriteDebugMessages     = $false
+        WriteDebugMessagesFrom = "Mock"
+        ReturnRawResultObject  = $true
+    }
+    Output = @{
+        Verbosity = "None"
+    }
+}
+$PSDefaultParameterValues = @{}
+
+function Invoke-PesterInProcess ([ScriptBlock] $ScriptBlock, [ScriptBlock] $Setup) {
+    # get the path of the currently loaded Pester to re-import it in the child process
+    $pesterPath = Get-Module Pester | Select-Object -ExpandProperty Path
+    $powershell = Get-Process -Id $pid | Select-Object -ExpandProperty Path
+    # run the test in a separate process to be able to grab all the output
+    $command = {
+        param ($PesterPath, [ScriptBlock] $ScriptBlock, [ScriptBlock] $Setup)
+        Import-Module $PesterPath
+
+        # Modify environment
+        . $Setup
+
+        $container = New-TestContainer -ScriptBlock $ScriptBlock
+        Invoke-Pester -Container $container
+    }.ToString()
+
+    $cmd = "& { $command } -PesterPath ""$PesterPath"" -ScriptBlock { $ScriptBlock } -Setup { $Setup }"
+    & $powershell -NoProfile -ExecutionPolicy Bypass -Command $cmd
+}
+
+# VSCode-powershell Problem Matcher pattern
+# Storing the original pattern below for easier comparison and maintenance
+$titlePattern = ('^\\s*(?:\\[-\\]\\s+)(.*?)(?:\\s+\\d+\\.?\\d*\\s*m?s)\\s*?(?:\\([\\w\\|]+\\))?\\s*?$' -replace '\\\\', '\')
+$atPattern = ('^\\s+[Aa]t\\s+([^,]+,)?(.+?):(\\s+line\\s+)?(\\d+)(\\s+char:\\d+)?$' -replace '\\\\', '\')
+
+
+i -PassThru:$PassThru {
+    b "Output in VSCode mode" {
+        t "Matches problem pattern with single error" {
+            $setup = {
+                $env:TERM_PROGRAM = 'vscode'
+                $psEditor = $null
+            }
+
+            $sb = {
+                Describe 'VSCode Output Test' {
+                    It 'Single error' {
+                        1 | Should -Be 2
+                    }
+                }
+            }
+
+            $output = Invoke-PesterInProcess -ScriptBlock $sb -Setup $setup
+
+            $hostSingleError = $output | Select-String -Pattern 'Single error' -Context 0, 1
+            $hostSingleError.Line -match $titlePattern | Verify-True
+            $hostSingleError.Context.PostContext[0] -match $atPattern | Verify-True
+
+
+        }
+
+        t "Matches problem pattern with multiple errors" {
+            $setup = {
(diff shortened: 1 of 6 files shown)


fflaten:
@nohwnd Got any ideas on how to build tests for output like this? We could mock `Write-Host` to capture the output to a file, but how can we invoke the plugin-step on a few samples tests? Not sure if I can borrow from your `Invoke-PesterInJob`-concept since TestDrive is unique per Pester-run.

nohwnd:
> We could mock Write-Host to capture the output to a file, but how can we invoke the plugin-step on a few samples tests? Not sure if I can borrow from your Invoke-PesterInJob-concept since TestDrive is unique per Pester-run.

Not sure if I understand what you mean here. But yeah, writing P tests (those in .ts.ps1), and invoking the actual Pester tests in a job would be a good way forward. You can capture the output directly by receiving the job output.

fflaten:
Can I mock in P-tests? Got a quick example? Prior to PSv5, you can't redirect or catch the output of `Write-Host` which is used in `(Get-WriteScreenPlugin).EachTestTeardownEnd` to generate the output we want to test. So I figured I'd have to mock `Write-Host` in the scope of `Invoke-Pester`. That way I could catch it (send to output pipeline or a file) so I'd later be able to run the regex-tests. Concept is:

```powershell
Describe "VSCode Problem Matcher Output" {

    BeforeAll {
        <#
        # Some way to catch the output. Only officially supported in PSv5 and above
        $TestFile = 'TestDrive:/VSCodeOutputTest.txt'

        Mock Write-Host {
            $Object | Out-File -FilePath $TestFile -NoNewline:$NoNewline
        }
        #>

        # Storing the original pattern from vscode-powershell below for easier comparison
        $titlePattern = ('^\\s*(?:\\[-\\]\\s+)(.*?)(?:\\s+\\d+\\.?\\d*\\s*m?s)\\s*?(?:\\([\\w\\|]+\\))?\\s*?$' -replace '\\\\','\')
        $atPattern = ('^\\s+[Aa]t\\s+([^,]+,)?(.+?):(\\s+line\\s+)?(\\d+)(\\s+char:\\d+)?$' -replace '\\\\','\')

        # Fake running as VScode task to trigger VSCode Problem matcher output-style
        $org_TERMPROGRAM = $env:TERM_PROGRAM

        $env:TERM_PROGRAM = 'vscode'
        $psEditor = $null

        $sb = {
            Describe 'VSCode Output Test' {
                It 'Single error' {
                    1 | Should -Be 2
                }

                It 'Multiple errors' {
                    1 | Should -Be 2
                    1 | Should -Be 3
                }
            }
        }

        $configuration = [PesterConfiguration]::Default
        $configuration.Run.ScriptBlock = $sb
        $configuration.Run.PassThru = $true
        $configuration.Should.ErrorAction = 'Continue'

        $results = Invoke-Pester -Configuration $configuration
        #Need to catch Write-Host called in Invoke-Pester somehow, lets say into $hostOut
    }

    It "Tests failed as expected" {
        $results.FailedCount | Should -Be 2
    }

    It "Matches test with single error" {
        $hostSingleError = $hostOut | Select-String 'Single error' -Context 0,1

        $hostSingleError.Line | Should -Match $titlePattern
        $hostSingleError.PostContext[0] | Should -Match $atPattern
    }

    It "Matches test with multiple errors" {
        $hostMultipleErrors = $hostOut | Select-String 'Multiple error' -Context 0,1

        $hostOut.Count | Should -Be 2
        
        $hostMultipleErrors[0].Line | Should -Match $titlePattern
        $hostMultipleErrors[0].PostContext[0] | Should -Match $atPattern

        $hostMultipleErrors[1].Line | Should -Match $titlePattern
        $hostMultipleErrors[1].PostContext[0] | Should -Match $atPattern
    }

    AfterAll {
        # Reverse
        $env:TERM_PROGRAM = $org_TERMPROGRAM
        Remove-Variable -Name psEditor -Scope Local
    }

}
```

Maybe I'm overcomplicating and you've got some `Invoke-Test` ++ magic to simplify this 🙂

fflaten:
@nohwnd Omg, I discarded child process as part of jobs, thinking they would behave the same due to the same host (e.g. conhost). Thanks! 👏 Will find some time the next days to update with tests and start QA on new regex-pattern.

Any thoughts on removing `WriteVSCodeMarker` as part of this PR? Or would you like to keep it just in case we revert from env-variables?

nohwnd:
WriteVSCodeMarker <- you are free to remove it everywhere from the code, if it is used somewhere. Just don't remove it from the config object, we should deprecate it somehow.

fflaten:
> WriteVSCodeMarker <- you are free to remove it everywhere from the code, if it is used somewhere. Just don't remove it from the config object, we should deprecate it somehow.

It's never been used in Pester 5.0.0+ besides a simple default-configuration test and a comment-block with placeholder-code for the VSCode output, which is removed in this PR. So not much to deprecate since v5.0 changed the configuration-object and broke old scripts anyways.

Only reason to keep it would be to easy revert to the option-value later or maybe implement "autodetect OR option" for the vscode-output in this PR. We could move the autodetect-logic based inside the assembly to use it for the default option-value. Then people might still use the option to override the default.

nohwnd:
@fflaten K remove it then. I know it did not work, I just did not want to break people who might have used it by accident. But it's probably very low risk.

asears:
Some of these comments are outside of this PR, though these could probably be addressed as part of opening up Output.ps1, either by fixing the issues below or ignoring them in the file.  Others could be issues, let me know if I should open any if important.

1. Script analyzer warnings.

```powershell
invoke-scriptanalyzer .\Output.ps1
```

```plaintext
RuleName                            Severity     ScriptName Line  Message
--------                            --------     ---------- ----  -------
PSReviewUnusedParameter             Warning      Output.ps1 455   The parameter 'Context' has been declared but not used.
PSUseProcessBlockForPipelineCommand Warning      Output.ps1 195   Command accepts pipeline input but has not defined a
                                                                  process block.
```

2. Is PSScriptAnalyzer being run as part of the pipelines?  Could it warn on issues?

3. As part of improving the help, could expand the function comments into proper synopsis comment-based help, ensure each function in this script has help or template for help.

4. I have seen script analyzer complain about Write-Host, should it be deprecated in favour of another method?  I didn't see it complain about this file.  Is it still "considered harmful"?  Would Write-Information or Write-Output be another option?  I understand there's some limitations here due to terminal outputs, mostly curious.

5. Commented code could be removed, put in branch or toggled with feature toggle.  There might be a better alternative to keeping codebase / commented code and TODOs in another location, though getting through code in lost or stale branches was probably a headache in past.  Have seen code scanners come up with this warning regularly with other code bases.  WIP should be on a branch.

6. Data and string outputs should be stored separate of code, localized where possible.

nohwnd:
@asears Are you commenting on the correct PR? I don't see how most of those are relevant to this issue 😵

fflaten:
@asears Those comments belong in #1635.

fflaten:
I believe this is good to go now. WIP PR in vscode-powershell is ready as well, but it's waiting for this to be implemented (in case they want to test).

The tests add 5sec due to creating two powershell sessions. I could replace them by one `Invoke-PesterInProcess` call outside the `i`-block and just focus on output-validation in the tests, but not sure if that's good practice for the P-tests. It would cut 2sec from the run.

tst/Pester.RSpec.Output.ts.ps1
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                }
+            }
+
+            $output = Invoke-PesterInProcess -ScriptBlock $sb -Setup $setup
+
+            $hostSingleError = $output | Select-String -Pattern 'Single error' -Context 0, 1
+            $hostSingleError.Line -match $titlePattern | Verify-True
+            $hostSingleError.Context.PostContext[0] -match $atPattern | Verify-True
+
```
nohwnd, line 71: Nit: extra spaces

nohwnd:
Looks good, do you need me to release this, to be able to test this with VSCode extension?

fflaten:
Not necessary, already tested by manual editing the package.json in the extension and using local Pester-build.
The other PR is pending this being merged just in case the output would be changed further.

If vscode-powershell maintainers wants to test the other PR, they can use local Pester-build and task like:

```
{
    "label": "Test ProblemMatcher v5",
    "type": "shell",
    "command": "Import-Module ./bin/Pester.psd1 -Force; Invoke-Pester ./demoProblemMatcher.tests.ps1",
    "group": "test",
    "problemMatcher": [
        "$pester"
    ]
}
```

PR #1726, opened 2020-10-23
Re-implement New-Fixture in Pester v5

### LINKED ISSUE #1723: New-Fixture generates code that isn't compatible with Pester 5?
I don't know if this is a problem (or already solved), but I found that New-Fixture is generating code that I believe is not compatible with the 5.x version of Pester?

```
$here = Split-Path -Parent $MyInvocation.MyCommand.Path
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.'
. "$here\$sut"

Describe "SomeFunction" {
    It "does something useful" {
        $true | Should Be $false
    }
}
```

Firstly, I guess the setup code should be in a BeforeAll block, and not use $MyInvocation. Secondly, the test fails since it's using legacy syntax on Should.

Pester 5.0.3

## 3. Expected Behavior

I guess it should produce something similar to:

```
BeforeAll {
    . $PSCommandPath.Replace('.Tests.ps1', '.ps1')
}

Describe "SomeFunction" {
    It "does something useful" {
        $true | Should -Be $false
    }
}
```

fflaten:

The `New-Fixture` function available in previous Pester versions was not migrated to Pester v5 in the first releases. This PR re-introduces the feature with the following improvements:
- Pester 5 compliant
- Updated sample/template code to be testable
- Improved parameter validation:
  - Now also strips .psm1 from Name in addition to the .ps1 like previous versions
  - Validates no whitespace in Name as it's not allowed in function names
- A P-test has also been added to validate the generated script and test code (previous todo comment in Pester v4 tests).

Fix #1723

### DIFF
--- a/src/Pester.psd1
+++ b/src/Pester.psd1
@@ -59,13 +59,15 @@
         'ConvertTo-Pester4Result'
 
         # config
-        'New-TestContainer',
+        'New-TestContainer'
 
         # legacy
         'Assert-VerifiableMock'
         'Assert-MockCalled'
         'Set-ItResult'
         'New-MockObject'
+
+        'New-Fixture'
     )
 
     # # Cmdlets to export from this module
--- a/src/Pester.psm1
+++ b/src/Pester.psm1
@@ -35,7 +35,7 @@ Set-Alias 'Get-AssertionOperator' 'Get-ShouldOperator'
     'Get-ShouldOperator'
 
     # config
-    'New-TestContainer',
+    'New-TestContainer'
 
     # export
     'Export-NunitReport'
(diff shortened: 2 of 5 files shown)


nohwnd:
I removed this on purpose.

nohwnd:
But it can be added and deprecated later.

fflaten:
I had a feeling you did, which is why this was mostly copy/paste of Pester v4 version in case we wanted it. Without mentioning the removal in README or keeping it with a warning for a while, users following old blogs and examples might auto-load Pester v3/4 without knowing it get confused when it didn't work like this time..

Thought this was a useful "get started" tool, but I have no problems with adding a deprecation warning immediately if you don't believe it's useful in the long run. Maybe refer to https://pester.dev/docs/quick-start in the warning as a quick-start replacement?

Issue #1740, opened 2020-10-26
Pester crashes if user does not have alias `foreach` defined


ninmonkey:
## 1. General summary of the issue

Invoke Pester crashes if the user's profile contains this:
```powershell
Remove-Alias -Name 'foreach' -Force
```

## 2. Describe Your Environment

```powershell
Pester version     : 5.0.4 C:\Users\cppmo_000\Documents\PowerShell\Modules\Pester\5.0.4\Pester.psd1
PowerShell version : 7.0.3
OS version         : Microsoft Windows NT 10.0.19041.0
```

## 3. Expected Behavior
`Invoke-Pester` should run without the user requiring user aliases.

## 4.Current Behavior

```powershell
Invoke-Pester

System.Management.Automation.CommandNotFoundException: The term 'foreach' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
   at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)    
   at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
   at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
   at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
at Invoke-Pester<End>, C:\Users\user\Documents\PowerShell\Modules\Pester\5.0.4\Pester.psm1: line 4501
at <ScriptBlock>, <No file>: line 
```

## 5. Possible Solution

Running `PSScriptAnalyzer` with `PSAvoidUsingCmdletAliases` to find the `ForEach-Object`s

fflaten:
Shouldn't be using `Foreach-Object` direct either, so double issue here. I've seen a PR (#1640) focusing on some of this cleanup, but it looks to be inactive. Should probably solve some of these in a new PR for now.

nohwnd:
There are also quite a few usages of Where-Object, Select, %, and sort.

fflaten:
I have problems reproducing the error in 5.0.4 (think I got it once), but the line in the stacktrace is updated in pending PR #1745 . Need help to confirm when the PR is reviewed and hopefully approved. 🙂

nohwnd:
[...]

PR #1745, opened 2020-10-28
Cleanup alias and missing SafeCommands

fflaten:
## PR Summary

- Cleanup all code in `/src/*` to avoid usage of alias (related #1740)
- Adds missing `$SafeCommands['xxxx']` on external function-calls in `/src/*`
- Adds a PSScriptAnalyzer configuration for rules with initial rules. Should be tuned later if it's going to be used in CI.
- Adds a custom PSScriptAnalyzer module for build-rules. Only rule atm. is looking for external function-calls that should use `$SafeCommands` (defined in the dictionary)

## PR Checklist

- [x] PR has meaningful title
- [x] All tests pass
    - Run `./build.ps1 -Clean; ./test.ps1 -NoBuild`. Use  a new PowerShell process when C# code is changed.
- [NA] Tests are added/update *(if required)*
- [NA] Documentation is updated/added *(if required)*
    - Should be added for PSSA in the future, but not mandatory or part of CI yet. Rule documented in comment-based help


src/Format.psm1
```diff
@@ -64,7 +65,7 @@ function Format-Hashtable ($Value) {
     $head = '@{'
     $tail = '}'
 
-    $entries = $Value.Keys | sort | foreach {
+    $entries = $Value.Keys | sort | & $SafeCommands['ForEach-Object'] {
```
nohwnd, line 68: Safe usage for sort.
```suggestion
    $entries = $Value.Keys | & $SafeCommands['Sort-Object'] | & $SafeCommands['ForEach-Object'] {
```

src/Format.psm1
```diff
@@ -75,7 +76,7 @@ function Format-Dictionary ($Value) {
     $head = 'Dictionary{'
     $tail = '}'
 
-    $entries = $Value.Keys | sort | foreach {
+    $entries = $Value.Keys | sort | & $SafeCommands['ForEach-Object'] {
```
nohwnd, line 79:
```suggestion
    $entries = $Value.Keys | & $SafeCommands['Sort-Object'] | & $SafeCommands['ForEach-Object'] {
```

src/functions/Mock.ps1
```diff
@@ -1038,6 +1038,7 @@ function Invoke-InMockScope {
 }
 
 function Test-ParameterFilter {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('Pester.BuildAnalyzerRules\Measure-SafeComands', 'Set-StrictMode', Justification = 'Used in mock-wrapper that runs in different context.')]
```
nohwnd, line 1041: We should pass command reference to the Set-Strict mode via parameter and call it from that.

```powershell
# when building mock parameters:
$parameters = @{
     [...]
     Set_StrictMode = $SafeCommands['Set-StrictMode']
}

# Then when we need to call it:
& $______mock_parameters.Set_StrictMode ...
```

src/functions/TestDrive.ps1
```diff
@@ -129,8 +129,8 @@ function Remove-TestDriveSymbolicLinks ([String] $Path) {
     # powershell 2-compatible
     $reparsePoint = [System.IO.FileAttributes]::ReparsePoint
     & $SafeCommands["Get-ChildItem"] -Recurse -Path $Path |
-    where-object { ($_.Attributes -band $reparsePoint) -eq $reparsePoint } |
-    foreach-object { $_.Delete() }
+    & $SafeCommands['where-object'] { ($_.Attributes -band $reparsePoint) -eq $reparsePoint } |
```
nohwnd, line 132: nit: capitalization

nohwnd:
Not part of this change, but if you want to come up with a rule that forces us to not use New-Object, Where-Object, Select-Object and ForEach-Object at all it would be nice to have that in a Draft PR. Those cmdlets are much slower than their normal counter parts. I don't have good replacement for some of the usages (as you saw with the Select-Object -First), but this analyzer rule would allow me to find all of them and fix them. We can probably replace Select-Object -First with call to .NET method that uses Linq. New-Object I already replace with Pester Factory where I create types using .NET constructor.

src/functions/Mock.ps1
nohwnd: You don't have to solve this in this PR. Just pointing out that it is possible.

fflaten:
Could look into it as a warning rule (no autofix-support). 👍
If support for PSv3 was dropped then `.Foreach()` and `.Where()` could've been used. `Select-Option` needs a custom function-replacement I guess.

fflaten:
`sort` is mapped to application in unix. Also found that some rules like PSUseSingularNouns don't run in unix, so note to self: Run PSSA in Windows or both.

src/functions/Mock.ps1
fflaten: Good point. Didn't cause any further failures in tests locally, so added it. 👍 
Now we just need to know why CI fails on PSv3 and PSv4

nohwnd:
> `sort` is mapped to application in unix. 

That is definitely not expected, Sort-Object should be used in all cases.

nohwnd:
> After this cleanup, a simple search for e.g. `(?:new|foreach|select|where)-object` will show all, but a rule could be used for in-editor syntax highlight I guess. Could look into it as a warning rule (no autofix-support). 👍

No pressure, it was just an idea. I should first get my profiler to work properly and start measuring (if only on PS5.1 because that is the only place where it currently works).

> If support for PSv3 was dropped then `.Foreach()` and `.Where()` could've been used. `Select-Option` needs a custom function-replacement I guess.

Having v3 support is important because people still write modules that are compatible down to v3, so they should be able to test them easily.

fflaten:
> > `sort` is mapped to application in unix.
> 
> That is definitely not expected, Sort-Object should be used in all cases.

Yup. Could have given some different results on unix vs windows. Fixed now.

Looks like the CI failure on PSv3 and v4 was that I removed BOM in Mock.ps1. PSSA auto-fix adds BOM, so had to remove manually, but didn't notice that Mock.ps1 had it originally. By design or should all files use BOM for consistency (in a new issue/PR)

nohwnd:
> Non-consistent use of BOM by design or new issue?

New issue.

PR #1753, opened 2020-10-31
Add analyzerrule for highlighting slow *-object commands

fflaten:
## PR Summary

New PSSA analyzer-rule to highlight usage of most `*-Object` built-in cmdlets as they can be slower than other alternatives.
Currently Information-level.

Based on https://github.com/pester/Pester/pull/1745#issuecomment-719455100

## PR Checklist

- [x] PR has meaningful title


fflaten:
I don't see a reason not to merge this. Not sure how useful the rule will be (ex. New-Object seems fast in 5.1), but since it's Information-level it will be ignored by the default settings in `PSScriptAnalyzer.psd1` so it should create noise unless severity is overridden to include Information. 👍

PR #1762, opened 2020-11-01
Remove Bug label from issue

nohwnd:
I know it was my idea to add the label but I find it unnecessarily noisy. Just having Feature and Documentation labels for the other types of issues is enough.


nohwnd:
^^ @fflaten

PR #1790, opened 2020-11-25
Add assembly version check during module import

### LINKED ISSUE #1770: Parameterized tests not working (Check pester assembly version)
## Question

I am trying to run a test that takes in parameters. I have followed this code within Pester/tst/Pester.RSpec.ts.ps1 

```powershell
$container = New-PesterContainer -Path $file -Data @{ Value = 1 }
$r = Invoke-Pester -Container $container -PassThru
```
So my code looks like this (`runPester.ps1`):
```powershell
$container = New-PesterContainer -Path "<absolute-path>\testPester.ps1" -Data @{ environmentName = "earth" }
Invoke-Pester -Container $container -PassThru
```
`testPester.ps1` looks like:
```powershell
param(
    [parameter(mandatory = $true)][string]$environmentName
)

Describe "###### Application" {
    Context "When Accessing the ##### Endpoint" {
        It "Returns Status Code 200" {
            $webInfo = Invoke-WebRequest -Uri "https://###$($environmentName)#####.###/#####"
            $webInfo.StatusCode | Should -Be 200
        }
    }
}
```

Hashes represent sensitive information

When `runPester.ps1` is ran, I get this output:
```
InvalidOperation: C:\Users\####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1:2865:5
Line |
2865 |      $c.Data = $Data
     |      ~~~~~~~~~~~~~~~
     | The property 'Data' cannot be found on this object. Verify that the property exists and can be set.

System.Management.Automation.RuntimeException: The property 'Container' cannot be found on this object. Verify that the property exists and can be set.
at Invoke-Pester<End>, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 4455
at <ScriptBlock>, C:\Users\#####\source\repos\#############\testRunPesterTests.ps1: line 22
```

I also tried running the following:
```
$container = New-PesterContainer -Path "<absolute-path>\testPester.ps1" -Data @{ environmentName = "earth" }
$configuration = [PesterConfiguration]@{
    Run = @{
        Container = $container
        Exit = $true
    }
    TestResult = @{
        Enabled = $true
        OutputPath = "<absolute-path>\TestResults.xml"
    }
}

Invoke-Pester -Configuration $configuration
```
which returned this error:
```
InvalidOperation: C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1:2865:5
Line |
2865 |      $c.Data = $Data
     |      ~~~~~~~~~~~~~~~
     | The property 'Data' cannot be found on this object. Verify that the property exists and can be set.

System.Management.Automation.RuntimeException: The property 'Data' cannot be found on this object. Verify that the property exists and can be set.
   at System.Management.Automation.Interpreter.Interpreter.Run(InterpretedFrame frame)
   at System.Management.Automation.Interpreter.LightLambda.RunVoid1[T0](T0 arg0)
   at System.Management.Automation.PSScriptCmdlet.RunClause(Action`1 clause, Object dollarUnderbar, Object inputToProcess)
   at System.Management.Automation.CommandProcessorBase.Complete()
at New-BlockContainerObject, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 2865
at <ScriptBlock>, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 4731
at Filter-Excluded, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 3348
at Find-File, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 3342
at Invoke-Pester<End>, C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1: line 4731
at <ScriptBlock>, C:\Users\#####\source\repos\#############\testRunPesterTests.ps1: line 20
```

I am less sure of the `[PesterConfiguration]` syntax.

Overall question: Is this correct syntax for running tests with parameters? If not, what should I be doing instead?

**Environment data**

The output of `(Invoke-WebRequest -Uri "https://git.io/JTinj" -UseBasicParsing).Content | Invoke-Expression` is:
```
Pester version     : 5.1.0-rc1 C:\Users\#####\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1
PowerShell version : 7.0.0-rc.3
```

fflaten:
## PR Summary
PowerShell (.NET) doesn't allow loading different version of the same assembly in a process/appdomain. This can give the user a non-terminating error if they import different versions of Pester in the same PowerShell-session. Usually the error can be ignored, but some versions of the Pester module can require changes in the assembly and the user might see weird errors as a result.

The PR adds a minimum assembly version-check during import of the assembly. The requirement is stored as value in the manifest. If the user has the same or newer (until this idea breaks) assembly loaded, the assembly will not be loaded again during import. If the version is older than the requirement, an exception will be thrown to stop the module import, warning about the conflict. Else, import as normal (current behavior). 

Fix #1770

## PR Checklist

- [x] PR has meaningful title


build.ps1
```diff
@@ -92,7 +92,8 @@ foreach ($f in $files) {
 $sb.ToString() | Set-Content $PSScriptRoot/bin/Pester.psm1 -Encoding UTF8
 
 if ($Clean) {
-    dotnet build "$PSScriptRoot/src/csharp/Pester.sln" --configuration Release
+    $manifest = (Get-Content -Path $PSScriptRoot/src/Pester.psd1 -Raw) | Invoke-Expression
+    dotnet build "$PSScriptRoot/src/csharp/Pester.sln" --configuration Release /p:VersionPrefix="$($manifest.ModuleVersion)" "$(if($manifest.PrivateData.PSData.Prerelease) { "/p:VersionSuffix=$($manifest.PrivateData.PSData.Prerelease)" })"
```
nohwnd, line 96: nit: would split that into multiple lines

src/Pester.Types.ps1
```diff
@@ -1,6 +1,15 @@
-if ($PSVersionTable.PSVersion.Major -ge 6) {
-    & $SafeCommands['Add-Type'] -Path "$PSScriptRoot/bin/netstandard2.0/Pester.dll"
-}
-else {
-    & $SafeCommands['Add-Type'] -Path "$PSScriptRoot/bin/net452/Pester.dll"
+# e.g. $minimumVersionRequired = "5.1.0.0" -as [version]
+$minimumVersionRequired = $ExecutionContext.SessionState.Module.PrivateData.RequiredAssemblyVersion -as [version]
```
nohwnd, line 2: We probably need to be more clever than this here to make it work with pre-releases. I typically use beta1, beta2, rc1, rc2 etc... Those can be sorted by `sort` as long as we don't get to beta10 (or whatever pre-release, then it no longer sorts correctly).  I am okay with just Sort and throw on when suffix has more than 1 number. 

Or something like this:
```powershell
function Sort-Version ($Versions) { 
    $vs = foreach ($version in $Versions) {
        $v, $pre = $version -split "-",2

        $r = [PSCustomObject]@{
            String = $version
            Version = [Version] $v
            Char = $null
            Num = $null
        }

        if ($pre) { 
            if ($pre -match "(?<char>[a-z]+)(?<num>\d+)$") {
                $r.Char = $Matches.char
                $r.Num = [int] $Matches.num
            }
            else { 
                $r.Char = $pre
            }
        }

        if (-not $r.Num -and -not $r.Char) { 
            $r.Num = 1
            $r.Char = "z"
        }

        $r
    }

   $vs | Sort-Object -Property Version, Char, Num -Descending | Select-Object -ExpandProperty String
}

function Compare-Version($Left, $Right) {
    if ($Left -eq $Right) { 
        return 0
    }

    $sorted = Sort-Version $left, $right
    if ($sorted[0] -eq $left) { 
        return 1
    }
    else { 
        return -1
    }
}

Sort-Version "5.1.0", "5.1.0-beta", "5.1.0-beta1", "5.1.0-alpha1", "5.1.0-rc10", "5.1.0-rc2" , "5.1.0-rc1", "4.10.1"

Compare-Version "5.1.0" "5.1.0-rc1" # -> 1 left is higher
Compare-Version "5.1.0-rc10" "5.1.0-rc1" # -> 1 left is higher
Compare-Version "4.10.1" "5.1.0-alpha1" # -> -1 left is lower
```

Btw I just realized I forgot to set the dll version in 5.1.0.

fflaten: Pre-releases was ignored intentionally as they're work in progress and something you opt-in too install (with the possible issues that may follow). Don't have anything against including it though, so please add the commit if you'd like 🙂

build.ps1
fflaten: See latest commit. Better?

nohwnd: Specifying the variable as empty has the same effect as not specifying it. I simplified that.

src/Pester.Types.ps1
nohwnd: Yes. Can confirm that. Not using pre-releases is okay.

build.ps1
```diff
@@ -155,24 +159,26 @@ $content = @(
     ,("$PSScriptRoot/src/Pester.psd1", "$PSScriptRoot/bin/")
 )
 
+Copy-Content -Content $content
+
+
 if ($Clean) {
-    $content += @(
+    $manifest = Test-ModuleManifest -Path "$PSScriptRoot/bin/Pester.psd1"
```
nohwnd, line 166: This fails on PS3 with "The wildcard is incorrect"

fflaten:
Well that was weird. Looks like `Test-ModuleManifest` invoked `Import-Dependency` for some absurd reason. 😕

build.ps1
fflaten: Well that was weird. Looks like `Test-ModuleManifest` invoked `Import-Dependency` for some absurd reason. 😕 
Stacktrace?

nohwnd: I don't have any, I can only see it not fail when running on ps7.

fflaten: Reverted to PS3 build with (ugly) workaround in case PS3 build is still required. If it's okay to use PS7 for build then we should use `Test-ModuleManifest` again.

The PS7/vs2019 run only failed due to the psversion-output commit. It made passthru object `$r` from p-tests become arrays which broke the if failed-check after.

nohwnd: We can build on ps7, but I would keep both approaches in case I need to debug something specifically on PS3. Thx. And thanks for helping here, I had no idea what was happening. 🙂

src/Pester.psd1
```diff
@@ -111,11 +111,14 @@
...
-            ReleaseNotes = 'https://github.com/pester/Pester/releases/tag/5.2.0-alpha3'
+            ReleaseNotes = 'https://github.com/pester/Pester/releases/tag/5.2.0-alpha4'
 
             # Prerelease string of this module
-            Prerelease   = 'alpha3'
+            Prerelease   = 'alpha4'
         }
+
+        # Minimum assembly version required
+        RequiredAssemblyVersion = '5.2.0.0'
```
nohwnd, line 121:
```suggestion
        RequiredAssemblyVersion = '5.2.0'
```

I don't see myself ever specifying anything else than the three pieces.

nohwnd:
Other than the nitpick and the possible reversal to use ps7 for build this is imho good to go. I am also okay with keeping it PS3 compatible and building with the existing setup. If nothing the build will be  slightly faster because it is on a deployed agent, and not on a shared (hosted) agent.

src/Pester.psd1
fflaten: It was your own commit, so okay by me 👍

PR #1796, opened 2020-12-01
Fix NUnit2.5 option

### LINKED ISSUE #1794: Nunit2.5 Test Output was recently changed to be invalid and NunitXml is go-forward name
Seems as though Nunit2.5 was recently changed to be disallowed and NunitXml is go-forward test format name.  This was probably long-time coming since Nunit3 is around.

May wish to regress test this change / include the old parameter with deprecation warning until related changes are made to Microsoft repos and some critical external links?  

Here's one place where external dependencies and pipeline examples may break with this change.
https://github.com/PowerShell/vscode-powershell/blob/master/InvokePesterStub.ps1

There seems to be some other changes to the latest release which is preventing tests from running / outputting results.
Will review.

## Describe your environment

Can't run this in an Azure DevOps pipeline due to foreign script link & *.io domain not potentially secure.
Pester 5.1.0

## Expected Behavior

Parameter changes should be deprecated with warnings.

## Current Behavior

Parameter changes broke pipeline.

Add old parameter or notify clearly on deprecation of critical parameters.

nohwnd:
Fix #1794

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,10 @@
 trigger:
   - v5.0
+  - rel/*
+
 pr:
   - v5.0
+  - rel/*
 
 variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -659,7 +659,7 @@ function Invoke-Pester {
         [string] $OutputFile,
 
         [Parameter(ParameterSetName = "Legacy")] # Legacy set for v4 compatibility during migration - deprecated
-        [ValidateSet('NUnitXml', 'JUnitXml')]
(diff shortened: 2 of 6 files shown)


fflaten (approved):
Looks good. Is it worth looking into input-validation for e.g. OutputFormat in the configuration-object (in a new issue)? To break early like we do with ValidateSet in the legacy-parameter.

asears:
I've switched to new NUnitXml format in pipelines we're using and will leave that as-is.

nohwnd:
> Looks good. Is it worth looking into input-validation for e.g. OutputFormat in the configuration-object (in a new issue)? To break early like we do with ValidateSet in the legacy-parameter.

Yes, I intended to do that, but it is not high on my priority list.

nohwnd:
> I've switched to new NUnitXml format in pipelines we're using and will leave that as-is.

Would you expect that to upgrade automatically to NUnit3 once we implement it?

Issue #1616, opened 2020-06-24
LastExitCode modified by an executable that is run as part of a test and then returned by Pester resulting in CI build failure


markwragg:
## 1. General summary of the issue

I have been writing some infrastructure tests and to test that chocolatey was installed I was doing this:
```
It 'Should have installed Chocolatey' {
    { choco } | Should -Not -Throw
}
```
My tests were all passing, but when running in my Packer build were causing the build to fail because the Exit Code was non-zero. I eventually tracked it down to the above test and have discovered that choco sets the `$LastExitCode` variable to 1 even though it has effectively executed successfully (note it only seems to do this when using `choco` on its own, rather than with a specific command).

It seems that because none of the tests had failed, Pester wasn't setting $LastExitCode at all and so it was just being passed through with whatever choco had set it to.

I've worked around this by testing for the install of choco by just looking for the presence of the .exe on disk, but I wondered if this non-handling of `$LastExitCode` when all tests have passed was expected behaviour or perhaps something that should be addressed, because it caused me some confusion.

## 2. Describe Your Environment

Pester version     : 5.0.2 C:\Program Files\WindowsPowerShell\Modules\Pester\5.0.2\Pester.psd1
PowerShell version : 5.1.17763.1007
OS version         : Microsoft Windows NT 10.0.17763.0

## 3. Expected Behavior

I expected `$LastExitCode` to always be 0 if all tests have passed.

## 4.Current Behavior

`$LastExitCode` is set to the exit code of an executable that is run as part of the tests.

## 5. Possible Solution

Have Pester set `$LastExitCode` even if all tests have passed. I am assuming currently it only sets its value if failed.
However there might be reasons not to do this and i'm not sure how much of a breaking change it might be for some use cases. The problem here is more choco than Pester for returning a non-zero exit code for a valid operation. I guess in most cases when an exe returns a non-zero exit code as part of a test run its perfectly valid for that to result in a failure state outside of the environment.

If the consensus is that this shouldn't really be solved, please close this issue but i'm going to raise it anyway as it might help someone who has similar in the future. 

## 6. Context

Pester being used for infrastructure validation tests as part of a Packer VM image build in Azure.

asears:
Not sure if Pester's responsibility to mask $LastExitCode of another process.  Warning might make sense though or something added to detailed diagnostics, I've been caught with this before too and it's hard to find what's causing it.  

Is $LastExitCode 1 considered an exception which Should -Throw?  Probably not?

[...]

PR #1797, opened 2020-12-01
Fix exit

### LINKED ISSUE #1795: New always Exit-behavior breaks scripts using Invoke-Pester
Seems that $configuration.Run.Exit = $true also was implemented as part of Pester 5.1.  When running invoke-pester twice in a Powershell script for 2 different folders, it would only run one folder unless this was set to $false.

_Originally posted by @asears in https://github.com/pester/Pester/issues/1794#issuecomment-735996878

nohwnd:
Fix #1795

### DIFF
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -1137,7 +1137,7 @@ function Invoke-Pester {
                 # just set exit code but don't fail when the option is not set
                 # or when there are no failed tests, to ensure that we can run
                 # multiple successful runs of Invoke-Pester in a row.
-                [System.Environment]::ExitCode = $failedCount
+                [System.Environment]::ExitCode = $global:LASTEXITCODE = $failedCount
             }
         }
         catch {
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -1137,7 +1137,7 @@ function Invoke-Pester {
                 # just set exit code but don't fail when the option is not set
                 # or when there are no failed tests, to ensure that we can run
                 # multiple successful runs of Invoke-Pester in a row.
-                [System.Environment]::ExitCode = $global:LASTEXITCODE = $failedCount
+                $global:LASTEXITCODE = $failedCount
             }
         }
         catch {
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -1126,11 +1126,18 @@ function Invoke-Pester {
 
             # exit with exit code if we fail and even if we succeed, othwerise we could inherit
             # exit code of some other app end exit with it's exit code instead with ours
-            if ($PesterPreference.Run.Exit.Value) {
-                exit ($run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount)
+            $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
+            if ($PesterPreference.Run.Exit.Value -and 0 -ne $failedCount) {
+                # fail with the number of failed tests when there are any
+                # and the exit preference is set. This will fail the run in CI
+                # when any tests failed.
+                exit $failedCount
             }
             else {
-                [System.Environment]::ExitCode = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
+                # just set exit code but don't fail when the option is not set
+                # or when there are no failed tests, to ensure that we can run
+                # multiple successful runs of Invoke-Pester in a row.
+                [System.Environment]::ExitCode = $failedCount
             }
         }
         catch {
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -1128,7 +1128,7 @@ function Invoke-Pester {
             # exit code of some other app end exit with it's exit code instead with ours
             $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
             if ($PesterPreference.Run.Exit.Value -and 0 -ne $failedCount) {
-                # fail with the number of failed tests when there are any
+                # exit with the number of failed tests when there are any
                 # and the exit preference is set. This will fail the run in CI
                 # when any tests failed.
                 exit $failedCount
--- a/tst/Pester.RSpec.InNewProcess.ts.ps1
+++ b/tst/Pester.RSpec.InNewProcess.ts.ps1
@@ -58,4 +58,102 @@ i -PassThru:$PassThru {
             }
         }
     }
+
+    b "Exit codes" {
+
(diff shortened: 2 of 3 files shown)


src/Pester.ps1
```diff
@@ -1126,11 +1126,18 @@ function Invoke-Pester {
 
             # exit with exit code if we fail and even if we succeed, othwerise we could inherit
             # exit code of some other app end exit with it's exit code instead with ours
-            if ($PesterPreference.Run.Exit.Value) {
-                exit ($run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount)
+            $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
+            if ($PesterPreference.Run.Exit.Value -and 0 -ne $failedCount) {
+                # fail with the number of failed tests when there are any
```
fflaten, line 1131:
```suggestion
                # exit with the number of failed tests when there are any
```

src/Pester.ps1
```diff
@@ -1126,11 +1126,18 @@ function Invoke-Pester {
...
+                # and the exit preference is set. This will fail the run in CI
+                # when any tests failed.
+                exit $failedCount
             }
             else {
-                [System.Environment]::ExitCode = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
+                # just set exit code but don't fail when the option is not set
+                # or when there are no failed tests, to ensure that we can run
+                # multiple successful runs of Invoke-Pester in a row.
+                [System.Environment]::ExitCode = $failedCount
```
fflaten, line 1140: This breaks the original issue. I'm not familiar with `[System.Environment]::ExitCode`, but it doesn't affect `$LASTEXITCODE` which we want to reset. AFAIK we need to either run an application that returns 0 or use `$global:LASTEXITCODE = 0` to reset it in the else-block (or during a teardown-step at the end of each container?).

Demo of original issue (#1616) :
```
$c = [PesterConfiguration]::Default
$c.Run.Scriptblock = {
    Describe "LastExitCode Demos" {
        It 'Works since next legacy app reset the lastexitCode' {
            { cmd /c exit /b 1 } | Should -Not -Throw
            $LASTEXITCODE | Should -Be 1
        }
    }
}
#$c.Run.Exit = $true

Write-Host "Running Tests"
Invoke-Pester -Configuration $c

Write-Host "LastExitCode: $($LASTEXITCODE)"
Write-Host "EnvironmentExitCode: $([Environment]::ExitCode)"
```

fflaten (commented):
No go unfortunately.
We should probably also add some tests for this as it's hard to remember all scenarios, 🙂

nohwnd:
@fflaten should be better.

src/Pester.ps1
nohwnd: implemented

tst/Pester.RSpec.InNewProcess.ts.ps1
```diff
@@ -58,4 +58,102 @@ i -PassThru:$PassThru {
...
+            $testpath = Join-Path $temp "$([Guid]::NewGuid().Guid).tests.ps1"
+            $powershell = (Get-Process -Id $pid).Path
+
+            try {
+                $c = "
+                Describe 'd' {
+                    It 'i' {
+                        # an executable exits with 9999 (we use powershell as the executable, because we know it will work cross platform)
+                        & '$powershell' -Command { exit 9999 }
+                        `$LASTEXITCODE | Should -Be 9999
```
fflaten, line 75:
```suggestion
                        # an executable exits with 99 (we use powershell as the executable, because we know it will work cross platform)
                        & '$powershell' -Command { exit 99 }
                        `$LASTEXITCODE | Should -Be 99
```
Unix exitcodes goes from 0-255. Resets to 0 at every 256.
```
& '/usr/share/powershell/pwsh' -Command { exit 256 }; $LASTEXITCODE
0

#This is the exit code return in the current sample
9999%256
15
```

tst/Pester.RSpec.InNewProcess.ts.ps1
```diff
@@ -58,4 +58,102 @@ i -PassThru:$PassThru {
...
+                }"
+                Set-Content -Path $testpath -Value $c
+
+                $sb = [scriptblock]::Create("
+                try {
+                    Invoke-Pester -Path $testpath -EnableExit
+                    `$exitCode = `$LASTEXITCODE
+                }
+                finally {
+                    # exitcode was set to 9999 in the test because the test passed,
```
fflaten, line 86:
```suggestion
                    # exitcode was set to 99 in the test because the test passed,
```
Same as above

tst/Pester.RSpec.InNewProcess.ts.ps1
```diff
@@ -58,4 +58,102 @@ i -PassThru:$PassThru {
...
+            $testpath = Join-Path $temp "$([Guid]::NewGuid().Guid).tests.ps1"
+            $powershell = (Get-Process -Id $pid).Path
+
+            try {
+                $c = "
+                Describe 'd' {
+                    It 'i' {
+                        # an executable exits with 9999 (we use powershell as the executable, because we know it will work cross platform)
+                        # we use this to fail the test
+                        & '$powershell' -Command { exit 9999 }
```
fflaten, line 126:
```suggestion
                        # an executable exits with 99 (we use powershell as the executable, because we know it will work cross platform)
                        # we use this to fail the test
                        & '$powershell' -Command { exit 99 }
```

tst/Pester.RSpec.InNewProcess.ts.ps1
```diff
@@ -58,4 +58,102 @@ i -PassThru:$PassThru {
...
+                }"
+                Set-Content -Path $testpath -Value $c
+
+                $sb = [scriptblock]::Create("
+                try {
+                    Invoke-Pester -Path $testpath -EnableExit
+                    `$codeAfterPester = `$true
+                }
+                finally {
+                    # exitcode was set to 9999 because one test failed in the test
```
fflaten, line 138:
```suggestion
                    # exitcode was set to 99 because one test failed in the test
```

fflaten (changes requested):
@nohwnd Looks good - just need to reduce the sample exitcode for Unix-compatibility

fflaten:
Btw. when would you use `[System.Environment]::ExitCode`? Not a request to remove it, just never had a reason to use it yet.

nohwnd:
IDK actually, I removed it to avoid code that is not needed. I could not come up with any way to make it propagate from powershell session into the calling process.

PR #1802, opened 2020-12-02
Fix warnings in tst/axiom

asears:
## PR Summary

This fixes the warnings in tst/axiom.  Introduces PROCESS to address the recommendation.  Includes a suppression of safecommand in a test with Out-String.  

WIP #1635 

## PR Checklist

- [X] PR has meaningful title

### DIFF
--- a/tst/axiom/Verify-AssertionFailed.ps1
+++ b/tst/axiom/Verify-AssertionFailed.ps1
@@ -1,28 +1,30 @@
 function Verify-AssertionFailed {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("Pester.BuildAnalyzerRules\Measure-SafeCommands", "")]
     param (
         [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
         [ScriptBlock]$ScriptBlock
     )
+    PROCESS {
+        $assertionExceptionThrown = $false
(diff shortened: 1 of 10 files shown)


nohwnd:
I think the default block is end, not process.

fflaten:
Yeah, I believe it's end (though end == begin either way AFAIK).

process is the default for filters: `filter test { "do something $_" }; 1..4 | test`
Using process will change the behavior of these functions if an array is passed to them as it will execute the block once per item. Ex:

```
function verify-demoProcess {
    param (
        [Parameter(ValueFromPipeline = $true)]
        $Actual
    )
    process {
	"The value is: $_"
	"Actual is: $Actual"
    }
}

function verify-demoEnd {
    param (
        [Parameter(ValueFromPipeline = $true)]
        $Actual
    )
    end {
	"The value is: $_"
	"Actual is: $Actual"
    }
}

> 1..4 | verify-demoProcess
The value is: 1
Actual is: 1
The value is: 2
Actual is: 2
The value is: 3
Actual is: 3
The value is: 4
Actual is: 4

> 1..4 | verify-demoEnd
The value is:
Actual is: 4
```

asears:
Updated to END.  We could also just ignore the warning if there are any problems with this approach.

Issue #1603, opened 2020-06-16
v5: InModuleScope fails to pass values from the current scope without explicit -Parameters


vexx32:
## 1. General summary of the issue

In V4, this worked:
```ps1
It 'replaces wildcard characters with regex equivalent' -TestCases @(
    @{ Pattern = 'About*son'; Expected = '^About.*son$' }
    @{ Pattern = 'AboutArrays*'; Expected = '^AboutArrays' }
    @{ Pattern = '*Arrays'; Expected = 'Arrays$' }
    @{ Pattern = '*Array*'; Expected = 'Array' }
) {
    param($Pattern, $Expected)
    InModuleScope 'PSKoans' { ConvertFrom-WildcardPattern $Pattern } | Should -Be $Expected
}
```

My attempt at converting to V5 was:
```ps1
It 'replaces wildcard characters with regex equivalent' -TestCases @(
    @{ Pattern = 'About*son'; Expected = '^About.*son$' }
    @{ Pattern = 'AboutArrays*'; Expected = '^AboutArrays' }
    @{ Pattern = '*Arrays'; Expected = 'Arrays$' }
    @{ Pattern = '*Array*'; Expected = 'Array' }
) {
    InModuleScope 'PSKoans' { ConvertFrom-WildcardPattern $Pattern } | Should -Be $Expected
}
```

This **does not work**. The following _does_ work:

```ps1
It 'replaces wildcard characters with regex equivalent' -TestCases @(
    @{ Pattern = 'About*son'; Expected = '^About.*son$' }
    @{ Pattern = 'AboutArrays*'; Expected = '^AboutArrays' }
    @{ Pattern = '*Arrays'; Expected = 'Arrays$' }
    @{ Pattern = '*Array*'; Expected = 'Array' }
) {
    InModuleScope 'PSKoans' -Parameters @{ Pattern = $Pattern } {
        param($Pattern)
        ConvertFrom-WildcardPattern $Pattern
    } | Should -Be $Expected
}
```

**Also**... it seems `-ArgumentList` does not work either. I tried this way with no success:

```ps1
It 'replaces wildcard characters with regex equivalent' -TestCases @(
    @{ Pattern = 'About*son'; Expected = '^About.*son$' }
    @{ Pattern = 'AboutArrays*'; Expected = '^AboutArrays' }
    @{ Pattern = '*Arrays'; Expected = 'Arrays$' }
    @{ Pattern = '*Array*'; Expected = 'Array' }
) {
[...]

PR #1812, opened 2020-12-19
Fix default InModuleScope args-value

### LINKED ISSUE #1809: InModuleScope: Cannot use [Parameter()] for the parameter inside the scriptblock
This is similar to issue #1603, but not entirely.

When using `[Parameter()]` on the parameter declaration inside a InModuleScope it fails.

## Describe your environment
```
Pester version     : 5.1.1 C:\source\SqlServerDsc\output\RequiredModules\Pester\5.1.1\Pester.psm1
PowerShell version : 7.1.0
```

## Steps to reproduce
```powershell
Describe 'Something' {
    Context 'When the desired protocol is Named Pipes' {
        BeforeAll {
            $mockPipeName = '\\.\pipe\$$\TESTCLU01A\MSSQL$SQL2014\sql\query'
        }

        It 'Should return the correct values' {
            $inModuleScopeParameters = @{
                MockPipeName = $mockPipeName
            }

            InModuleScope -ModuleName $script:dscResourceName -Parameters $inModuleScopeParameters -ScriptBlock {
                param
                (
                    [Parameter()]
                    [System.String]
                    $MockPipeName
                )
            }
        }
    }
}
```

## Expected Behavior
Not to fail.

## Current Behavior
It fails with the following error:
``` 
[-] Something.When the desired protocol is Named Pipes.Should return the correct values 8ms (7ms|1ms)
 ParameterBindingException: A positional parameter cannot be found that accepts argument '$null'.
```

Removing the `[Parameter()]` solves the problem. This need to be documented if Pester will never support using `[Parameter()]`.

fflaten:
## PR Summary
Using ParameterAttribute or CmdLetBinding in InModuleScope scriptblock causes ParameterBindingException for the default when no arguments are provided. $ArgumentList-parameter is default null and is always passed to the scriptblock, which throws the error since advanced functions don't implicitly allow arguments.

PR changes $ArgumentList to be empty array by default as it won't cause the exception when not defined. Similar to #1730 for Mocks.

Fix #1809 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -71,7 +71,7 @@ function InModuleScope {
         [HashTable]
         $Parameters,
 
-        $ArgumentList
+        $ArgumentList = @()
     )
 
     $module = Get-ScriptModule -ModuleName $ModuleName -ErrorAction Stop
--- a/tst/functions/InModuleScope.Tests.ps1
+++ b/tst/functions/InModuleScope.Tests.ps1
@@ -82,3 +82,61 @@ Describe "Get-ScriptModule behavior" {
     }
 
 }
+
+Describe 'InModuleScope parameter binding' {
+    # do not put this into BeforeAll this needs to be imported before calling InModuleScope
+    # that is below, because it requires the module to be loaded
+
+    Get-Module TestModule2 | Remove-Module
+    New-Module -Name TestModule2 { } | Import-Module -Force
+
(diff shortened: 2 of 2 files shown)


JustinGrote:
Is this possibly related to #1603?

fflaten:
No, it won't add any parameters for InModuleScope scriptblock automatically if that's what you mean. 

This PR is just a fix to support an advanced functions/script as the InModuleScope scriptblock.

PR #1865, opened 2021-03-07
Bunch bps and one hit bps

nohwnd:
## PR Summary
Add SingleHitBreakpoints and DelayWritingBreakpoints experimental options to coverage. They will be renamed later probably. 

Some extra messages are printed to screen that will be removed later.


nohwnd:
In my measurements, running all my Pester tests runs ~40s without Code Coverage and ~300s with Code Coverage, which is 7 times more. Code Coverage sets around 7k breakpoints for my codebase.

With the fix, it runs ~40s without CC and ~42s with CC, including all the overhead of setting up breakpoints, calculating and printing the coverage report, so the execution is probably <1% slower with 7000 breakpoints enabled.

PR #1866, opened 2021-03-07
Little improvements to coverage output

nohwnd:
Improve output, add percentage target so we can print green or red coverage message. Disable outputting missed commands in non-detailed output, but keep it in the result object.


nohwnd:
When I lower the  setting enough to 1.6%: `$script:PesterPreference.codecoverage.CoveragePercentTarget = 1.6`
image


And with the default of 75 % that I don't meet with this setup:
image

PR #1870, opened 2021-03-08
Enable only discovery

### LINKED ISSUE #1609: v5: Provide Invoke-Discovery function to retrieve test metadata/counts without running test code

In PSKoans we have a need to give users a progress counter, and to do this we need to be able to count the number of tests that there are in total. Up until now, I've been using an AST-based counter, but this method has limitations; it can't count -TestCases well for the majority of circumstances, and will never be able to reliably count tests if for some reason we have a file generating tests in a loop in some fashion, or if tests need to be conditionally skipped/omitted due to platform constraints.

## 4. Current Behavior

There is currently no supported way to get the discovery-time metadata. I'm currently using an adapted version of a code snippet provided by @nohwnd that does work quite well -- but only in practice:

```ps1
$Result = & (Get-Module Pester) {
    [CmdletBinding()]
    param(
        $Path,
        $ExcludePath,
        $SessionState
    )

    $_Pester_State_Backup = $state.PSObject.Copy()
    $state.Stack = [System.Collections.Stack]@()
    try {
        Reset-TestSuiteState

        # to avoid Describe thinking that we run in interactive mode
        $invokedViaInvokePester = $true

        $fileList = Find-File -Path $Path -ExcludePath $ExcludePath -Extension '.Koans.ps1'
        $containers = foreach ($file in $fileList) {
            New-BlockContainerObject -File (Get-Item $file)
        }

        Find-Test -BlockContainer $containers -SessionState $SessionState
    }
    finally {
        $state = $_Pester_State_Backup
        Remove-Variable -Name _Pester_State_Backup
    }
} -Path $KoanInfo.Path -SessionState $PSCmdlet.SessionState
```

The limitation here is that if you attempt to invoke Pester to _test_ a function containing this code, this code corrupts the internal Pester state, interrupting the ongoing test. I've observed:

- Mock table being empty / null
- Timers reporting negative times
- Test output failing to actually list errors as they occur, though the number of failed tests is (sometimes) still reported.

## 5. Possible Solution

Provide a (preferably completely stateless) discovery-only function for users to get this kind of metadata in advance of any kind of test run. I personally have no need to be able to reuse this for a test run, I just need the test counts ultimately, which I can get from this data. I'm sure someone will probably want that at some point though, so maybe having a way to put that back through to Invoke-Pester would be a good idea too.

I'd suggest a name like `Get-DiscoveryData` or `Invoke-Discovery` perhaps.

nohwnd:
Enable the run phase to be skipped. This will report only discovered tests and all will look failed.

Fix #1609


nohwnd:
@vexx32  please give it a try to see if this fits your needs. It prints some extra info, but you can suppress that using -Output None. Simply define $global:DiscoveryOnly=$true before your discovery run, and use -PassThru to get the object from Invoke-Pester. Once you confirm it works we can start polishing it. Imho it's easier to have this as an option on .Run, because giving it it's own cmdlet seems like a lot of duplication for little benefit at the moment. It might evolve to more fully fledged solution later, but for that we would need to be able to run tests from the provided result object, and I am dumping all the internal info from it before return so this is not going to happen any time soon.  And would be only useful in a handful of scenarios.

vexx32:
Awesome, thank you! I'll have a look at it this evening. 💖

nohwnd:
@vexx32  ping ping.

vexx32:
@nohwnd this looks like it works very well!

Only a couple notes, I think:

- It registering as Failed strikes me as somewhat confusing. Perhaps it'd make more sense to default them to Skipped or just blank?
- I think an option in the configuration object to set this up works just fine, that should be pretty easy to work with.

Thanks for putting this together! 💖

nohwnd:
Thanks. I made them all NotRun. Added config object property called Run.SkipRun.

PR #1880, opened 2021-03-21
Useful string assert message

### LINKED ISSUE #1313: String comparison message is not useful
In #532 the output of the string was changed to show only a short excerpt, this excerpt is too short that it is hard to tell where exactly the difference is, and the output is not useful when reviewing message that holds interesting data such as when comparing expected error messages.

I think that a better approach would be to return to the previous functionality combined with a much longer excerpt. 

The host size could be used to determine how wide the output should be, or a sensible default of 80 characters. 

As a first step I would make the excerpt longer to make it a bit more useful.

nohwnd:
Fix  #1313

### DIFF
--- a/src/functions/assertions/Be.ps1
+++ b/src/functions/assertions/Be.ps1
@@ -150,9 +150,21 @@ function Get-CompareStringMessage {
         [AllowEmptyString()]
         [String]$Actual,
         [switch]$CaseSensitive,
-        $Because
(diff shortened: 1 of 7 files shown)


nohwnd:
Notice we have difference at index 985 out of 4998: 

Wide window: 

image
image

Narrower window: 

image



Narrow window: 

image

PR #1882, opened 2021-03-21
Exclude noisy Measure-SafeCommands rule by default

### LINKED ISSUE #1761: Reduce analyzer for unsafe command call to src/*
@fflaten  please reduce the scope of the new unsafe-call rule to src only. I am getting ton of those in tests: 

image

fflaten:
## PR Summary
The custom "Measure-SafeCommands" PSScriptAnalyzer rule is noisy when editing files in VSCode outside of "src\\*" where it shouldn't apply.

There currently isn't a way to limit the paths where VSCode use PSSA automatically, so the rule is excluded in "PSScriptAnalyzerSettings.psd1" for now. That way, we can still benefit from automatic checks against other useful default rules.

The SafeCommands-rule can still be invoked manually using `Invoke-ScriptAnalyzer -Path ./src -CustomRulePath ./Pester.BuildAnalyzerRules/ -Recurse`

Fix #1761 

## PR Checklist

- [X] PR has meaningful title

### DIFF
--- a/PSScriptAnalyzerSettings.psd1
+++ b/PSScriptAnalyzerSettings.psd1
@@ -5,5 +5,6 @@
     ExcludeRules=@(
         'PSUseShouldProcessForStateChangingFunctions'
         'PSUseApprovedVerbs'
+        'Measure-SafeCommands'
     )
 }


nohwnd:
To be honest I don't even pay attention to the squiggly lines because of this. Could this be added as continue if failed CI step so I can get it to a good state over time?

nohwnd:
Merged it as is. I don't want to overload this PR.

nohwnd:
The existing azure pipeline please.

PR #1884, opened 2021-03-24
CC using Measure-Script

nohwnd:
Draft of using powershell profiler for Code Coverage


nohwnd:
Done this using my own profiler code, this is possible if they ever release the built in profiler, but not strictly necessary.

PR #1887, opened 2021-03-28
Coverage gutters

nohwnd:
Add a coverage format specific to Coverage Gutters, to make it easy to use with Pester in VSCode and also fix some of the problems with it. Now you can specify whatever path your want covered, including the path to which coverage.xml is output, and not just a subfolder. The root is determined based on the location of coverage.xml not the common root of files, so it is static and not dynamic. 

Building on this article: https://web.archive.org/web/20201107233838/https://www.pwsh.site/powershell/2019/01/10/how-to-enable-coverage-markings-in-vscode-for-your-powershell-projects.html

Integration with the PowerShell extension Code Lens in VSCode is not done yet, because the pester integration does not have the ability to provide a custom setup script, but can be either done by a custom launch json task, which points to a script. 

```json
# .vscode/launch.json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Run tests in current file with CC",
            "type": "PowerShell",
            "request": "launch",
            "script": "${workspaceFolder}/test.ps1",
            "args": [
                "-Path", "'${file}'",
                "-Output", "${config:powershell.pester.debugOutputVerbosity}",
                "-CodeCoverage",
            ],
            "cwd": "${file}",
        }
    ]
}
```
```powershell 
# test.ps1
param (
    [String] $Path = ".",
    [String] $Output = "Normal",
    [Switch] $CodeCoverage
)

$Configuration = [PesterConfiguration]::Default

$Configuration.Run.Path = $Path

$Configuration.Output.Verbosity = $Output

$Configuration.CodeCoverage.Enabled = [bool] $CodeCoverage
# CoverageGutters is new option in Pester 5.2 pre-release
$Configuration.CodeCoverage.OutputFormat = "CoverageGutters"
$Configuration.CodeCoverage.Path = "$PSScriptRoot/src"
$Configuration.CodeCoverage.OutputPath = "$PSScriptRoot/coverage.xml"
$Configuration.CodeCoverage.CoveragePercentTarget = 90

$Configuration.Debug.WriteDebugMessages = $true
$Configuration.Debug.WriteDebugMessagesFrom = "CodeCoverage"


# for convenience just run the passed file when we invoke this via launchConfig.json when
# the file is a non-test file
$isDirectory = (Get-Item $Path).PSIsContainer
$isTestFile = $Path.EndsWith($Configuration.Run.TestExtension.Value)
if (-not $isDirectory -and -not $isTestFile) {
    & $Path
    return
}

Invoke-Pester -Configuration $Configuration
```

```json
# .vscode/settings.json
{
    // disable gutters to make breakpoints usable
    "coverage-gutters.showGutterCoverage": false,
    "coverage-gutters.showLineCoverage": true
}
```
Ensure that you have [CoverageGutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) extension installed VSCode, and run "Coverage Gutters: Watch". 

( You can also integrate with the code lens, but it's fragile. You can set `$script:PesterPreference = $Configuration` from test.ps1, which would be then used while it is in scope. But because not all files are run when you use the code lens it will produce confusing code coverage results. )

### DIFF
--- a/src/Pester.ps1
+++ b/src/Pester.ps1
@@ -1127,8 +1127,43 @@ function Invoke-Pester {
                 $breakpoints = @($run.PluginData.Coverage.CommandCoverage)
                 $coverageReport = Get-CoverageReport -CommandCoverage $breakpoints
                 $totalMilliseconds = $run.Duration.TotalMilliseconds
-                $jaCoCoReport = Get-JaCoCoReportXml -CommandCoverage $breakpoints -TotalMilliseconds $totalMilliseconds -CoverageReport $coverageReport
-                $jaCoCoReport | & $SafeCommands['Out-File'] $PesterPreference.CodeCoverage.OutputPath.Value -Encoding $PesterPreference.CodeCoverage.OutputEncoding.Value
+                [xml] $jaCoCoReport = [xml] (Get-JaCoCoReportXml -CommandCoverage $breakpoints -TotalMilliseconds $totalMilliseconds -CoverageReport $coverageReport)
+
+                $settings = [Xml.XmlWriterSettings] @{
+                    Indent              = $true
+                    NewLineOnAttributes = $false
+                }
+
+
+                $stringWriter = $null
+                $xmlWriter = $null
+                try {
+                    $stringWriter = [Pester.Factory]::CreateStringWriter()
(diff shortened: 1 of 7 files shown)


nohwnd:
cc

PR #1893, opened 2021-04-06
Fix output for currently set fullname-filter

### LINKED ISSUE #1892: FullName filter discovery output always shows "System.String[]"
A seperate issue I know but it would be good if we could also fix the purple text output when providing a FullName string/array. It always shows "Filter 'FullName' set to ('System.String[]')." and doesn't show the actual values. Tag works ok as you can see below

``` powershell
Filter 'FullName' set to ('System.String[]').
Filter 'Tag' set to ('LOAD_FROM').
```

_Originally posted by @o-o00o-o in https://github.com/pester/Pester/issues/1891#issuecomment-813678302_

fflaten:
## PR Summary
Fixes broken output for currently set FullName-filter. Caused by a cast to `string[][]` which seems to be a typo.

Fix #1892 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -2266,7 +2266,7 @@ function flattenBlock ($Block, $Accumulator) {
 function New-FilterObject {
     [CmdletBinding()]
     param (
-        [String[][]] $FullName,
+        [String[]] $FullName,
         [String[]] $Tag,
         [String[]] $ExcludeTag,
         [String[]] $Line


nohwnd:
I think the fullname filter was meant to be an array of fullnames to the test, with possible wildcards. That is `@("describe1.describe2.test1", "describe45.context1.test20")` so you can copy paste it from the normal output and use it without providing arrays in arrays. This was meant for interactive use and has it's limits, because it might not pin-point just a single test in some cases, but instead should be easier to user interactively.

I think there were tests for this, weird that this typing change does not break any of them, maybe it's because of the way the array is expanded?

fflaten:
> I think there were tests for this, weird that this typing change does not break any of them, maybe it's because of the way the array is expanded?

Yeah, casting a string-array to `[string[][]]` places each value in their own inner array. So when we looped this array and performed a `string -like Array[1]` the implicit cast back to string saved us. Pure luck 😄 

```
PS /workspaces/Pester/Samples> $t = [string[]]("l","b"); $t.Count; $t[0].GetType().FullName
2
System.String

PS /workspaces/Pester/Samples> $l = [string[][]]$t; $l.Gettype().FullName;$l.Count; $l[0].GetType().FullName;$l[0].Count
System.String[][]
2
System.String[]
1
```

PR #1926, opened 2021-04-29
Generate config docs

### LINKED ISSUE #1922: Configuration strings should be extracted and injected into inline help of New-Pester configuration
Configuration strings should be extracted and injected into inline help of New-Pester configuration, to keep them correct, and to "automatically" add new options.

nohwnd:
Generate config section from the object. 

Fix #1922

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -47,7 +47,6 @@ param (
     [switch] $Inline
 )
 
-"PSVersion: $($PSVersionTable.PSVersion)"
 $ErrorActionPreference = 'Stop'
 Get-Module Pester | Remove-Module
 if ($Clean -and (Test-Path "$PSScriptRoot/bin")) {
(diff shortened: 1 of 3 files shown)


fflaten:
Maybe remove the same PesterConfiuration-overview in `Invoke-Peater` help and reference this function?

nohwnd:
@fflaten  no idea what I broke this time. A second pair of eyes would be very helpful :)

build.ps1
```diff
@@ -53,6 +53,125 @@ if ($Clean -and (Test-Path "$PSScriptRoot/bin")) {
...
+Copy-Content -Content $content
+
+
+# update help for New-PesterConfiguration
+if ($PSVersionTable.PSVersion.Major -gt 5) {
+    Import-Module "$PSScriptRoot/bin/bin/netstandard2.0/Pester.dll"
+}
+else {
+    Import-Module "$PSScriptRoot/bin/bin/net452/Pester.dll"
+}
```
fflaten, line 104: @nohwnd This causes the failures. Needs cleanup after as it causes the dll to be shown as a binary module in all p-tests since they invoke `& "$PSScriptRoot\..\build.ps1"` at the beginning. So later when you build the path for the schema, you get paths for both "modules":

```
ModuleType Version PreRelease Name   ExportedCommands
---------- ------- ---------- ----   ----------------
Script     5.2.0   rc1        Pester {Add-ShouldOperator, AfterAll, AfterEach, Assert-MockCalled…}
Binary     5.2.0.0            Pester 
```

nohwnd: Ah! Need to run the doc generation in Job in that case (probably better but might be slow). Or load it via assembly.loadfrom. Thanks!

fflaten: Do jobs have their own appdomain? If not, then the assembly will be loaded in the powershell session either way, so could just do `Remove-Module`  after doc generation to avoid the module-metadata conflict later in the tests. Or would e.g. a job give other benefits?

PR #1950, opened 2021-05-17
Save all PowerShell files with UTF8BOM

### LINKED ISSUE #1747: Inconsistent use of BOM in UTF8 files
## Question
Pester uses BOM inconsistently in files with UTF8. Only some files, like `Mock.ps1`, `Mock.Tests.ps1` etc. use BOM, while the rest use UTF8 without BOM.

CI failed for PS v3/v4 in #1745 after BOM was removed in `Mock.ps1` by mistake. It caused mock-test `Command with single quote in name should be mockable` to fail due to use of smart-quotes AFAIK.

As BOM seems to it's required for older PowerShell-versions on Windows, should all files have BOM?
Or is the use of smart-quotes in the test itself the issue? Copy-paste from mail/word-document? 

**Environment data**

N/A.
Pester 5.1.0-beta3+

nohwnd:
Use utf8 with bom for all files and add setting to make it the default

Fix #1747

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,8 @@
   "editor.formatOnSave": true,
   "powershell.codeFormatting.addWhitespaceAroundPipe": true,
   "powershell.codeFormatting.pipelineIndentationStyle": "IncreaseIndentationForFirstPipeline",
-  "yaml.format.singleQuote": true
+  "yaml.format.singleQuote": true,
+  "[powershell]": {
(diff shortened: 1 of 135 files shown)


fflaten:
As long as the tests passes, I'm good. That's what sparked the issue in the first place.
Can take a quick look later tonight if needed. 👍

nohwnd:
All of them just have the same edit. Should be good.

PR #1955, opened 2021-05-20
Use profiler to improve perf of test drive, name expanding and add-member

nohwnd:


nohwnd:
image

This is how it looks after the changes, can't change those pieces without disabling TestDrive. I also think I changed the described behavior of test drive to setup new one for each top-level describe as it was in v4. Will revert that to the described behavior of one test drive per file.

nohwnd:
Reverted that was easier than expected. And TestDrive no longer shows up that much: 

image

PR #1957, opened 2021-05-23
Make InModuleScope-parameters available as variables in scriptblock

### LINKED ISSUE #1603: v5: InModuleScope fails to pass values from the current scope without explicit -Parameters

fflaten:
`InModuleScope` currently requires param-block to make parameters available by name (if not, only available as arguments). This PR add parameters as dynamic variables inside the scriptblock similar to how it works with mocks.

Fix #1603 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -75,26 +75,52 @@
     )
 
     $module = Get-ScriptModule -ModuleName $ModuleName -ErrorAction Stop
+    $sessionState = Set-SessionStateHint -PassThru -Hint "Module - $($module.Name)" -SessionState $module.SessionState
 
-    # TODO: could this simply be $PSCmdlet.SessionState? Because the original scope we are moving from
-    # is the scope in which this command is running, right?
-    # $originalState = $Pester.SessionState
-    # $originalScriptBlockScope = Get-ScriptBlockScope -ScriptBlock $ScriptBlock
+    $wrapper = {
+        param (
+            [Parameter(Mandatory = $true)]
+            [scriptblock]
+            ${Script Block},
 
-    # try {
-    # $sessionState = Set-SessionStateHint -PassThru -Hint "Module - $($module.Name)" -SessionState $module.SessionState
-    # $Pester.SessionState = $sessionState
-    # Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
+            [hashtable]
+            $___Parameters___ = @{ },
 
-    # do {
-    # Write-ScriptBlockInvocationHint -Hint "InModuleScope" -ScriptBlock $ScriptBlock
-    & $module $ScriptBlock @Parameters @ArgumentList
-    # } until ($true)
-    # }
-    # finally {
-    # $Pester.SessionState = $originalState
-    # Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionStateInternal $originalScriptBlockScope
-    # }
+            [object[]]
+            $___ArgumentList___ = @(),
+
+            [System.Management.Automation.SessionState]
+            ${Session State},
+
+            ${Set Dynamic Parameter Variable}
+        )
+
+        # This script block exists to hold variables without polluting the test script's current scope.
+        # Dynamic parameters in functions, for some reason, only exist in $PSBoundParameters instead
+        # of being assigned a local variable the way static parameters do.  By calling Set-DynamicParameterVariable,
+        # we create these variables for the caller's use in a Parameter Filter or within the mock itself, and
+        # by doing it inside this temporary script block, those variables don't stick around longer than they
+        # should.
+
+        & ${Set Dynamic Parameter Variable} -SessionState ${Session State} -Parameters $___Parameters___
+
+        # define this in the current scope to be used instead of $PSBoundParameter if needed
+        # $PesterBoundParameters = if ($null -ne $___Parameters___) { $___Parameters___ } else { @{} }
+        & ${Script Block} @___Parameters___ @___ArgumentList___
+    }
+
+    Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
+    Set-ScriptBlockScope -ScriptBlock $wrapper -SessionState $sessionState
+    $splat = @{
+        'Script Block'                   = $ScriptBlock
+        '___ArgumentList___'             = $ArgumentList
+        '___Parameters___'               = $Parameters
+        'Session State'                  = $sessionState
+        'Set Dynamic Parameter Variable' = $SafeCommands['Set-DynamicParameterVariable']
+    }
+
+    Write-ScriptBlockInvocationHint -Hint "InModuleScope" -ScriptBlock $ScriptBlock
+    & $wrapper @splat
 }
 
 function Get-ScriptModule {
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -71,6 +71,7 @@
         [HashTable]
         $Parameters = @{},
 
+        [object[]]
         $ArgumentList = @()
     )
 
@@ -92,7 +93,7 @@
         $private:______parameters = $private:______inmodule_parameters.Parameters
 
         if ($private:______parameters.Count -gt 0) {
-            & $private:______inmodule_parameters.ScriptBlock @private:______parameters @private:______arguments
+            & $private:______inmodule_parameters.ScriptBlock @private:______arguments @private:______parameters
         }
         else {
             # Not splatting parameters to avoid polluting args
@@ -103,9 +104,9 @@
     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
         $hasParams = 0 -lt $Parameters.Count
         $hasArgs = 0 -lt $ArgumentList.Count
-        $arguments = $($(if ($hasArgs) { foreach ($a in $ArgumentList) { "'$($a)'" } }) -join ", ")
-        $params = $(if ($hasParams) { foreach ($p in $Parameters.GetEnumerator()) { "$($p.Key) = $($p.Value)" } }) -join ", "
-        Write-PesterDebugMessage -Scope Runtime -Message "Running scriptblock { $scriptBlock } in module $($ModuleName)$(if ($hasParams) { " with parameters: $params" })$(if ($hasArgs) { "$(if ($hasParams) { ' and' }) with arguments: $arguments" })."
+        $inmoduleArguments = $($(if ($hasArgs) { foreach ($a in $ArgumentList) { "'$($a)'" } }) -join ", ")
+        $inmoduleParameters = $(if ($hasParams) { foreach ($p in $Parameters.GetEnumerator()) { "$($p.Key) = $($p.Value)" } }) -join ", "
+        Write-PesterDebugMessage -Scope Runtime -Message "Running scriptblock { $scriptBlock } in module $($ModuleName)$(if ($hasParams) { " with parameters: $inmoduleParameters" })$(if ($hasArgs) { "$(if ($hasParams) { ' and' }) with arguments: $inmoduleArguments" })."
     }
 
     Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -95,12 +95,8 @@
             ${Set Dynamic Parameter Variable}
         )
 
-        # This script block exists to hold variables without polluting the test script's current scope.
-        # Dynamic parameters in functions, for some reason, only exist in $PSBoundParameters instead
-        # of being assigned a local variable the way static parameters do.  By calling Set-DynamicParameterVariable,
-        # we create these variables for the caller's use in a Parameter Filter or within the mock itself, and
-        # by doing it inside this temporary script block, those variables don't stick around longer than they
-        # should.
+        # This script block is used to create variables for provided parameters that
+        # the real scriptblock can inherit. Makes defining a param-block optional.
 
         & ${Set Dynamic Parameter Variable} -SessionState ${Session State} -Parameters $___Parameters___
 
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -19,6 +19,7 @@
    The code to be executed within the script module.
 .PARAMETER Parameters
    A optional hashtable of parameters to be passed to the scriptblock.
+   Parameters are automatically made available as variables in the scriptblock.
 .PARAMETER ArgumentList
    A optional list of arguments to be passed to the scriptblock.
 .EXAMPLE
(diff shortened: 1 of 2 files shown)


fflaten:
@nohwnd Am I on the right track? I've yet to understand when to set sessionstate vs scriptblock hints, so I know they're wrong. Haven't used the debug-messages before, so not sure what I'd expect or what would be useful in this scenario.

src/functions/InModuleScope.ps1
```diff
@@ -75,26 +75,48 @@
...
+    Set-ScriptBlockScope -ScriptBlock $wrapper -SessionState $sessionState
+    $splat = @{
+        'Script Block'                   = $ScriptBlock
+        '___ArgumentList___'             = $ArgumentList
+        '___Parameters___'               = $Parameters
+        'Session State'                  = $sessionState
+        'Set Dynamic Parameter Variable' = $SafeCommands['Set-DynamicParameterVariable']
+    }
+
+    Write-ScriptBlockInvocationHint -Hint "InModuleScope" -ScriptBlock $ScriptBlock
```
nohwnd, line 118: The hint looks okay. 

I added them for one case and at that time it seemed like a great idea, but I don't remember ever using it again. It is sometimes helpful to have an Id on the session state, but normally I just check the $executioncontext.sessionsstate and look at module to see in which session state I am in, and when debugging issues with having multiple session states with the same name it is usually together with Mock, so I just look for Get-Command Pester* which gives me mocks with id's and I can tell if I am in the same session state. 

I actually plan to get rid of the hints. Or maybe we should incorporate them into the diag log? I need to see how verbose it is and how much I can gather from them, e.g. when debugging an issue like the one where scriptblock was running in a different session state than the wrapper.

src/functions/InModuleScope.ps1
```diff
@@ -75,26 +75,48 @@
...
-    # }
-    # finally {
-    # $Pester.SessionState = $originalState
-    # Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionStateInternal $originalScriptBlockScope
-    # }
+    $wrapper = {
+        param (
+            [Parameter(Mandatory = $true)]
+            [scriptblock]
+            ${Script Block},
```
nohwnd, line 84: Consider using $private: so the variables don't leak to child scopes. In here you don't have any variables that you need to use in child scope, so it's a better way than having all of those variables have weird names. But I see that you are just copying the same approach from a different place in the code base.

src/functions/InModuleScope.ps1
```diff
@@ -75,26 +75,48 @@
...
+        )
+
+        # This script block is used to create variables for provided parameters that
+        # the real scriptblock can inherit. Makes defining a param-block optional.
+
+        & ${Set Dynamic Parameter Variable} -SessionState ${Session State} -Parameters $___Parameters___
+
+        # define this in the current scope to be used instead of $PSBoundParameter if needed
+        # $PesterBoundParameters = if ($null -ne $___Parameters___) { $___Parameters___ } else { @{} }
+        & ${Script Block} @___Parameters___ @___ArgumentList___
```
nohwnd, line 105: This is problematic when you don't define param and use $args. In that case parameters will become the first items in the $args array. You should do `@___ArgumentList___ @___Parameters___` instead that way it works slightly differently than normal powershell would, but your args will still work, and parameters will also bind if present.

src/functions/InModuleScope.ps1
```diff
@@ -75,26 +75,48 @@
...
+    }
+
+    Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
+    Set-ScriptBlockScope -ScriptBlock $wrapper -SessionState $sessionState
+    $splat = @{
+        'Script Block'                   = $ScriptBlock
+        '___ArgumentList___'             = $ArgumentList
+        '___Parameters___'               = $Parameters
+        'Session State'                  = $sessionState
+        'Set Dynamic Parameter Variable' = $SafeCommands['Set-DynamicParameterVariable']
```
nohwnd, line 115: I don't think you need to use this, it's trying to avoid setting common parameters, because it's working together with the proxy function generated from a real function. With InModuleScoe we won't be getting those unless user explicitly specifies them, and then they would be surprised if they did not have them set.

This is a pattern to copy: 

https://github.com/pester/Pester/blob/c28bdffec9818a5ff1c03f8c5fa5607e919d6900/src/functions/Mock.ps1#L1115-L1164

Context is a hashtable of variableName -> value, as you would have from -Parameters

$______isInMockParameterFilter you don't need to define. 

& $private:______mock_parameters.ScriptBlock @______arguments here also define the parameters as second splat.

Change the debug logging to go to Runtime instead of Mock.

src/functions/InModuleScope.ps1
nohwnd: See below for a link where you can mostly just copy paste the implementation.

src/functions/InModuleScope.ps1
fflaten: Good point about the common-parameters. Attempt to reuse some logic, but I agree it adds unnecessary overhead calling the function and extra check.

Thoughts on whether to include `$PesterBoundParameters`? I commented it out for now as the name wasn't logic in this case considering the parameters are bound through a single `-Parameters`. Could consider `$_` like `Add-DataToContext` to provide the whole dictionary. Or just stick with variables-only?

src/functions/InModuleScope.ps1
fflaten: Yeah, I stripped down behavior execution. 👍 Will absolutely simplify and use private-scope. Thanks

src/functions/InModuleScope.ps1
fflaten: Good point, adding test to make sure first item in args is exptected argument. I'm checking if `$___Parameters___ ` is empty and avoid splatting it all together. That way args won't be polluted with a null-item at the end.

src/functions/InModuleScope.ps1
nohwnd: let's keep it just the variables for now.

fflaten: https://github.com/pester/Pester/blob/c28bdffec9818a5ff1c03f8c5fa5607e919d6900/src/functions/Mock.ps1#L1132-L1135
What's going on here?

nohwnd: This is using a hashtable to pass a single parameter to the script block. This is to make it easier to pass the parameters to the scriptblock, so you don't have to add new ones to two places (param block and "argument list" when calling the scriptblock).

But you cannot splat from member on an object, so the values are saved into a variable. And then splatted.

It's a weird mix of historical approaches to this where ____ and single variable is used to avoid polluting the local scope, and to avoid naming conflicts, but $private should be avoiding all of that.

fflaten: > You cannot splat from member on an object, so the values are saved into a variable. And then splatted.

Just figured that one out the hard way. Was about to silently delete the question, hehe 😄 

Style question: Do you prefer `& $wrapper $ScriptBlock $Parameters $ArugmentList $sessionState`, the single hashtable-parameter or the one below? private-scoped parameters makes splatting a bit ugly.
```powershell
$splat = @{
    'private:ScriptBlock'    = $ScriptBlock
    'private:Parameters'     = $Parameters
    'private:ArgumentList'   = $ArgumentList
    'private:SessionState'   = $sessionState
}
& $wrapper @splat
```

nohwnd: single hashtable, then list of parameters then splatting the parameters.

nohwnd: Interesting that splatting considers the modifiers to be part of the name. Looking at that the parameter itself is actually called private:ScriptBlock, never noticed that. Luckily we only need those for internal purposes. (otherwise it is possible to remove the variable from the local scope to get the same effect with public parameters.

src/functions/InModuleScope.ps1
```diff
@@ -78,45 +78,47 @@
...
+        foreach ($private:______current in $private:______inmodule_parameters.Parameters.GetEnumerator()) {
+            $private:______inmodule_parameters.SessionState.PSVariable.Set($private:______current.Key, $private:______current.Value)
+        }
+
+        # Splatting expressions isn't allowed. Assigning to new private variable
+        $private:______arguments = $private:______inmodule_parameters.ArgumentList
+        $private:______parameters = $private:______inmodule_parameters.Parameters
+
+        if ($private:______parameters.Count -gt 0) {
+            & $private:______inmodule_parameters.ScriptBlock @private:______parameters @private:______arguments
```
nohwnd, line 95: Change the order here please.

src/functions/InModuleScope.ps1
```diff
@@ -78,45 +78,47 @@
...
+        $hasParams = 0 -lt $Parameters.Count
+        $hasArgs = 0 -lt $ArgumentList.Count
+        $arguments = $($(if ($hasArgs) { foreach ($a in $ArgumentList) { "'$($a)'" } }) -join ", ")
+        $params = $(if ($hasParams) { foreach ($p in $Parameters.GetEnumerator()) { "$($p.Key) = $($p.Value)" } }) -join ", "
+        Write-PesterDebugMessage -Scope Runtime -Message "Running scriptblock { $scriptBlock } in module $($ModuleName)$(if ($hasParams) { " with parameters: $params" })$(if ($hasArgs) { "$(if ($hasParams) { ' and' }) with arguments: $arguments" })."
     }
 
     Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
     Set-ScriptBlockScope -ScriptBlock $wrapper -SessionState $sessionState
     $splat = @{
```
nohwnd, line 113: Rename to inmoduleParameters?

src/functions/InModuleScope.ps1
fflaten: of the splats or flip the if/else?

nohwnd: The splatting. In WindowsPowerShell you will get the parameters as first in the args array if you splat them first and there is no param block: 

```powershell

$p = @{ a = 10 }
$a = 1, 2

& {
    $args[0]
} @p @a

& {
    $args[0]
} @a @p

# output 
-a:
1
```

fflaten: Updated, but it feels wrong considering it breaks with previous behavioro and how PowerShell does it.

nohwnd: PowerShell does it correctly in v7, but not in v5. If you specified params you would not see change in behavior with this approach. It is more that the previous version of InModuleScope does it incorrectly and you are now fixing it. :) If someone was depending on $args[0] they would not be able to get them consistently across versions and we would have an issue for that already I think.

fflaten: Aha. That must be 7.1 fix. I see it working as you say in 7.1.3, while 7.0.6 (used in devcontainer atm.) behaves the same as 5.1. 👍

nohwnd: kk did not know that.

nohwnd (approved):

nohwnd:
LGTM, merge if you are done, or re-request review later :)

fflaten:
Added another test and updated help with example. Good to go. I don't have permissions to merge with protected branch, so your turn 👍

plastikfan:
I've just tested this functionality on a test suite and am not getting the expected result. I am not sure though if I'm getting this right so it may be something I'm doing wrong. Should I raise a new issue? I've have a sample module that you can test this with.

Effectively, the old test code defines the InModuleScope statement as:

```powershell
    ) {
      InModuleScope ims -Parameters @{ Scenario = $Scenario; Result = $Result; ExpectedSuccess = $ExpectedSuccess; } {
        param(
          [string]$Scenario,
          [object]$Result,
          [boolean]$ExpectedSuccess
        )
```

Where the template parameters are __Scenario__, __Result__ and __ExpectedSuccess__.

I've changed this to become:

```powershell
    ) {
      InModuleScope ims {

```

Please note that the binding error I'm currently seeing is with $Result:

```
[-] convert-ActionResult.given:  .should: return result with Success = True 22ms (21ms|1ms)
 RuntimeException: The variable '$Result' cannot be retrieved because it has not been set.
 at <ScriptBlock>, C:\Users\Plastikfan\dev\PoSh\ims\convert-ActionResult.tests.ps1:75
```

... and $Result can be a string or a PSCustomObject.

The whole module and test suite are available at [ims](https://github.com/plastikfan/ims/tree/master)

If you really need to raise a new issue, then let me know. I suspect that I'm not doing this the way as intended, so I'd like this to be checked first before I raise an issue.

Cheers,

fflaten:
> I've changed this to become: `) { InModuleScope ims {`

For future readers; as mentioned in the discussion, this PR removes the requirement for the param-block only. You still need to specify `-Parameters <hash table>` to pass the values into the module state where the scriptblock is executed. Each key will be available as named variables inside the scriptblock.

PR #1959, opened 2021-05-25
Improve tracer based CC

nohwnd:
Resolve each command to see if we need to hoist it up to have a hit point associated to some of the parent parts of the code. 

There is one edge case where we don't report the command as hit, when param() block is used and default value is provided to a parameter which is a command. 

There is one edge case where we report coverage, but breakpoint based coverage does not report it, when a switch returns a scriptblock. In such case I think we are correct to report the scriptblock literal as covered.

See CoverageTestFile.ps1

There are also some extremely weird issues on PowerShell 3, where collections stop being iteratable by foreach. not sure what that is about.


src/csharp/Pester/Tracing/CodeCoverageTracer.cs
```diff
@@ -53,14 +85,21 @@ public void Trace(IScriptExtent extent, ScriptBlock _, int __)
                 return;
 
 
-            var point = lineColumn[key2];
-            if (point.Hit == true)
+            var points = lineColumn[key2];
+            if (points.TrueForAll(a => a.Hit))
```
nohwnd, line 89: This seems overkill, we are not adding hits once the collection is established, so we might just check the first 1 if it is true or false.

Issue #1448, opened 2020-02-06
'Should' does not log arrays with $null elements accurately


johndog:
## 1. General summary of the issue

Run this command:

` echo a b c $null | should be "a","b","c"   

Result:
```
Expected @('a', 'b', 'c'), but got @('a', 'b', 'c').
At C:\Program Files\WindowsPowerShell\Modules\Pester\4.9.0\Functions\Assertions\Should.ps1:169 char:9
+         throw ( New-ShouldErrorRecord -Message $testResult.FailureMes ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidResult: (System.Collections.Hashtable:Hashtable) [], Exception
    + FullyQualifiedErrorId : PesterAssertionFailed

```

Specifically, the first line logs that the arrays are the same when they aren't (or the test would have failed).

## 2. Describe Your Environment

Pester version     : 4.9.0 C:\Program Files\WindowsPowerShell\Modules\Pester\4.9.0\Pester.psm1
PowerShell version : 5.1.19555.1001
OS version         : Microsoft Windows NT 10.0.19555.0

## 3. Expected Behavior

If the test fails, the line:

`Expected @('a', 'b', 'c'), but got @('a', 'b', 'c').`

should indicate why, instead of printing two equal arrays and saying they aren't equal.

Note that this test passes:

`echo a b c $null | should be "a","b","c",$null  `

And this test actually outputs \<empty\>:

```
> echo a b c $null | should be "a","b","c",'' 

Expected @('a', 'b', 'c', <empty>), but got @('a', 'b', 'c').
```

## 4. Possible Solution
Suggested outcomes:

1) `Expected @('a', 'b', 'c'), but got @('a', 'b', 'c', $null).`
2) `Expected @('a', 'b', 'c'), but got @('a', 'b', 'c', '').`

## 6. Context
This makes it difficult to debug a test in which an array evaluates to an empty element.

nohwnd:
This looks like an issue with Format-Nicely, but looking at the code it should work correctly. Would you be willing to PR a fix for this? 

https://github.com/pester/Pester/blob/master/Dependencies/Format/Format.psm1#L11 <- this code should call Format-Null which would print `$null`.

I would start by adding tests here https://github.com/pester/Pester/blob/master/Dependencies/Format/Format.Tests.ps1#L21
[...]

PR #1961, opened 2021-05-30
Fixed formatting for null elements in an array

ArmaanMcleod:
## PR Summary
I had some time and added a fix for #1448.

`Format-Nicely` from `Format.ps1` now formats `$null` elements to `$null` in the array output.
e.g. `Format-Collection -Value @('a', 'b', 'c', $null)` -> `@('a', 'b', 'c', $null)`

The previous behaviour would omit the `$null` value.
e.g. `Format-Collection -Value @('a', 'b', 'c', $null)` -> `@('a', 'b', 'c')`

To achieve the result from the first example, I changed the implementation to not use `Select-Object`, since it removes null values, and instead use a simple loop to format each element and join back up to a comma separated string. 

I've also changed the ellipsis to indicate how many elements have been left out with `...<leftover count>` instead of just `...`. Let me know if this needs to be tweaked or its fine to keep `...`.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Format.ps1
+++ b/src/Format.ps1
@@ -13,7 +13,14 @@ function Format-Collection ($Value, [switch]$Pretty) {
     }
     $count = $Value.Count
     $trimmed = $count -gt $Limit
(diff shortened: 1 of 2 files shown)


nohwnd (approved):

nohwnd:
Thanks!

Issue #1328, opened 2019-07-02
Build on PowerShell 7


nohwnd:
It would be nice to start building Pester on PowerShell 7 as well to see if it is compatible and keep it that way. To achieve that we need to research on which build servers v7 is already available. Right now we are building on TravisCI (Linux and MacOS), on AppVeyor (PowerShell 4+) and AzureDevOps (PowerShell 2&3).

Not sure if the build task needs to run on all three platforms, but it would be nice.

The change should apply to Pester v4, and then be adopted for v5 as well.

kilasuit:
There's no reason you could not add a task in sideload v7 in either build service using the release zip deployment method and invoking pwsh from the extracted directory.

fflaten:
I'd say the primary focus of this should be Pester v5 now.
@nohwnd considering you now build with local runners in AzDO only then you probably have to fix assign yourself :)

nohwnd:
v5 CI already builds on every version since ps3. with PS 3, 4 and 6.2 being self-hosted and the rest is on hosted agents.

PR #1964, opened 2021-05-31
Add formatdata to ensure list view for all sections in PesterConfiguration

fflaten:
## PR Summary
Dynamically generates a ps1xml with formatdata for configuration sections in [PesterConfiguration] to ensure that all sections use list view. Currently Output and Should uses table due to less than 5 options.

```powershell
# Before
(New-PesterConfiguration).Should

ErrorAction
-----------
Controls if Should throws on error. Use 'Stop' to throw on error, or 'Continue' to fail at the end of the test. (Stop, default: Stop)

(New-PesterConfiguration).Output

Verbosity
---------
The verbosity of output, options are None, Normal, Detailed and Diagnostic. (Normal, default: Normal)

# After 
(New-PesterConfiguration).Should

ErrorAction : Controls if Should throws on error. Use 'Stop' to throw on error, or 'Continue' to fail at the end of the test. (Stop, default: Stop)

(New-PesterConfiguration).Output

Verbosity : The verbosity of output, options are None, Normal, Detailed and Diagnostic. (Normal, default: Normal)
```

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -53,6 +53,11 @@ param (
 
 $ErrorActionPreference = 'Stop'
 Get-Module Pester | Remove-Module
+
+if ($Clean -and $PSVersionTable.PSVersion -lt [version]'5.1') {
+    throw "Clean build of Pester requires PowerShell 5.1 or greater. If you have already compiled the assemblies and only modified powershell-files, try calling ./build.ps1 without -Clean."
+}
+
 if ($Clean -and (Test-Path "$PSScriptRoot/bin")) {
     Remove-Item "$PSScriptRoot/bin" -Recurse -Force
 }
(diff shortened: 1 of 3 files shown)


fflaten:
@nohwnd Not sure if this has been discussed in a issue or if I'm confusing with Profiler, but thoughts on this? The table view of Output and Should has been driving me mad 😄 

> v5 CI already builds on every version since ps3. with PS 3, 4 and 6.2 being self-hosted and the rest is on hosted agents.

Source: https://github.com/pester/Pester/issues/1328#issuecomment-843908904

Are you sure it **builds** on PS 7? Looks like it builds on something older while the published artifact is tested on all versions. Do we need to support building Pester at PSv3 or could we upgrade that to PS v5.1 (or later) and just keep testing it on older versions?

nohwnd:
Yeah looks like nice addition. 


> build

I think in the original issue I said "build", but meant "runs and passes tests". It was also for Pester v4, and imho there we were compiling all types inline, or importing them from external libaries, so we did not have a build phase. 

In Pester v5 I am 100% sure it (actually) builds on PS7 because I do that all the time, and also on PS5, and I've built it yesterday in PS3. So imho we are good in that regard. Why do you ask? 

I remember we had a change in the build.ps1, but I reverted it for the auto generation of help because it made a "circular" dependency, where Test-ModuleManifest needed the whole module to be loadable (we get some info from there), but we did not have that yet, because we did not generate the help yet. Or maybe it was because I am loading the assembly version from it?

In any case I reverted to just reading it as psd, so no we should not have any problem on any PS version.

I fail to see how this issue is connected to it. Why do you ask? 🙂

fflaten:
> Why do you ask? 

See failed checks. The code added to build script uses a Create-constructor not available in current build runner PS-version. Works in 5.1 and above.  

It's only used to generate the format.ps1xml during build, so the constructor is not required for end-users.

Haven't found a server old enough so I can see if there's an alternative way to create the object. 😅

fflaten:
Confirmed: `[System.Management.Automation.ListControl]::Create()` is new in 5.1

nohwnd:
Any chance you could encapsulate that into the  Pester Factory class? I use that when I don't have simple way to create a class without New-Object. 

Or is the type new, and we would not have it in the older version of powershell? 

Or maybe you could wrap that build part into powershell > 5, and so I can still build on PS3 (which happens once in a while), but our CI build will always running on PS7 or PS5 at worst?

fflaten:
New types in 5.0, so factory would just hide the mess to support PSv3 builds

Wrapping in >PS3 would leave FormatsToProcess reference in psd1 to non-existing file after clean build which breaks import. 

Then it's better to use static ps1xml for now OR write the XML directly with string templates (like EzOut does, but a build dependency felt overkill for this simple case). The current list-views are very simple to write as xml-strings, I just preferred to avoid it when builder-types are available. 🙂

nohwnd:
Nah just make it build on ps5+ only, it's okay

fflaten:
> Nah just make it build on ps5+ only, it's okay

Not sure how you'd like that. CI build stage needs to be changed to a PSv5.1+ runner (maybe just latest PS7?) so the checks in this and future PRs will pass. Would prefer if you do add that commit in case you want to build locally as now :-)

nohwnd:
Yeah, we just need to build on a different machine. I moved it to the hosted VS2019, seemed like the easiest option, even though the initial build might be a bit slower now because we wait for a machine to be available. We don't really care which dotnet SDK is used to build us, and it has .Net 452 targetting pack installed.  https://github.com/pester/Pester/pull/1966

nohwnd:
Merged the other PR.

fflaten:
P-tests were calling incremental build on PS3 and PS4. Made PesterConfiguration docs and format a Clean-build step as they both depend on changes to the assembly. That allow incremental script-updates possible in the test-runs. Let's see how it goes now. 

If it works, we might even clean up:
https://github.com/pester/Pester/blob/67ab4172020712eda173abf906ab8604151cd0b7/build.ps1#L61-L64

and
https://github.com/pester/Pester/blob/67ab4172020712eda173abf906ab8604151cd0b7/build.ps1#L103-L108

nohwnd (approved):

nohwnd:
Yeah nice. Looks like the best "compromise", this way when I don't need to change the C# code on PS3 it will still work nice.

fflaten:
@nohwnd Can you take it for a spin on PSv3 or PSv4 just to verify the format works as expected. Still don't have a VM available (TODO.. :-) )

PR #1969, opened 2021-06-04
Update output for Discovery and Filter

### LINKED ISSUE #1552: [v5] Consider moving "Discovering" messages to Output Diagnostic Level

### LINKED ISSUE #1746: Debugging generated tests takes for ever due to the output of filter commands

If I'm generating tests (100s due to permutations), when you try and debug them, the discovery phase outputs lots of FILTER .... diagnostic information, and thats really slow.  takes minutes to output them all event though discovery in non debug mode is 1second

Provide some way to prevent the display of the FILTER messages.

### LINKED ISSUE #1972: Changes made at runtime leaks to user's advanced configuration object
Using advanced configuration `Invoke-Pester -Configuration $conf`, changes made by `Invoke-Pester` at runtime is leaked as the configuration-object is used directly.

## Describe your environment
Pester 5.2.0 and latest 5.3.0 alpha.

## Steps to reproduce
```powershell
$conf = New-PesterConfiguration; $conf.Run.ScriptBlock = { }; $conf.Output.Verbosity = 'Diagnostic'
$conf.Debug.WriteDebugMessagesFrom = 'foo'
Invoke-Pester -Configuration $conf
```

## Expected Behavior
My configuration object should include original values after running `Invoke-Pester -Configuration $conf`
```powershell
$conf.Debug.WriteDebugMessages.Value
False

$conf.Debug.WriteDebugMessagesFrom.Value
foo
```

## Current Behavior
Value is overwritten.
```powershell
$conf.Debug.WriteDebugMessages.Value
True

$conf.Debug.WriteDebugMessagesFrom.Value
Discovery
Skip
Filter
Mock
CodeCoverage
```

fflaten:
## PR Summary
- Removes file-specific discovery message from Detailed view. This is still available as Discovery debug-messages in Diagnostic
- Output number of tests to be executed.
- Removes redundant and noisy active tag filter debugmessages (covered in initial discovery-output in Detailed + Diagnostic), see before-image here https://github.com/pester/Pester/issues/1746#issuecomment-854043879
- Fix runtime configuration leak when using advanced configuration with `Invoke-Pester -Configuration ...`

Example of output after PR:
image
image

Diagnostic:
image

image

Fix #1552
Fix #1746 
Fix #1972

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1,4 +1,4 @@
-function Assert-ValidAssertionName {
+﻿function Assert-ValidAssertionName {
     param([string]$Name)
     if ($Name -notmatch '^\S+$') {
         throw "Assertion name '$name' is invalid, assertion name must be a single word."
@@ -911,8 +911,14 @@ function Invoke-Pester {
             }
 
             if ('Diagnostic' -eq $PesterPreference.Output.Verbosity.Value) {
+                # Enforce the default debug-output as a minimum. This is the key difference between Detailed and Diagnostic
                 $PesterPreference.Debug.WriteDebugMessages = $true
-                $PesterPreference.Debug.WriteDebugMessagesFrom = "Discovery", "Skip", "Mock", "CodeCoverage"
+                $missingCategories = foreach ($category in @("Discovery", "Skip", "Mock", "CodeCoverage")) {
+                    if ($PesterPreference.Debug.WriteDebugMessagesFrom.Value -notcontains $category) {
+                        $category
+                    }
+                }
+                $PesterPreference.Debug.WriteDebugMessagesFrom = $PesterPreference.Debug.WriteDebugMessagesFrom.Value + @($missingCategories)
             }
 
             $plugins +=
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1,4 +1,4 @@
-﻿function Assert-ValidAssertionName {
+function Assert-ValidAssertionName {
     param([string]$Name)
     if ($Name -notmatch '^\S+$') {
         throw "Assertion name '$name' is invalid, assertion name must be a single word."
@@ -887,7 +887,12 @@ function Invoke-Pester {
             # preference is inherited in all subsequent calls in this session state
             # but we still pass it explicitly where practical
             if (-not $hasCallerPreference) {
-                [PesterConfiguration] $PesterPreference = $Configuration
+                if ($PSBoundParameters.ContainsKey('Configuration')) {
+                    # Advanced configuration used, merging to get new reference
+                    [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge([PesterConfiguration]::Default, $Configuration)
+                } else {
+                    [PesterConfiguration] $PesterPreference = $Configuration
+                }
             }
             elseif ($hasCallerPreference) {
                 [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge($callerPreference, $Configuration)
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1,4 +1,4 @@
-﻿function Assert-ValidAssertionName {
+function Assert-ValidAssertionName {
     param([string]$Name)
     if ($Name -notmatch '^\S+$') {
         throw "Assertion name '$name' is invalid, assertion name must be a single word."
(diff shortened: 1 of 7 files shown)


src/Main.ps1
```diff
@@ -907,7 +907,7 @@ function Invoke-Pester {
 
             if ('Diagnostic' -eq $PesterPreference.Output.Verbosity.Value) {
                 $PesterPreference.Debug.WriteDebugMessages = $true
-                $PesterPreference.Debug.WriteDebugMessagesFrom = "Discovery", "Skip", "Filter", "Mock", "CodeCoverage"
+                $PesterPreference.Debug.WriteDebugMessagesFrom = "Discovery", "Skip", "Mock", "CodeCoverage"
             }
```
fflaten, line 911: Is it right that Diagnostic should always override the debug-categories? These are already set as default in PesterConfiguration-object. If I use advanced configuration with Diagnostic output and have manually added the 'Filter' category, this will remove that.

Maybe we should rely completely on the default in PesterConfiguration? Or at least just ensure here that all these are present **as a minimum**?

src/functions/Output.ps1
```diff
@@ -553,7 +542,13 @@ function Get-WriteScreenPlugin ($Verbosity) {
...
 
     $p.RunStart = {
-        & $SafeCommands["Write-Host"] -ForegroundColor Magenta "Running tests."
+        param ($Context)
+
+        $testsToRun = 0
+        foreach ($test in @(View-Flat -Block $Context.Blocks)) {
+            if ($test.ShouldRun) { $testsToRun++ }
+        }
+        & $SafeCommands["Write-Host"] -ForegroundColor Magenta "Running $testsToRun tests."
```
fflaten, line 551: Rephrase? "Running x tests in total" maybe? Might feel weird currently considering the next line is "Running tests from <file number 1>"

src/Main.ps1
nohwnd: I think you can check on the object if it was changed and only set this if not changed. I probably did not take that in consideration, because adding the Diagnostic output was an afterthought :)

src/functions/Output.ps1
nohwnd: Maybe the filter output should move after discovery? Then:
Discovery found 16 tests in 34 ms. (instead of Discovery finished)
Filtering tests with filter: ....
Filters selected 4 tests to run.

src/Main.ps1
fflaten: If `IsOriginalValue() -eq $true` then it would already have those categories defined (set in `DebugConfiguration.cs`). No need to re-apply here. What about WriteDebugMessages itself? We should probably keep force-enabled here like it is, right? Can't remember if Diagnostic provides any value to the user without any debugmessages enabled.

nohwnd: It does not, it is Detailed view + Debug log. I checked the code, we should probably just append the values, they are array of strings, and we don't have to care if there are duplicates. That way you will get what user specified in config or pesterpreference (or both, we merge the object above this call) plus the values from Diagnostic.

src/functions/Output.ps1
fflaten: I'm guessing that includes the list of active filters. It would be more natural to delay all filter-related messages in the output until they're actually processed. This is a "known issue" from https://github.com/pester/Pester/pull/1727#issue-509026272 as we don't have a plugin-step between discovery and filter/post-process. Would have to do host-output in the middle of the code (or implement something like DiscoveryPostProcessStart-step).

Currently we can update the summary in `DiscoveryEnd`-step like below and then keep the "Running tests." message after to indicate Run-phase has begun.
```
Discovery found 16 tests in 34 ms. 
Filters selected 4 tests to run.
```

nohwnd: I see. How about just postponing writing the filter out, and doing it after we are done filtering in DiscoveryEnd. 

<actual filtering is happening>
Filtered tests with filters: ....
Filters selected 4 tests to run.
Discovery finished in 34 ms.

src/Main.ps1
fflaten: Ended up adding the missing values only to keep the configuration clean. First to avoid the discussed leak in advanced config (now fixed), but then realized it's visible in passthru object.

nohwnd:
Perfect, thanks!

Issue #1912, opened 2021-04-20
The new mock logging for conflicting vars reports that variable is redefined when two mocks are used


nohwnd:

nohwnd:
It reports conflict when we run in function from the current scope. Because everything in in the same scope.

fflaten:
Do you remember how to repro this by any chance?

nohwnd:
Yeah, I think it was just having two mocks for the same command with a filter. Then the second mock filter would report all vars as duplicated.

PR #1973, opened 2021-06-05
Log conflicting variables in Should -Invoke -ParameterFilter

### LINKED ISSUE #1873: Test fails while it should pass
Hard to explain, but it took us a complete breakdown of a very complicated script to see what was happening. Please see the test code below.

PowerShell version : 5.1.14393.3866

## Steps to reproduce
The code below should pass for all tests but it reports the following errors:
>   [-] should pass because argument SamAccountName is only called once 425ms (421ms|4ms)
   Expected Get-Stuff to be called 1 times exactly but was called 2 times
>   [-] should pass because argument DistinguishedName is only called once 412ms (410ms|2ms)
   Expected Get-Stuff to be called 1 times exactly but was called 2 times

It seems like the parameter used in the function called `$SamAccountName` cannot be used as variable names in other places in the script outside of the function or it will break the check on how many times the function/mock is called.

```powershell
Describe 'test' {
    BeforeAll {
        Function Get-Stuff {
            Param (
                [String]$SamAccountName,
                [String]$DistinguishedName
            )
        }
        Mock Get-Stuff
    }
    it 'should pass because argument SamAccountName is only called once' {
        Get-Stuff -SamAccountName  'a'
        Get-Stuff -DistinguishedName  'b'
        
        Should -Invoke Get-Stuff -Times 1 -Exactly -ParameterFilter { 
            $SamAccountName
        }
    }
    it 'should pass because argument DistinguishedName is only called once' {
        Get-Stuff -SamAccountName  'a'
        Get-Stuff -DistinguishedName  'b'
        
        Should -Invoke Get-Stuff -Times 1 -Exactly -ParameterFilter { 
            $DistinguishedName
        }
    }
    it 'should pass because argument SamAccountName is only called once' {
        Get-Stuff -SamAccountName  'a'
        Get-Stuff -DistinguishedName  'b'
        
        foreach (
            $samAccountName in (0..5) # rename the variable and it works
        ) {
            # do something
        }

        Should -Invoke Get-Stuff -Times 1 -Exactly -ParameterFilter { 
            $SamAccountName
        }
    }
    it 'should pass because argument DistinguishedName is only called once' {
        Get-Stuff -SamAccountName  'a'
        Get-Stuff -DistinguishedName  'b'
        
        foreach (
            $DistinguishedName in (0..2) # rename the variable and it works
        ) {
            # do something
        }

        Should -Invoke Get-Stuff -Times 1 -Exactly -ParameterFilter { 
            $DistinguishedName
        }
    }
}
```
## Expected Behavior
Expected all tests to pass.

## Current Behavior
First test fails.

fflaten:
## PR Summary
New attempt at debug-logging variables that may cause false matches during `Should -Invoke` due to conflict with a known parameter or alias name in the mocked command.

Fix #1873
Related #1912

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -2798,13 +2798,15 @@ Describe "When inherited variables conflicts with parameters" {
         { Should -InvokeVerifiable } | Should -Throw
     }
 
-    It "KNOWN ISSUE: Should Invoke ParameterFilter will count false positive" {
+    It "Should Invoke ParameterFilter will count false positive for the first FunctionUnderTest call" {
         # https://github.com/pester/Pester/issues/1873
+        # this will pass the parameter filter because we define a variable param1 with the same name and value as the expected parameter value
         FunctionUnderTest | Should -Be 'default'
         FunctionUnderTest -param1 'abc' | Should -Be 'filtered'
         $param1 = 'abc'
 
         # This should show warning about conflict when in Diagnostic output (Mock debug message)
+        # about already having a variable in the scope that is the same as parameter name
         Should -Invoke FunctionUnderTest -ParameterFilter { $param1 -eq 'abc' } -Times 2 -Exactly
     }
 
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -2779,3 +2779,51 @@ Describe "Debugging mocks" {
         }
     }
 }
+
+Describe "When inherited variables conflicts with parameters" {
+    BeforeAll {
+        Mock FunctionUnderTest { 'default' }
+        Mock FunctionUnderTest { 'filtered' } -ParameterFilter { $param1 -eq 'abc' } -Verifiable
+    }
+
+    It "parameterized mock should not be called due to inherited variable" {
+        $param1 = 'abc'
+        FunctionUnderTest | Should -Be 'default'
+    }
+
+    It "InvokeVerifiable should not pass due to test variable" {
+        # Uses same logic as mock execution, so should not be tricked
+        $param1 = 'abc'
+        FunctionUnderTest | Should -Be 'default'
+        { Should -InvokeVerifiable } | Should -Throw
+    }
+
+    It "KNOWN ISSUE: Should Invoke ParameterFilter will count false positive" {
+        # https://github.com/pester/Pester/issues/1873
+        FunctionUnderTest | Should -Be 'default'
+        FunctionUnderTest -param1 'abc' | Should -Be 'filtered'
+        $param1 = 'abc'
+
+        # This should show warning about conflict when in Diagnostic output (Mock debug message)
+        Should -Invoke FunctionUnderTest -ParameterFilter { $param1 -eq 'abc' } -Times 2 -Exactly
+    }
+
+    It "Invoke ParameterFilter works as expected when PesterBoundParamters is used" {
+        # Workaround mentioned in debug message warning mentioned in previous test
+        FunctionUnderTest | Should -Be 'default'
+        FunctionUnderTest -param1 'abc' | Should -Be 'filtered'
+        $param1 = 'abc'
+
+        # No warning will be shown in debug as there's no conflict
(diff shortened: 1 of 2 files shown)


fflaten:
I can't repro the issue with mock execution, so the parameterfilter used to execute behaviors (and `Should -InvokeVerifiable`) doesn't seem to be affected. With this approach I don't get the false debug-warning when defining a second mock using ParameterFilter (#1912).

Ex. output:
image

Checking all parameters because `$context` only cover bound parameters + their aliases. Should invoke parameterfilter would still match calls where the variable exists in test-session and the parameter wasn't called (because only then we overwrite it).

Leaving as draft for now as tests are missing and I'm not sure I've covered all.

nohwnd:
Will have a look later, maybe I the problem I was seeing was different.

fflaten:
👍 Haven't taken the time to look into why the previous attempt failed, but I think I've reproduced it in the first test [here](https://gist.github.com/fflaten/c03e4d28e99e2a1c7bb44ffafe38c357). Also some other tests used to keep me sane while testing.

fflaten:
> 👍 Haven't taken the time to look into why #1912 happened, but I've either isolated the issue even more or found another false positive in that attempt. See first test here [here](https://gist.github.com/fflaten/c03e4d28e99e2a1c7bb44ffafe38c357).

The unexpected warnings in 5.2.0-alpha3 (first test in my link above) is caused by the mock hook parameter binding. It has already set variable `$SamAccountName` for the provided parameter in `Get-Stuff -SamAccountName 123` in the scope. When the mock is invoked later in hook, we execute `Test-ParameterFilter` in the same state as hook as part of `FindMatchingBehavior` at which point your code ran and noticed the existing inherited variable from the hook - always with the same value as it's actually the same parameter being set twice.

Since my code above is only invoked in `Should -Invoke`, the session isn't "polluted" by the hook parameters.

nohwnd:
Okay that is why I was seeing it 👍

fflaten:
Updated the debug message and added some tests to cover the scenarios.
Currently no test for the debug message itself as mocking `Write-PesterDebugMessage` is not a great idea. Running test in new process (time consuming) feels overkill for diagnostic, but can add if wanted.

tst/functions/Mock.Tests.ps1
```diff
@@ -2779,3 +2779,51 @@ Describe "Debugging mocks" {
...
+
+    It "KNOWN ISSUE: Should Invoke ParameterFilter will count false positive" {
+        # https://github.com/pester/Pester/issues/1873
+        FunctionUnderTest | Should -Be 'default'
+        FunctionUnderTest -param1 'abc' | Should -Be 'filtered'
+        $param1 = 'abc'
+
+        # This should show warning about conflict when in Diagnostic output (Mock debug message)
+        Should -Invoke FunctionUnderTest -ParameterFilter { $param1 -eq 'abc' } -Times 2 -Exactly
+    }
```
nohwnd, line 2809:
```suggestion
    It "Should Invoke ParameterFilter will count false positive for the first FunctionUnderTest call" {
        # https://github.com/pester/Pester/issues/1873
        # this will pass the parameter filter because we define a variable param1 with the same name and value as the expected parameter value
        FunctionUnderTest | Should -Be 'default'
        FunctionUnderTest -param1 'abc' | Should -Be 'filtered'
        $param1 = 'abc'

        # This should show warning about conflict when in Diagnostic output (Mock debug message)
        # about already having a variable in the scope that is the same as parameter name
        Should -Invoke FunctionUnderTest -ParameterFilter { $param1 -eq 'abc' } -Times 2 -Exactly
    }
```

nohwnd: That does not highlight the changes. I removed the Known issue as it sounds like we are doing something wrong that we don't expect. And added more details to comments. I did not change other code.

nohwnd:
> Currently no test for the debug message itself as mocking `Write-PesterDebugMessage` is not a great idea. Running test in new process (time consuming) feels overkill for diagnostic, but can add if wanted.


That's okay. It's great as is :)

nohwnd:
Yet another great contribution. :)

DarkLite1:
Indeed, great work @fflaten . Thank you very much for the work you put into this 👍

PR #1986, opened 2021-06-11
Add missing container-parameters to Data

### LINKED ISSUE #1980: Default parameter values are not applied in the test script
I'm working on a new test script and I need it to have a parameter defining the environment to be tested (Dev/Prod). I configured it to use Dev as a default, however it seems that those defaults are just ignored.

## Describe your environment
```
Pester version     : 5.2.0 C:\Users\username\OneDrive - PPD\Documents\PowerShell\Modules\Pester\5.2.0\Pester.psm1
PowerShell version : 7.1.3
```

## Steps to reproduce
Test script:

```powershell
[CmdletBinding()]
param (
    [Parameter()]
    [string]
    [ValidateSet('DEV', 'PROD')]
    $Environment = 'DEV'
)
 
Describe "My tests" {
    It "Environment parameter should be accessible" {
        $Environment | Should -BeExactly 'DEV'
    }
}
```

Running it either with `Invoke-Pester -Path` or with VSCode's `Debug tests` intellisense link resolves with a failed test:
```
[-] My tests.Environment parameter should be accessible 4ms (3ms|1ms)
 Expected exactly 'DEV', but got $null.
 at $Environment | Should -BeExactly 'DEV', C:\Users\username\Desktop\Tests\MyPester.Tests.ps1:11
 at <ScriptBlock>, C:\Users\username\Desktop\Tests\MyPester.Tests.ps1:11
```
-->

It only works if I initialize a PesterContainer with the Data provided

## Expected Behavior
I wish it would pick up the default value for the environment, cause that would make it so much easier to debug it through VSCode

## Current Behavior
Defaults for script parameters are completely ignored

I can workaround the script to use environmental variables to drive the tests, but this approach seems more intuitive.

fflaten:
## PR Summary
Pester containers may use parameters to support data-driven test generation.

This scenario currently has the following limitations:
- Default values for parameters are available as variables in Discovery, but not in Run (test execution) unless passed using tricks like `Describe 'd1' -ForEach { myparam = $myparam}`. This makes debugging harder since `Debug Tests` in VSCode doesn't allow the user to provide `-Data ...` for the container (file).
- Default values are also not available in `*All/*Each` setup-blocks placed outside any Describe/Context-blocks. The `Describe -ForEach ..`-workaround above does not work for these root-scoped setup-blocks.
- Discovery and Run are inconsistent when using aliases for script parameters. Given a container with parameter "MyParam" which has a alias of "MyAlias". When invoked with `New-PesterContainer ... -Data @{ MyAlias = 123 }`:
    - In Discovery: `$MyParam` would be 123 and `$MyAlias` would not exist. This is default PowerShell-behavior.
    - In Run: `$MyParam` would not exist and `$MyAlias` would be 123. This is because Pester adds uses the provided `-Data` directly during Run.
  
This PR adds logic to detect container-parameters not defined by the user at runtime and automatically adds them to Data for the Root-block of the  container. This fixes:

- Script-parameters are always available as variables in Discovery and Run. Has default value (null or value set in Parma-block) if not provided by the user.
- Default values are available in root-scoped setup blocks.
- `$MyParam` (see example above) is also defined in Run like it is in Discovery to be more consistent and similar to normal PowerShell-function/script behavior. 

Fix #1980

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -2649,7 +2649,7 @@ function Add-ContainerParameterDefaultValues ($RootBlock, $Container, $CallingFu
             }
         }
         "File" {
-            $originCommand = $PSCmdlet.SessionState.InvokeCommand.GetCommand($Container.Item.PSPath, [System.Management.Automation.CommandTypes]::ExternalScript)
+            $originCommand = $CallingFunction.SessionState.InvokeCommand.GetCommand($Container.Item.PSPath, [System.Management.Automation.CommandTypes]::ExternalScript)
             $command.Ast = $originCommand.ScriptBlock.Ast
 
             if ($callerCmd = $CallingFunction.SessionState.PSVariable.GetValue("PSCmdLet")) {
@@ -2676,10 +2676,10 @@ function Add-ContainerParameterDefaultValues ($RootBlock, $Container, $CallingFu
         }
 
         foreach ($param in $parametersToCheck) {
-            $v = $PSCmdlet.SessionState.PSVariable.Get($param)
+            $v = $CallingFunction.SessionState.PSVariable.Get($param)
             if ((-not $RootBlock.Data.ContainsKey($param)) -and $v) {
                 if ($PesterPreference.Debug.WriteDebugMessages.Value) {
-                    Write-PesterDebugMessage -Scope Discovery "Container parameter '$param' is undefined, but has default value '$($v.Value)'. Adding it to Data in Root-block for container."
+                    Write-PesterDebugMessage -Scope Discovery "Container parameter '$param' is undefined, adding to container Data with default value $(Format-Nicely $v.Value)."
                 }
                 $RootBlock.Data.Add($param, $v.Value)
             }
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -2395,7 +2395,7 @@ function New-BlockContainerObject {
         [String] $Path,
         [Parameter(Mandatory, ParameterSetName = "File")]
         [System.IO.FileInfo] $File,
-        $Data
+        $Data = @{}
     )
 
     $type, $item = switch ($PSCmdlet.ParameterSetName) {
@@ -2623,4 +2623,31 @@ function Invoke-InNewScriptScope ([ScriptBlock] $ScriptBlock, $SessionState) {
     $script:ScriptBlockSessionStateInternalProperty.SetValue($wrapper, $SessionStateInternal, $null)
 
     . $wrapper $Path $Data
+}
+
+function Add-ContainerParameterDefaultValues ($RootBlock, $Container, $SessionState) {
+    $scriptParameters = @(switch ($Container.Type) {
+            "ScriptBlock" {
+                if ($null -ne $container.Item.Ast.ParamBlock -and $container.Item.Ast.ParamBlock.Parameters.Count -gt 0) {
+                    foreach ($v in $container.Item.Ast.ParamBlock.Parameters.Name.VariablePath.UserPath) { $v }
+                }
+            }
+            "File" {
+                if ($originCommand = & $SafeCommands['Get-Command'] $Container.Item.PSPath) {
+                    foreach ($k in $originCommand.Parameters.Keys) { $k }
+                }
+            }
+            default { throw [System.ArgumentOutOfRangeException]"" }
+        })
+
+    foreach ($param in $scriptParameters) {
+        $v = $SessionState.PSVariable.Get($param)
+        # ignoring $null-valued variables to avoid polluting Data because unsure how to detect no default value $null vs explicit $null default value
+        if ((-not $RootBlock.Data.ContainsKey($param)) -and $v -and $null -ne $v.Value) {
+            if ($PesterPreference.Debug.WriteDebugMessages.Value) {
+                Write-PesterDebugMessage -Scope Discovery "Container parameter '$param' is undefined, but has default value '$($v.Value)'. Adding it to Data in Root-block for container."
+            }
+            $RootBlock.Data.Add($param, $v.Value)
+        }
+    }
 }
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -2627,53 +2627,24 @@ function Invoke-InNewScriptScope ([ScriptBlock] $ScriptBlock, $SessionState) {
     . $wrapper $Path $Data
 }
 
-function Add-ContainerParameterDefaultValues ($RootBlock, $Container, $CallingFunction) {
-    $command = [PSCustomObject]@{
-        Ast = $null
-        Parameters = $null
-    }
+function Add-MissingContainerParameters ($RootBlock, $Container, $CallingFunction) {
+    # Adds default values for container parameters not provided by the user.
+    # Also adds real parameter name as variable in Run-phase when alias was used, just like normal PowerShell will.
 
-    # $CallingFunction is $PSCmdlet passed from our caller. $CallingFunction.MyInvocation.MyCommand will return Describe/Context as it's the latest advanced functions to be called at that point.
-    # $CallingFunction.SessionState however is the state used to call Context/Describe, usually the script-state, so by geting "PSCmdLet"-variable from it, we can get the last
-    # advanced function in script-state which may be the container if it was advanced (we need to verify)
-
-    switch ($Container.Type) {
-        "ScriptBlock" {
-            $command.Ast = $container.Item.Ast
-
-            if ($callerCmd = $CallingFunction.SessionState.PSVariable.GetValue("PSCmdLet")) {
-                if ($callerCmd.MyInvocation.MyCommand.CommandType -eq 'Function' -and $callerCmd.MyInvocation.MyCommand.ScriptBlock.Id -eq $container.Item.Id) {
-                    # Correct scriptblock. Container is advanced, get related parameter names
-                    $command.Parameters = $callerCmd.MyInvocation.MyCommand.ParameterSets | & $SafeCommands['Where-Object'] { $_.Name -eq $callerCmd.ParameterSetName } | & $SafeCommands['Foreach-Object'] { $_.Parameters.Name }
-                }
-            }
-        }
+    # Using AST to get parameter-names as $PSCmdLet.MyInvocation.MyCommand only works for advanced functions/scripts/cmdlets.
+    # No need to filter on parameter sets OR whether default values are set because Powershell adds all parameters (not aliases) as variables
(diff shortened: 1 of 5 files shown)


src/Pester.Runtime.ps1
```diff
@@ -2624,3 +2624,30 @@ function Invoke-InNewScriptScope ([ScriptBlock] $ScriptBlock, $SessionState) {
 
     . $wrapper $Path $Data
 }
+
+function Add-ContainerParameterDefaultValues ($RootBlock, $Container, $SessionState) {
+    $scriptParameters = @(switch ($Container.Type) {
```
nohwnd, line 2629:
```
$PSCmdlet.MyInvocation.MyCommand.ParameterSets | ? Name -eq $PSCmdlet.ParameterSetName 
```

Did you consider taking the name from the calling command info? It does include some extra parameters, but will make sure we are getting only params from the currently used parameter set.

You can get the names this way and then query the variables as you do now.

Does not help with the $null problem. But if you are doing ast you should be able to see from it if there is assignment and default value. We have similar code here: https://github.com/pester/Pester/blob/main/src/functions/assertions/HaveParameter.ps1

fflaten: Good point about parameter sets. Hadn't thought about that yet.

Won't `$PSCmdlet.MyInvocation.MyCommand` reference Describe/Context itself? I couldn't think of a way to get CommandInfo except Get-Command which will only work for file-containers AFAIK. AST was a last-minute addition to support scriptblock-containers in this PoC-draft. You're right that it can also help with detecting if a parameter has default-value or not for both container-types.

fflaten: Looks like I can get it with `$PSCmdLet.SessionState.PSVariable.GetValue("PSCmdLet")` in Describe/Context 🎉

nohwnd: It should work as posted. $PSCmdlet is the caller of the current function. So for Describe that would be the script, ehm, unless that is inside of InModuleScope maybe?

The code I posted I actually called that from a module function in the same fashion you'd do that in Describe. Imho your code just resolves PSCmdlet from the same state in a convoluted way. But correct me if I am wrong please :)

fflaten: `$PSCmdLet` refers to the command/advanced function that's being executed (or a parent if the current is not advanced/cmdlet). Context/Describe are both advanced functions so calling `$PSCmdLet` inside those methods will return themselves since they're the latest.

My code returns the latest advanced function/cmdlet in the script-state. Since `Describe` runs in Pester's state, this will return the test-container or a parent like invokePesterStub.ps1 etc. if the container itself was not advanced. Demo:

```powershell
Remove-Module demoModule
New-Module demoModule {
    function descsimple { param()
        Write-Host "PSCmdlet in descsimple (Module state): $($PSCmdlet.MyInvocation.MyCommand.Name)" -ForegroundColor Green
        # Can't check sessionstate variable here because PSCmdlet is null. This is because both descsimple or simple are simple functions
     }
    function descadv { [CmdletBinding()]param()
        Write-Host "PSCmdlet in descadv (Module state): $($PSCmdlet.MyInvocation.MyCommand.Name)" -ForegroundColor Green
        Write-Host "PSCmdlet from sessionstate var in descadv (Caller state): $($PSCmdlet.SessionState.PSVariable.GetValue('PSCmdlet') | % { $_.MyInvocation.MyCommand.Name})" -ForegroundColor Green
    }
} | Import-Module -Force

function simple ($p) { Write-Host "PSCmdlet in simple: $($PSCmdlet.MyInvocation.MyCommand.Name)" -ForegroundColor Red; descsimple; descadv }
function advanced { [CmdletBinding()]param() Write-Host "PSCmdlet in advanced: $($PSCmdlet.MyInvocation.MyCommand.Name)" -ForegroundColor Yellow; descsimple; descadv }

PS> simple
PSCmdlet in simple:
PSCmdlet in descsimple (Module state):
PSCmdlet in descadv (Module state): descadv
PSCmdlet from sessionstate var in descadv (Caller state):

PS> advanced
PSCmdlet in advanced: advanced
PSCmdlet in descsimple (Module state):
PSCmdlet in descadv (Module state): descadv
PSCmdlet from sessionstate var in descadv (Caller state): advanced
```

nohwnd: Sounds correct… What the hell was I trying in my code then that it worked😁 I need to find it if I saved it.

fflaten: Magic code should always be saved 👍 That's why I commented the psvariable code here. I'm 99% sure I tried this and failed ten times before it suddenly started working consistently. Probably a solar storm last night affecting all worldwide CPUs 😅

fflaten: Omg. Default values are applied regardless of parametersets, so I might have wasted some time here. At least I've learned something. 😅 

```powershell
function t {
    [CmdLetBinding(DefaultParameterSetName = "Def")]
    param (
        [Parameter(ParameterSetName = "Def")]
        $ImportantParam = 'DEV',
        [Parameter(ParameterSetName = "NonDef")]
        $SomethingElse = 'NONE',
        $AnotherOne
    )

    Write-Host "ParameterSetName '$($PSCmdlet.ParameterSetName)'"
    Write-Host "ImportantParam (Def set): $ImportantParam"
    Write-Host "SomethingElse (NonDef set): $SomethingElse"
    Write-Host "AnotherOne (All sets): $AnotherOne"
}

PS > t
ParameterSetName 'Def'
ImportantParam (Def set): DEV
SomethingElse (NonDef set): NONE
AnotherOne (All sets):
PS > t -ImportantParam 123
ParameterSetName 'Def'
ImportantParam (Def set): 123
SomethingElse (NonDef set): NONE
AnotherOne (All sets):

PS > t -SomethingElse 456
ParameterSetName 'NonDef'
ImportantParam (Def set): DEV
SomethingElse (NonDef set): 456
AnotherOne (All sets):
```

Should probably include all missing parameters with default values after all to keep it similar to how PowerShell works. Thoughts?

tst/Pester.RSpec.ts.ps1
```diff
@@ -982,6 +982,130 @@ i -PassThru:$PassThru {
                 }
             }
         }
+
+        t "Undefined parameters with default values are automatically added to container Data" {
+            try {
+                $sb = {
+                    param (
```
fflaten, line 989: Maybe do a quick rewrite of the tests to keep simple integration-tests with `Invoke-Pester` here while checking the logic and different parametertypes etc. in `Pester.Runtime.ts.ps1` with `Invoke-Test` + `New-BlockContainerObject`?

src/Pester.Runtime.ps1
fflaten: Well, this thing took a complete u-turn. Even parameters without defaults are created as variables with null-value. So all script-parameters are available with a value (null as last resort) during Discovery, probably to avoid inheritance from parent scope.

I'm leaning towards wiping all the clever logic and just add all the parameters that the user forgot so Run-phase behaves the same as Discovery (and PowerShell in general) when we're first modifying the Data at all.

@Glober777 @nohwnd : What do you think?

Glober777: @fflaten, makes sense to me. Thank you for looking into this.

nohwnd: If that is how it works then we don't have another choice I am afraid. Only thing I am concerned about is backwards compat, and perf. But I don't see any way we could do it better, because we already get the variables the fastest way Imho. 

Another thing I just thought of is that we should handle aliases as well. If the data hashtable is our source of data and user has alias on the params the resulting variable could be different than what the specify. I also don't know what their expectation would be if they use different parameter sets. 

I also don't remember how it works for the root with data. Need to try it out to get back in picture.

fflaten: Thanks for the feedback and brainstorming.

> If that is how it works then we don't have another choice I am afraid. Only thing I am concerned about is backwards compat, and perf. 

The add-default function runs in 2ms in my demo with I believe I ran this with stopwatch-instance in the add-default-function and got less than a millisecond. Will double-check and add a condition to run only when there's 0 blocks in the Root-block 👍  As for passing the data to the base-level context/describe-blocks I'm not sure if there's much we can improve (or worry about).

Variables can be overridden in lower scopes if already used in something else, so the only backwards-compatibility issue I can think of atm would conflicting variables with `Should -Invoke -ParameterFilter`, but we'll add the logging in #1973 and mention `$PesterBoundParameters.<param>` in docs to help with those edge cases.

> Another thing I just thought of is that we should handle aliases as well. If the data hashtable is our source of data and user has alias on the params the resulting variable could be different than what the specify.

Just tested. Aliases aren't defined as variables inside a function, even if you use them when invoking the function. It's always the parameter name, which is what we parse + add here as well.

> I also don't know what their expectation would be if they use different parameter sets.

Probably the same as my expectation when I asked in discord #powershell-help; that the parameters I don't provide or the ones belonging to a different set are not defined as variables, but that's apparently not how PowerShell works. 😅

fflaten: Even tested private-scoped script parameters. They are only available in BeforeAll at root-level 👍

src/Pester.Runtime.ps1
```diff
@@ -2624,3 +2626,34 @@ function Invoke-InNewScriptScope ([ScriptBlock] $ScriptBlock, $SessionState) {
...
+    }
+
+    if ($null -ne $Ast -and $null -ne $Ast.ParamBlock -and $Ast.ParamBlock.Parameters.Count -gt 0) {
+        $parametersToCheck = foreach ($param in $Ast.ParamBlock.Parameters) { $param.Name.VariablePath.UserPath }
+
+        foreach ($param in $parametersToCheck) {
+            $v = $CallingFunction.SessionState.PSVariable.Get($param)
+            if ((-not $RootBlock.Data.ContainsKey($param)) -and $v) {
+                if ($PesterPreference.Debug.WriteDebugMessages.Value) {
+                    Write-PesterDebugMessage -Scope Discovery "Container parameter '$param' is undefined, adding to container Data with default value $(Format-Nicely $v.Value)."
```
fflaten, line 2653: Change to Runtime? Not sure which category/scope is correct here

nohwnd: yup

fflaten: fixed

fflaten:
I think this is ready for review. Wondered if it should have some tests when using `InModuleScope .. -Parameters ...` outside the blocks etc, but I'm not really sure how that's supposed to work even before this. Couldn't really pass any Data to run-phase, but that's probably user-error. I try to stay away from InModuleScope outside *All/*Each/It.

nohwnd (approved):

nohwnd:
I am not sure how that should work either. 😬

PR #1996, opened 2021-06-16
Azure Devops & Github Actions error logging

### LINKED ISSUE #1364: Output Azure DevOps Logging Commands in Passthru to help find failed tests and exceptions

We are running Pester as part of a Deploy Pipeline on Azure DevOps, in hundreds of tests if an issue happens in one, it is very difficult to find in full log output. 

Pester version     : 4.8.1 C:\Users\myusername\OneDrive - Value Retail Plc\Documents\WindowsPowerShell\Modules\Pester\4.8.1\Pester.psd1
PowerShell version : 5.1.17763.592

## 3. Expected Behavior

Azure DevOps provides some help for this. https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash 

If you write to the output with a tag 

##vso[task.logissue]error/warning message

It will highlight this in red and make it easy to find. 

## 4.Current Behavior

example log output I'm seeing today

Executing script C:\vstsagent\A1\_work\r1\a\_VegaDW-CI\Deployment Package\test\suites\character-test\v2\GL Forecast\FactGLBudgetBalance.Tests.ps1
2019-09-17T16:17:47.3761564Z 
2019-09-17T16:17:47.3775598Z   Describing GL Forecast
2019-09-17T16:17:47.4150045Z     [+] should have the correct sql count value in dbo.FactGLBudgetBalance 69ms
2019-09-17T16:17:47.4931722Z     [+] should have the correct sql values in dbo.FactGLBudgetBalance 78ms
2019-09-17T16:17:47.4996016Z VERBOSE: Connecting to server: 
2019-09-17T16:17:47.5006349Z "Provider=MSOLAP;DataSource=asazure://westeurope.asazure.windows.net/weu0assrv0dev0vegadw01;Integrated 
2019-09-17T16:17:47.5017659Z Security=ClaimsToken; Database="VR_BI_1864"; Connect Timeout=30; Connect Timeout=5"
2019-09-17T16:17:47.5340229Z     [-] Should have the correct cube values - GL Forecast by ForecastType 36ms
2019-09-17T16:17:47.6117988Z       ArgumentException: Authentication failed: User ID and Password are required when user interface is not available.
2019-09-17T16:17:47.6130509Z       at <ScriptBlock>, C:\vstsagent\A1\_work\r1\a\_VegaDW-CI\Deployment Package\test\suites\character-test\v2\GL Forecast\FactGLBudgetBalance.Tests.ps1: line 71

No colouring or anything in the output to distinguish errors from normal log output that the ##vso pattern would provide

## 5. Possible Solution

using ##vso tags as indicated above. Perhaps with a -ADO flag to ask pester inject the appropriate text at the beginning of the exception/failure stdout output?

Any other ideas on how I could make this happen as is or will it need a code change? 

## 6. Context

It is often useful to see the details in the log rather than using the Log output.

ArmaanMcleod:
## PR Summary

Fix #1364.

I've added functionality to highlight CI errors in red for ADO and GithubActions. 

This is configured from the `Output.CIFormat`, which covers the following options:

- **None** -> No CI autodetection
- **Auto** -> Autodetects CI
- **AzureDevops**
- **GithubActions**

For AzureDevops, we check the `TF_BUILD` environment variable is equal to `True`.
For GithubActions, we check the `GITHUB_ACTIONS` environment variable is equal to `True`. 

For **AzureDevops** we can use `##vso[task.logissue type=error]` to highlight the error header, and `##[error]` to highlight the messages. This ensures only the error header is reported in the build log. 

For **GithubActions**, you can only set an error, so I had to limit this to just the header, as shown below. The messages are also collapsed under the header using expandable groups.

### Azure Devops Output

ADO

### Build Log

ADOBuildLog

### Github Actions

GithubActions

### Build Log

GithubActionsBuildLog

I did testing in my own repo by running a simple github action to test out this code: https://github.com/ArmaanMcleod/TestGithubActions/actions/runs/952195814

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -899,18 +899,30 @@ function Invoke-Pester {
                 [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge($callerPreference, $Configuration)
             }
 
-            if ($PesterPreference.Output.CIFormat.IsOriginalValue() -or $PesterPreference.Output.CIFormat.Value -eq 'Auto') {
+            if ($PesterPreference.Output.CIFormat.Value -eq 'Auto') {
 
-                # Variable is set to 'True' if the script is being run by a build task. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
+                # Variable is set to 'True' if the script is being run by a Azure Devops build task. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
                 # Do not fix this to check for boolean value, the value is set to literal string 'True'
-                if ([System.Environment]::GetEnvironmentVariable('TF_BUILD') -eq 'True') {
+                if ($env:TF_BUILD -eq 'True') {
                     $PesterPreference.Output.CIFormat = 'AzureDevops'
                 }
                 # Variable is set to 'True' if the script is being run by a Github Actions workflow. https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
                 # Do not fix this to check for boolean value, the value is set to literal string 'True'
-                elseif ([System.Environment]::GetEnvironmentVariable('GITHUB_ACTIONS') -eq 'True') {
+                elseif ($env:GITHUB_ACTIONS -eq 'True') {
                     $PesterPreference.Output.CIFormat = 'GithubActions'
                 }
+
+                else {
+                    $PesterPreference.Output.CIFormat = 'None'
+                }
+            }
+
+            elseif ($PesterPreference.Output.CIFormat.Value -eq 'AzureDevops' -and $env:TF_BUILD -ne 'True') {
+                $PesterPreference.Output.CIFormat = 'None'
+            }
+
+            elseif ($PesterPreference.Output.CIFormat.Value -eq 'GithubActions' -and $env:GITHUB_ACTIONS -ne 'True') {
+                $PesterPreference.Output.CIFormat = 'None'
             }
 
             & $SafeCommands['Get-Variable'] 'Configuration' -Scope Local | Remove-Variable
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -917,14 +917,6 @@ function Invoke-Pester {
                 }
             }
 
-            elseif ($PesterPreference.Output.CIFormat.Value -eq 'AzureDevops' -and $env:TF_BUILD -ne 'True') {
-                $PesterPreference.Output.CIFormat = 'None'
-            }
-
-            elseif ($PesterPreference.Output.CIFormat.Value -eq 'GithubActions' -and $env:GITHUB_ACTIONS -ne 'True') {
-                $PesterPreference.Output.CIFormat = 'None'
-            }
-
             & $SafeCommands['Get-Variable'] 'Configuration' -Scope Local | Remove-Variable
 
             # $sessionState = Set-SessionStateHint -PassThru  -Hint "Caller - Captured in Invoke-Pester" -SessionState $PSCmdlet.SessionState
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -899,7 +899,7 @@ function Invoke-Pester {
                 [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge($callerPreference, $Configuration)
             }
 
-            if ($PesterPreference.Output.CIFormat.IsOriginalValue()) {
+            if ($PesterPreference.Output.CIFormat.IsOriginalValue() -or $PesterPreference.Output.CIFormat.Value -eq 'Auto') {
 
                 # Variable is set to 'True' if the script is being run by a build task. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
                 # Do not fix this to check for boolean value, the value is set to literal string 'True'
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1179,7 +1179,19 @@ function Invoke-Pester {
 
         }
         catch {
-            Write-ErrorToScreen $_ -Throw:$PesterPreference.Run.Throw.Value -StackTraceVerbosity:$PesterPreference.Output.StackTraceVerbosity.Value
+            $formatErrorParams = @{
+                Err                 = $_
+                StackTraceVerbosity = $PesterPreference.Output.StackTraceVerbosity.Value
+            }
+
+            if ($PesterPreference.Output.CIFormat.Value -in 'AzureDevops', 'GithubActions') {
+                $errorMessage = (Format-ErrorMessage @formatErrorParams) -split [Environment]::NewLine
+                Write-CIErrorToScreen -CIFormat $PesterPreference.Output.CIFormat.Value -Header $errorMessage[0] -Message $errorMessage[1..($errorMessage.Count - 1)]
+            }
+            else {
+                Write-ErrorToScreen @formatErrorParams -Throw:$PesterPreference.Run.Throw.Value
+            }
+
             if ($PesterPreference.Run.Exit.Value) {
                 exit -1
             }
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -899,6 +899,20 @@ function Invoke-Pester {
                 [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge($callerPreference, $Configuration)
             }
 
+            if ($PesterPreference.Output.CIFormat.IsOriginalValue()) {
+
+                # Variable is set to 'True' if the script is being run by a build task. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
+                # Do not fix this to check for boolean value, the value is set to literal string 'True'
+                if ([System.Environment]::GetEnvironmentVariable('TF_BUILD') -eq 'True') {
+                    $PesterPreference.Output.CIFormat = 'AzureDevops'
+                }
+                # Variable is set to 'True' if the script is being run by a Github Actions workflow. https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
+                # Do not fix this to check for boolean value, the value is set to literal string 'True'
+                elseif ([System.Environment]::GetEnvironmentVariable('GITHUB_ACTIONS') -eq 'True') {
+                    $PesterPreference.Output.CIFormat = 'GithubActions'
+                }
+            }
+
             & $SafeCommands['Get-Variable'] 'Configuration' -Scope Local | Remove-Variable
 
             # $sessionState = Set-SessionStateHint -PassThru  -Hint "Caller - Captured in Invoke-Pester" -SessionState $PSCmdlet.SessionState
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -398,6 +398,9 @@ function New-PesterConfiguration {
 
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
+
+      CIFormat: The CI format of output, options are None, Auto, AzureDevops and GithubActions.
+      Default value: 'Auto'
     ```
 
     .PARAMETER Hashtable
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -399,7 +399,7 @@ function New-PesterConfiguration {
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
 
-      CIFormat: The CI format of output in build logs, options are None, Auto, AzureDevops and GithubActions.
+      CIFormat: The CI format of error output in build logs, options are None, Auto, AzureDevops and GithubActions.
       Default value: 'Auto'
 
     TestDrive:
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -399,7 +399,7 @@ function New-PesterConfiguration {
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
 
-      CIFormat: The CI format of output in build logs, options are None, Auto, AzureDevops and GithubActions. Environment variable TF_BUILD(https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables-devops-services) is automically set to true when Azure Devops is detected. Environment variable GITHUB_ACTIONS(https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) is automatically set to true when Github Actions is detected.
+      CIFormat: The CI format of output in build logs, options are None, Auto, AzureDevops and GithubActions.
       Default value: 'Auto'
 
     TestDrive:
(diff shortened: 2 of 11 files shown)


ArmaanMcleod:
I guess another problem is that this floods the Azure Devops logs with error messages now. 
buildlog

ArmaanMcleod:
I will need to add some tests for this as well.

src/Main.ps1
```diff
@@ -926,6 +926,10 @@ function Invoke-Pester {
                 $PesterPreference.Output.StackTraceVerbosity = "Full"
             }
 
+            if ([Environment]::GetEnvironmentVariable("TF_BUILD") -eq 'True') {
```
nohwnd, line 929:
```suggestion
            # Variable is set to 'True' if the script is being run by a build task. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
            # Do not fix this to check for boolean value, the value is set to literal string 'True'
            if ([Environment]::GetEnvironmentVariable("TF_BUILD") -eq 'True') {
```

src/Pester.RSpec.ps1
```diff
@@ -398,6 +398,9 @@ function New-PesterConfiguration {
 
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
+
+      FormatAzureDevopsErrorMessages: Formats Azure Devops error messages with red color.
```
nohwnd, line 402:
```suggestion
      FormatAzureDevopsErrorMessages: Formats Azure Devops error messages with red color. This is automatically set to true when AzureDevOps run is detected, by checking that TF_BUILD environment variable is set to 'True'. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables
```

nohwnd (commented):
It might be nicer if the option was more generic, azdo can do also warning and info (IIRC). Even if this PR would implement just the errors, having just a single option for enabling the azdo formatted output would be better.

We should also ensure that user can override this in config to be disabled even if the env variable is set. Or vice versa.

nohwnd:
Changed this to be draft, because it is work in progress.

ArmaanMcleod:
> It might be nicer if the option was more generic, azdo can do also warning and info (IIRC). Even if this PR would implement just the errors, having just a single option for enabling the azdo formatted output would be better.
> 
> We should also ensure that user can override this in config to be disabled even if the env variable is set. Or vice versa.

Thanks for the feedback. Agree fully about supporting other log types. 

For overriding the option to `$false` even if `TF_BUILD` exists, should we include another option to ignore this environment variable? Thinking that might be easier, since the option is set to `$false` by default, and it would be difficult to determine how the env variable would get ignored without something else to override it. Perhaps I'm overcomplicating it, let me know your thoughts 👍

nohwnd:
There is `IsOriginalValue()` on the options on config for this purpose. 

```powershell
PS> $c = New-PesterConfiguration
PS> $c.CodeCoverage.Enabled.Value -eq $false
True
PS> $c.CodeCoverage.Enabled.IsOriginalValue()
True
PS> $c.CodeCoverage.Enabled = $false
PS> $c.CodeCoverage.Enabled.IsOriginalValue()
False
```

I look at this during the merging, so this is probably automatic, I think the setting to pester preference happens before merging it to user provided config, and in that case user provided value will take precedence if they set it explicitly.

nohwnd:
`FormatAzureDevopsLogMessages`

Sounds complicated, give me some time to come up with some better name. @fflaten do you have any opinion?

ArmaanMcleod:
Ah that's perfect, `IsOrignalValue()` will definitely solve the issue 👍 .

Yeah I don't like `FormatAzureDevopsLogMessages` either now that I'm using it. Happy to suggestions for a better name. Perhaps one improvement would be to replace `Format` with `Write`. Could also shorten `AzureDevops` to something like `ADO` or `AzDO`.

fflaten:
Do we need a ADO-specific option? Today it's Azure, tomorrow it's GitHub Actions (if we can use https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions from Pester), Travis CI etc. I'm not familiar all the different CI/CD systems. 

Would a enum like Output.CIFormat be better? Since it modifies the output while supporting only one value should be supported at a time (None being default).

fflaten:
Just tested, GitHub Actions format works:
image

So I'd personally expect that to be the next step as ADO (Pester's own CI) and GitHub Actions are equally popular.

src/functions/Output.ps1
```diff
@@ -773,6 +829,25 @@ function Get-WriteScreenPlugin ($Verbosity) {
...
+        [ValidateSet('Error', 'Warning', IgnoreCase)]
+        [string]
+        $LogType
+    )
+
+    foreach ($line in $Message) {
+
+        # https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell#logissue-log-an-error-or-warning
+        & $SafeCommands['Write-Host'] "##vso[task.logissue type=$LogType]$line"
+    }
```
fflaten, line 848: I think we should skip the foreach and only append to first line, typically `[-] mypath.to.failing.test`. The error flags are used in the report, so appending the special prefix to every message will fill the report with noise for the same error. Ex.
image

fflaten:
Can we limit the prefix to the first line only, typically the failing block/test (`[-] mypath.to.failing.test`)?
The error flags are used in the pipeline report as shortcuts to the log, so appending the special prefix to every message will fill the report with noise for the same error. Ex.
image

ArmaanMcleod:
> Would a enum like Output.CIFormat be better? Since it modifies the output while supporting only one value should be supported at a time (None being default).

I'm happy to create a `Output.CIFormat` option instead, which would be more generic for multiple CI providers. I'm thinking we can just support ADO and Github Actions like you said. 

I'm also wondering what levels of output would we support?

ADO supports the following:

```pwsh
Write-Host "##[section] This is colored green!"
Write-Host "##[command] This is colored blue!"
Write-Host "##[debug] This is colored purple!"
Write-Host "##[warning] This is colored orange!"
Write-Host "##vso[task.logissue type=warning;]This is colored orange!" 
Write-Host "##[error] This is colored red!"
Write-Host "##vso[task.logissue type=error;]This is colored red!"
```

Which looks like this:

CIOutput

For now it seems errors and warnings should definitely be supported. Should we support the other format types as well? Like Debug for instance. 

> Can we limit the prefix to the first line only, typically the failing block/test (`[-] mypath.to.failing.test`)?
> The error flags are used in the pipeline report as shortcuts to the log, so appending the special prefix to every message will fill the report with noise for the same error. Ex.

Also agree on this. The error logs get flooded with what I've currently done and limiting just the first line makes it less noisy.

nohwnd:
Option name `CIColor` ?


Coloring with just a single issue: `Write-Host "##vso[task.logissue type=error;]This is colored red!"` This for first line and then the color for the rest? `Write-Host "##[error] This is colored red!"`

Is original value, will probably by set after merging on the $PesterPreference, that we do few lines above. You need to move the code before  `& $SafeCommands['Get-Variable'] 'Configuration' -Scope Local | Remove-Variable` and check the value that user provided. And also the $callerPreference from the context. If either of those are set we should not overwrite with automatic value.

nohwnd:
Levels: I would support only error for now.

fflaten:
> Option name `CIColor` ?

Isn't this limiting? Color itself is better handled with ANSI as it's generic and supported in many CI-systems. 

The added benefit with this type of support is the integration with CI-summary/reports.

nohwnd:
AutodetectCI and specify what that means for each CI in the help?

ArmaanMcleod:
> AutodetectCI and specify what that means for each CI in the help?

So this autodetect option would be set to `true` by default? I guess that means disabling it might be simpler now since user can override to `false` and it won't autodetect. Of course even if its enabled we would still need to check if the env variable exists, and if it doesn't exist we set it to `false` and output a warning that no CI is detected? This seems like a simple solution to me.

fflaten:
> Is original value, will probably by set after merging on the $PesterPreference, that we do few lines above. You need to move the code before `& $SafeCommands['Get-Variable'] 'Configuration' -Scope Local | Remove-Variable` and check the value that user provided. And also the $callerPreference from the context. If either of those are set we should not overwrite with automatic value.

If the merged configuration falsely reports IsOriginalValue=false for options even though it wasn't modified in neither Configuration nor callerPreference, then that sounds more like a bug in Merger imho. I haven't tested it, but I'd expect it to not lie even after merging considering it only intends to modified already user-modified settings. We include it in the result-object so users should be able to detect modified/default options if they want.

> AutodetectCI and specify what that means for each CI in the help?

Sounds like it might conflict with the `-CI` switch some people are used to.  I'd like it to be a bit more explicit, but I won't argue anymore on this.

Personally I like 
```
Output.CIFormat = 'Auto' (default), 'None', 'AzureDevOps', 'GithubActions'
# As StringOption or EnumOption (to get faster validation of false input)

# Auto-detect if not specified
if ($PesterPreference.Output.CIFormat.IsOriginalValue()) {
    if ($env:thatAzDOdetectionThing) {
        $PesterPreference.Output.CIFormat = 'AzureDevOps'
    }
    elseif ($env:MagicGitHubActionsDetectionCondition {
        $PesterPreference.Output.CIFormat = 'GithubActions'
    }
}
```

fflaten:
TODO right before merging:
Disable in PesterPreference for RSpec tests since they fail on purpose.

ArmaanMcleod:
> TODO right before merging:
> Disable in PesterPreference for RSpec tests that fail on purpose.

Thanks @fflaten, I added some commits to disable this for deliberate failing tests. If I have to test and/or add more more changes I will stash these commits and apply them later. 

@nohwnd This is probably ready for review again. I've updated the PR description with the new changes.

Let me know what you guys think 🙂 .

src/Main.ps1
```diff
@@ -899,6 +899,32 @@ function Invoke-Pester {
                 [PesterConfiguration] $PesterPreference = [PesterConfiguration]::Merge($callerPreference, $Configuration)
             }
 
+            if ($PesterPreference.Output.CIFormat.Value -eq 'Auto') {
```
ArmaanMcleod, line 902: I didn't check `IsOriginalValue` here since its `Auto` by default. This will work for when its manually set or left by default.

src/Pester.RSpec.ps1
```diff
@@ -399,6 +399,9 @@ function New-PesterConfiguration {
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
 
+      CIFormat: The CI format of output in build logs, options are None, Auto, AzureDevops and GithubActions. Environment variable TF_BUILD(https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables-devops-services) is automically set to true when Azure Devops is detected. Environment variable GITHUB_ACTIONS(https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) is automatically set to true when Github Actions is detected.
```
ArmaanMcleod, line 402: I've added documentation here explaining what those environment vars do and why they are set to `True`. Let me know if I should be more explicit.

src/functions/Output.ps1
```diff
@@ -773,6 +829,69 @@ function Get-WriteScreenPlugin ($Verbosity) {
...
+
+    $CIOutputMapping = @{
+        'AzureDevops'   = @{
+            # header task issue error, so it gets reported to build log. https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell#example-log-an-error
+            Header        = "##vso[task.logissue type=error] $Header"
+
+            # message error prefx, but doesn't get reported in build log. Only task issue errors get reported.
+            MessagePrefix = '##[error] '
+        }
+        'GithubActions' = @{
```
ArmaanMcleod, line 855: Seems a bit hacky to format like this for GithubActions, but seems to align header with rest of message ok for now.

src/Main.ps1
```diff
@@ -899,6 +899,32 @@ function Invoke-Pester {
...
+            }
+
+            elseif ($PesterPreference.Output.CIFormat.Value -eq 'AzureDevops' -and $env:TF_BUILD -ne 'True') {
+                $PesterPreference.Output.CIFormat = 'None'
+            }
+
+            elseif ($PesterPreference.Output.CIFormat.Value -eq 'GithubActions' -and $env:GITHUB_ACTIONS -ne 'True') {
+                $PesterPreference.Output.CIFormat = 'None'
+            }
+
```
fflaten, line 927: The intention is good, but I would remove this. If the environment variables changes in the future, it's okay if auto-detect breaks, but explicit user settings should not break.

What do you think?

src/Pester.RSpec.ps1
fflaten: I'd keep this short and simple + accepted values. How we auto-detect and how it looks should be added to https://pester.dev. For example in https://github.com/pester/docs/pull/128

fflaten (commented):
Looks good 👍 Added some thoughts

src/Main.ps1
ArmaanMcleod: Yeah I see your point. I think we probably just remove `AzureDevops` and `GithubActions` options and keep just `Auto` and `None`. In that case might it better to just use a boolean flag like `AutoDetectCI` as we discussed before?

ArmaanMcleod: On the other hand keeping explicit CI options like this helps with determining what CI formatting to apply. I'd assume if we did `AutoDetectCI`, I would still need another property to set the CI that has been detected, since I currently pass this in into a hashtable with the CI header/message formatting.

ArmaanMcleod:
I've also realised I need to add this to `Main.ps1`, since there is a call to `Write-ErrorToScreen` in there that shows exceptions to the screen, which need to be shown in CI as well. Not sure why I missed that. 

Will mark this as draft and work on that 🙂

src/Main.ps1
ArmaanMcleod: I've removed these lines and the associated tests.

nohwnd:
I will have another look on this later this week.

ArmaanMcleod:
I made some changes to the GHA formatting, where the messages under the error header are collapsed into an expandable group. 

Screenshots are in the PR description.

tst/Pester.RSpec.Configuration.ts.ps1
```diff
@@ -623,11 +627,247 @@ i -PassThru:$PassThru {
...
+            $previousTfBuildVariable = $env:TF_BUILD
+            $previousGithubActionsVariable = $env:GITHUB_ACTIONS
+
+            $env:TF_BUILD = $true
+            $env:GITHUB_ACTIONS = $false
+
+            $r = Invoke-Pester -Configuration $c
+            $r.Configuration.Output.CIFormat.Value | Verify-Equal "AzureDevops"
+
+            $env:TF_BUILD = $previousTfBuildVariable
```
nohwnd, line 685: Those would not be set back if the test fails. You should use try finally to wrap the call to Invoke-Pester.

nohwnd:
The output looks great, and I am reasonably happy about the naming (can't come up with better one though). Only a small nit pick about the variables in tests, and then it's imho ready to merge. Will wait for the changes, and @fflaten approve.

tst/Pester.RSpec.Configuration.ts.ps1
ArmaanMcleod: Ah good catch, thanks for spotting that 👍 . I've wrapped those calls in `try/finally`.

nohwnd (approved):

fflaten:
Great work @ArmaanMcleod ! 🎉 

I've been busy the last days, but I didn't have anything to add in a review. It's mostly the option-design/name I'm not sure about, but I think we've settled on this for now. Maybe we get some feedback from other users with the now released alpha. 👍

PR #2008, opened 2021-06-19
Add option to disable TestDrive and TestRegistry

### LINKED ISSUE #1340: Version 4.8.1 Freezing in Describe

We run our normal jobs using Jenkins, which lacks impersonation.  I turned to the SecurityFever module and am using Push-Impersonation to impersonate the necessary service account.  From there Pester now locks up during the Describe function.  If I run a very vanilla Pester suite it's fine, but as soon as I start doing Sql queries with Invoke-SqlCmd I see the behavior.  This also occurs with 4.7.3 but works fine on 3.4.0 (that's my workaround for now).

Code to recreate the behavior:

```powershell
Import-Module Pester #-RequiredVersion 3.4.0
Import-module SqlServer

$Cred = Get-Credential -UserName "prod\serviceaccount" -Message "Need serviceaccount password"
Push-ImpersonationContext -Credential $Cred -LogonType Interactive

$Servers = @(
	"server1"
	"server2"
	"server3"
	"server4\Instance2"
)

Describe "Testing it" {	

	$Query = @"
SELECT s.name as Spec, 
s.is_state_enabled AS Enabled
FROM sys.server_audit_specifications AS s 
"@
	ForEach ($Server in $Servers)
	{
		$AuditSpec = Invoke-SqlCmd -ServerInstance $Server -Database Master -Query $Query
		If ($null -eq $AuditSpec)
		{
			It "Security set properly to query AuditSpc" {
				$AuditSpec | Should Not Be $null
			}
		}
		Else
		{
			It "$($Server): Audit is enabled" {
				$AuditSpec.Enabled | Should Be $true
			}
			If ($AuditSpec.Enabled)
			{
				It "Audit Spec should be ""AthenaServerAuditSpec""" {
					$AuditSpec.Spec | Should Match "AthenaServerAuditSpec"
				}
			}
		}
	}
}

Pop-ImpersonationContext
```

Pester version     : 4.8.1 C:\Program Files\WindowsPowerShell\Modules\Pester\4.8.1\Pester.psd1
PowerShell version : 5.1.14393.2969

## 3. Expected Behavior

Tests run normally.

## 4.Current Behavior

Pester freezes while running Describe

## 5. Possible Solution

For now I'm just using the older version of Pester.

## 6. Context

While not a show stopper for us, running on such an old version of Pester is a problem that will continue to get worse as time goes on.

### LINKED ISSUE #1344: Test script using impersonation fails with Access Denied when intialising TestDrive

When running a script which uses impersonation (e.g. issue #1340) and the impersonating user doesn't have Write permissions to the logged-on user's AppData folder, an exception will be thrown when Pester tries to initialise the TestDrive.

Note this issue isn't related to #1340 other than I found it while investigating it - I don't think it's the root cause of #1340.

```text
Describing example describe
New-Item : Access to the path 'C:\Users\My.User\AppData' is denied.
At C:\src\scratch\pester\issue-1340\pester\Functions\TestDrive.ps1:53 char:5
+     & $SafeCommands['New-Item'] -ItemType Container -Path $Path
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (C:\Users\My.User...2d-522094c52020:String) [New-Item], UnauthorizedAccessException
    + FullyQualifiedErrorId : CreateDirectoryUnauthorizedAccessError,Microsoft.PowerShell.Commands.NewItemCommand

Get-PSDrive : Cannot find drive. A drive with the name 'TestDrive' does not exist.
At C:\src\scratch\pester\issue-1340\pester\Functions\TestDrive.ps1:33 char:14
+     $Path = (& $SafeCommands['Get-PSDrive'] -Name TestDrive).Root
+              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (TestDrive:String) [Get-PSDrive], DriveNotFoundException
    + FullyQualifiedErrorId : GetLocationNoMatchingDrive,Microsoft.PowerShell.Commands.GetPSDriveCommand

  [-] Error occurred in Describe block 0ms
    ParameterBindingValidationException: Cannot bind argument to parameter 'Path' because it is null.
    at Clear-TestDrive, C:\src\scratch\pester\issue-1340\pester\Functions\TestDrive.ps1: line 34
    at DescribeImpl, C:\src\scratch\pester\issue-1340\pester\Functions\Describe.ps1: line 210
```

```text
Pester version     : 3.4.0 C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\Pester.psd1
PowerShell version : 5.1.17763.592
```
but note that Pester is being imported from a private folder at version 4.8.1

## 3. Expected Behavior

Pester should not throw an exception if the TestDrive cannot be initialised, so that scripts that do not use it can continue without exception.

## 4.Current Behavior

Pester currently throws an Access Denied exception trying to intialise the testDrive when the test script starts, even if the TestDrive is not needed in the test script.

## 5. Possible Solution

See #3 ^^^

## 6. Context

Some scripts using impersonation might fail unnecessarily if Pester cannot initialise the TestDrive when it is not used.

### LINKED ISSUE #2004: Implement an option to disable TestRegistry
We are working to run Pester inside Azure Functions, a blocker we have run into is that access to the Windows registry is restricted. When running Pester we receive the following error:

`System.Exception: Was not able to create a Pester Registry key for TestRegistry System.Security.SecurityException: Requested registry access is not allowed.`

This seems related to the TestRegistry feature, which we are not using. Similar to issue #1345

It would be helpful to have the option to disable TestRegistry on Pester runs. We have experimented with this successfully by just removing the following line:
https://github.com/pester/Pester/blob/989dac0620fd1312b4fa24ad111c9a48614f53ae/src/Main.ps1#L940

nohwnd:
## PR Summary

Fix #2004 
Fix #1340
Fix #1344

Add option to disable test drive and test registry. This will make the run faster if you are not using any of those capabilities. And also will allow runs that run in restricted machines such as ones that don't have access to registry.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -933,10 +933,13 @@ function Invoke-Pester {
                 # and that way output can consume the fixed object that decorator
                 # decorated, not nice but works
                 Get-RSpecObjectDecoratorPlugin
-                Get-TestDrivePlugin
             )
 
-            if ("Windows" -eq (GetPesterOs)) {
+            if ($PesterPreference.TestDrive.Enabled.Value) {
+                $plugins += @(Get-TestDrivePlugin)
+            }
+
+            if ($PesterPreference.TestRegistry.Enabled.Value -and "Windows" -eq (GetPesterOs)) {
                 $plugins += @(Get-TestRegistryPlugin)
             }
 
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -398,6 +398,14 @@ function New-PesterConfiguration {
 
       StackTraceVerbosity: The verbosity of stacktrace output, options are None, FirstLine, Filtered and Full.
       Default value: 'Filtered'
+
+    TestDrive:
+      Enabled: Enable TestDrive.
+      Default value: $true
+
+    TestRegistry:
+      Enabled: Enable TestRegistry.
+      Default value: $true
     ```
 
     .PARAMETER Hashtable
--- a/src/csharp/Pester/PesterConfiguration.cs
+++ b/src/csharp/Pester/PesterConfiguration.cs
@@ -33,6 +33,8 @@ public class PesterConfiguration
         cfg.Should = ShouldConfiguration.ShallowClone(configuration.Should);
         cfg.Debug = DebugConfiguration.ShallowClone(configuration.Debug);
         cfg.Output = OutputConfiguration.ShallowClone(configuration.Output);
+        cfg.TestDrive = TestDriveConfiguration.ShallowClone(configuration.TestDrive);
+        cfg.TestRegistry = TestRegistryConfiguration.ShallowClone(configuration.TestRegistry);
         return cfg;
     }
 
@@ -46,6 +48,8 @@ public class PesterConfiguration
         cfg.Should = Merger.Merge(configuration.Should, @override.Should);
         cfg.Debug = Merger.Merge(configuration.Debug, @override.Debug);
         cfg.Output = Merger.Merge(configuration.Output, @override.Output);
+        cfg.TestDrive = Merger.Merge(configuration.TestDrive, @override.TestDrive);
+        cfg.TestRegistry = Merger.Merge(configuration.TestRegistry, @override.TestRegistry);
         return cfg;
     }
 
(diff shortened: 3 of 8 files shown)


StephenFerrero:
Thank you!

PR #2020, opened 2021-07-02
Enable additional plugins to be added via $additionalPlugins

JustinGrote:
## PR Summary

Adds support for a `$SCRIPT:additionalPlugins` to be defined to allow outside modules to contribute their own plugins via this method:

```powershell
& (gmo pester) {$f = New-PluginObject etc..; $SCRIPT:additionalPlugins = @($f)
```

See: https://github.com/pester/Pester/discussions/2011#discussioncomment-953966

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1011,7 +1011,7 @@ function Invoke-Pester {
                 $pluginConfiguration["Coverage"] = $CodeCoverage
             }
 
-            if ($SCRIPT:additionalPlugins) {$plugins += $SCRIPT:additionalPlugins}
+            if (defined $script:additionalPlugins) {$plugins += $script:additionalPlugins}
 
             $filter = New-FilterObject `
                 -Tag $PesterPreference.Filter.Tag.Value `
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1012,7 +1012,7 @@ function Invoke-Pester {
             }
 
             # this is here to support Pester test runner in VSCode. Don't use it unless you are prepared to get broken in the future. And if you decide to use it, let us know in https://github.com/pester/Pester/issues/2021 so we can warn you about removing this.
-            if (defined $script:additionalPlugins) {$plugins += $script:additionalPlugins}
+            if (defined additionalPlugins) {$plugins += $script:additionalPlugins}
 
             $filter = New-FilterObject `
                 -Tag $PesterPreference.Filter.Tag.Value `
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1011,6 +1011,7 @@ function Invoke-Pester {
                 $pluginConfiguration["Coverage"] = $CodeCoverage
             }
 
+            # this is here to support Pester test runner in VSCode. Don't use it unless you are prepared to get broken in the future. And if you decide to use it, let us know in https://github.com/pester/Pester/issues/2021 so we can warn you about removing this.
             if (defined $script:additionalPlugins) {$plugins += $script:additionalPlugins}
 
             $filter = New-FilterObject `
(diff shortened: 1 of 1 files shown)


nohwnd:
This would not allow you to add a plugin in the middle of our built-in plugins, or remove other plugins, hope that is okay for your usecase. You could use a list of scriptblocks, that would take list of plugins and give you another list, but that gets complicated fast. 

Please add a note that this is a quick hack for test adapter, and that no-one should depend on it. And if they do then they should let us know in this issue: https://github.com/pester/Pester/issues/2021 to  help drive the future plugins public api.

src/Main.ps1
```diff
@@ -1011,6 +1011,8 @@ function Invoke-Pester {
                 $pluginConfiguration["Coverage"] = $CodeCoverage
             }
 
+            if ($SCRIPT:additionalPlugins) {$plugins += $SCRIPT:additionalPlugins}
```
nohwnd, line 1014:
```suggestion
            if ($script:additionalPlugins) {$plugins += $script:additionalPlugins}
```

JustinGrote: > This would not allow you to add a plugin in the middle of our built-in plugins, or remove other plugins, hope that is okay for your usecase. You could use a list of scriptblocks, that would take list of plugins and give you another list, but that gets complicated fast.
> 
> Please add a note that this is a quick hack for test adapter, and that no-one should depend on it. And if they do then they should let us know in this issue: #2021 to help drive the future plugins public api.


How about this approach then that basically lets you attach a callback to manipulate the plugins array, then you can do whatever you need to inject one or more plugins wherever you need. Obviously this isn't "safe" but it's pretty much nothing you can't do manipulating the module anyways :)

```powershell
  # Enable the ability to manipulate the plugin array to inject custom plugins via a callback
  # Example Usage:
  # & (gmo pester) {
  #     $SCRIPT:PluginHandler = {
  #     param($Plugins)
      
  #     $InjectionPoint = $Plugins.IndexOf($Plugins | Where-Object Name -eq 'RSpecObjectDecoratorPlugin')
  #     $Plugins.Insert($InjectionPoint, New-PluginObject -Name 'MyCustomPlugin')
  # }

  if ($SCRIPT:PluginHandler) {
      . $SCRIPT:PluginHandler -Plugins $Plugins
  }
```

nohwnd: Yeah something like what I had in mind. Have you tried it thought? For me it was pretty complicated to produce a plugin that would still bind correctly to the correct script and possibly have some external state. But maybe I was just going too deep into it.

JustinGrote: @nohwnd I haven't tried it via this specific method, but here's my current shim implementation that is "backwards compatible"

https://github.com/pester/vscode-adapter/blob/cf4a100401eae0582fa6f0614cf1a3ca6822c918/Scripts/PesterInterface.ps1#L289-L348

src/Main.ps1
```diff
@@ -1011,6 +1011,8 @@ function Invoke-Pester {
                 $pluginConfiguration["Coverage"] = $CodeCoverage
             }
 
+            if (defined $script:additionalPlugins) {$plugins += $script:additionalPlugins}
```
nohwnd, line 1014:
```suggestion
            # this is here to support Pester test runner in VSCode. Don't use it unless you are prepared to get broken in the future. And if you decide to use it, let us know in https://github.com/pester/Pester/issues/2021 so we can warn you about removing this.
            if (defined $script:additionalPlugins) {$plugins += $script:additionalPlugins}
```

nohwnd:
```suggestion
            if (defined additionalPlugins) {$plugins += $script:additionalPlugins}
```

PR #2023, opened 2021-07-03
Added Run.SkipRemainingOnFailure option

### LINKED ISSUE #360: How to stop Pester at first failed test?

ArmaanMcleod:
## PR Summary

Fix #360

I've added a `Run.SkipRemainingOnFailure` option. The following scopes are supported:

**Block** -> Skip all remaining tests in current block(including child tests) after a failed test.
**Container** -> Skip all remainng tests in the container(file or scriptblock) after a failed test.
**Run** -> Skip all tests across all containers in a run after a failed test.
**None** -> default, keep original behaviour.

### Sample Test 1

```pwsh
# C:\Users\Armaan\Documents\powershell-dev\sample.Tests.ps1
Describe "d1" { 
    Describe "d2" {
        It "i2" {
            $true | Should -BeTrue
        }
        Describe "d3" { 
            It "i3" {
                $false | Should -BeTrue # Fail
            }
            It "i4" {
                $true | Should -BeTrue
            }
        }
        It "i5" {
            $true | Should -BeTrue
        }
    }
    Describe "d4" {
        It "i6" {
            $true | Should -BeTrue
        }
    }
}
```

### Sample Test 2

```pwsh
# C:\Users\Armaan\Documents\powershell-dev\sample2.Tests.ps1
Describe "d5" { 
    It "i7" {
        $false | Should -BeTrue # Fail
    }
    It "i8" {
        $false | Should -BeTrue # Fail
    }
}
```

### Setup code

```pwsh
$pesterConfig = New-PesterConfiguration
$pesterConfig.Run.Container = @((New-PesterContainer -Path C:\Users\Armaan\Documents\powershell-dev\sample.Tests.ps1), (New-PesterContainer -Path C:\Users\Armaan\Documents\powershell-dev\sample2.Tests.ps1))
$pesterConfig.Output.Verbosity = "Detailed"
```

### Block

```pwsh
$pesterConfig.Run.SkipRemainingOnFailure = 'Block'
Invoke-Pester -Configuration $pesterConfig
```

Block

### Container

```pwsh
$pesterConfig.Run.SkipRemainingOnFailure = 'Container'
Invoke-Pester -Configuration $pesterConfig
```

Container

### Run

```pwsh
$pesterConfig.Run.SkipRemainingOnFailure = 'Run'
Invoke-Pester -Configuration $pesterConfig
```

Run

## PR Checklist

- [x] PR has meaningful title


ArmaanMcleod:
This should be ready for review. 

I've attached screenshots in the PR description to show behaviour of the different options. Let me know if I've interpreted something incorrectly 👍 .

Also added the warning to the skipped tests. Not sure if it adds too much noise, but seems useful to see why the tests were skipped. Perhaps It would be better to limit this to just one message at the bottom? It might also be too generic, and could probably also say which test failed instead of just `Skipped due to previous failure and Run.SkipRemainingOnFailure set to <option>`. Didn't want to overcomplicate this so just left it as is for now 🙂 .

src/functions/Output.ps1
```diff
@@ -477,6 +477,83 @@ function ConvertTo-HumanTime {
...
+                    if (-not $test.Executed) {
+                        $test.Skipped = $true
+                    }
+                }
+
+                foreach ($test in ($Context.Block | View-Flat)) {
+                    if (-not $test.Executed) {
+                        $test.Skipped = $true
+                    }
+                }
```
fflaten, line 503: Setting `Skipped = $true` will report skipped, but won't actually stop the tests and setups from being executed. Try `$test.Skip = $true`.

src/functions/Output.ps1
```diff
@@ -724,6 +805,10 @@ function Get-WriteScreenPlugin ($Verbosity) {
                 if ($PesterPreference.Output.Verbosity.Value -in 'Detailed', 'Diagnostic') {
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped "$margin[!] $out" -NoNewLine
                     & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.SkippedTime " $humanTime"
+
+                    if ($PesterPreference.Run.SkipRemainingOnFailure.Value -in 'Block', 'Container', 'Run') {
+                        & $SafeCommands['Write-Host'] -ForegroundColor $ReportTheme.Skipped "$margin Skipped due to previous failure and Run.SkipRemainingOnFailure set to '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'"
+                    }
```
fflaten, line 811: Will print message (false positive) when users use `-Skip` and plugin is active.  Ex. `It 'test1' -Skip { }`. No way to split plugin-caused skip from user-caused skip.

Only `Set-ItResult` stores the reason why skip was set by adding a PesterTestSkipped-error record to the test-object. Ex.

```powershell
$test.ErrorRecord.Add([Pester.Factory]::CreateErrorRecord( 
     'PesterTestSkipped', #string errorId 
     $Message, #string message 
     $File, #string file 
     $Line, #string line 
     $LineText, #string lineText 
     $false #bool terminating 
 ))
```

Should we drop the printed output and add a similar error-record to the tests being marked to skip? Could also mention the failing test that caused it.

src/functions/Output.ps1
```diff
@@ -477,6 +477,83 @@ function ConvertTo-HumanTime {
     }
 }
 
+function Get-SkipRemainingOnFailurePlugin ($SkipRemainingOnFailure) {
```
fflaten, line 480: Move plugin to it's own file? Not related to screen output.

fflaten (commented):
Good start! Added some thoughts

src/functions/Output.ps1
```diff
@@ -670,6 +747,10 @@ function Get-WriteScreenPlugin ($Verbosity) {
             throw "Unsupported CI format '$($PesterPreference.Output.CIFormat.Value)'"
         }
 
+        if ($PesterPreference.Run.SkipRemainingOnFailure.Value -notin 'None', 'Block', 'Container', 'Run') {
+            throw "Unsupported SkipRemainingOnFailure option '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'"
+        }
+
```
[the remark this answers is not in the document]
ArmaanMcleod: Yeah since `Get-WriteScreenPlugin` throws this error inside `EachTestTearDown` for the output options, I might do the same for my plugin.

src/functions/Output.ps1
ArmaanMcleod: Ah yeah was going to ask about this. I saw both while debugging, make sense to have one for reporting and one for setting test in skipped state 👍

src/functions/Output.ps1
ArmaanMcleod: Yeah wanted to do this, since it doesn't belong in `Output.ps1`. Will make sure to separate this into a separate script 👍

src/functions/Output.ps1
ArmaanMcleod: Yeah I think dropping printed output would be a good idea here, I completely forgot about `-Skip` for tests. Adding to error record seems like a better idea, which could contain the failing test inside `Message`.

ArmaanMcleod: I can probably create another function to set add this error record, with an errorID of something like `pesterRemainingTestSkipped`? Just to differentiate the fully qualified error id from the one `Set-ItResult` sets.

src/functions/Output.ps1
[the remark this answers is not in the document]
ArmaanMcleod: Yeah wanted to do this but it only seems to work when put in `Get-WriteScreenPlugin`? If I do it in my own plugin in `RunStart` it shows a huge stack trace:

```pwsh
if ($PesterPreference.Run.SkipRemainingOnFailure.Value -notin 'None', 'Block', 'Container', 'Run') {
    $p.RunStart = {
        throw "Unsupported SkipRemainingOnFailure option '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'"
    }
}
```

```pwsh
Invoking step RunStart failed:
Result :Error :
Exception: C:\Users\Armaan\Documents\github-repos\Pester\src\functions\Get-SkipRemainingOnFailurePlugin.ps1:8:13
Line |
   8 |              throw "Unsupported SkipRemainingOnFailure option '$($Pest …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Unsupported SkipRemainingOnFailure option 'Non'

at <ScriptBlock>, C:\Users\Armaan\Documents\github-repos\Pester\src\functions\Get-SkipRemainingOnFailurePlugin.ps1: line 8
at Invoke-PluginStep, C:\Users\Armaan\Documents\github-repos\Pester\src\Pester.Runtime.ps1: line 1243
at Run-Test, C:\Users\Armaan\Documents\github-repos\Pester\src\Pester.Runtime.ps1: line 1043
at Invoke-Test, C:\Users\Armaan\Documents\github-repos\Pester\src\Pester.Runtime.ps1: line 1950
at Invoke-Pester<End>, C:\Users\Armaan\Documents\github-repos\Pester\src\Main.ps1: line 1059
```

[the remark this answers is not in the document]
ArmaanMcleod: This doesn't happen though when the option checking is in `Get-WriteScreenPlugin` like what I've already done. Perhaps the ordering of plugins if affecting this?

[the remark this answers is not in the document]
ArmaanMcleod: What I could do is just throw the error inside `Invoke-Pester` like so when adding the plugin:

```pwsh
if ($PesterPreference.Run.SkipRemainingOnFailure.Value -notin 'None', 'Block', 'Container', 'Run') {
    throw "Unsupported SkipRemainingOnFailure option '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'"
}
else {
    $plugins += @(Get-SkipRemainingOnFailurePlugin)
}
```

src/functions/Output.ps1
ArmaanMcleod: @fflaten Would you be happy with something like the following being added:

```pwsh
$failingTestPath = $Context.Test.ExpandedPath
$failingTestLine = $Context.Test.ScriptBlock.StartPosition.StartLine

$errorRecord = [Pester.Factory]::CreateErrorRecord(
    'PesterRemainingTestSkipped',
    "Skipped due to previous failure ($($failingTestPath):$($failingTestLine)) and Run.SkipRemainingOnFailure set to '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'",
    $test.ScriptBlock.File,
    $test.ScriptBlock.StartPosition.StartLine,
    $test.ExpandedPath,
    $false
)
$test.ErrorRecord.Add($errorRecord)
```

This would add the failing test path and number to the message, and File/Line/LineText would just contain the info of the current skipped test.

src/functions/Output.ps1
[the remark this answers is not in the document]
fflaten: I see. It's not about the plugin, but the plugin-step being invoked when you throw. The setup/teardown-steps are called using `Invoke-ScriptBlock` which has a different error-handling logic than the general steps like RunStart which are called with `Invoke-PluginStep`. The latter re-throws the exception.

If you use `$p.EachTestTeardownEnd = { throw 'demo' }` in your plugin it should look the same. At that point, we've already executed too much though. If we're gonna throw, we might as well do it prior to discovery etc. to avoid wasting the user's time. I'm not sure why the Output-validation is done so late tbh. Do you remember @nohwnd ?

If you validate in `Main.ps1` or as part of `Get-SkipRemainingOnFailurePlugin`  (to keep it isolated) then you can throw when the plugin is generated. It's not a single-line error, but simple enough. Ex.

```powershell
# code
function Get-SkipRemainingOnFailurePlugin {
    ....
    elseif ($PesterPreference.Run.SkipRemainingOnFailure.Value -notin 'None', 'Block', 'Container', 'Run') {
        $p.EachTestSetupStart = { throw "Unsupported Run.SkipRemainingOnFailure option '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'" }
    }

    New-PluginObject @p
}

# demo
PS /workspaces/Pester/Samples> /workspaces/Pester/Samples/demoSkipRemaining.ps1
System.Management.Automation.RuntimeException: Unsupported Run.SkipRemainingOnFailure option 'Blocks'
at Get-SkipRemainingOnFailurePlugin, /workspaces/Pester/src/functions/Get-SkipRemainingOnFailurePlugin.ps1: line 76
at Invoke-Pester<End>, /workspaces/Pester/src/Main.ps1: line 966
at <ScriptBlock>, /workspaces/Pester/Samples/demoSkipRemaining.ps1: line 38
```

src/functions/Output.ps1
fflaten: Not sure if errorid matters, but I'd probably use `PesterTestSkipped` to be consistent. This is a user-controlled skip just like `Set-ItResult`, only with a different trigger. That reason is shown in the error-message.

The error-record should include `$Context.Test.ExpandedPath` which might help identify the testcase that failed if they used variables in the test or block-names. I don't think File, StartPosition and Scriptblock provides any value in the record since they belong to tests that hasn't been executed. If not null, they could point at the file/startpos/line for the original error that caused the failed test, but that might be bad practice (referencing different code in a error). Thoughts @nohwnd ?

Btw. `$test.StartLine` is preferred over `$test.ScriptBlock.Startline` as the latter will return the line of `{` which might be multiple lines down when using testcases. `$test.StartLine` should always return the line of `It ...`

src/functions/Output.ps1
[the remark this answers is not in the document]
ArmaanMcleod: Thanks for this, really insightful 🙂. After stepping through the code I ran into similar behaviour, which makes me understand more how the setup/teardown steps work in the background. 

For now I'm just handling this in `Main.ps1`, which will just throw the exception like what you've shown.

src/functions/Output.ps1
ArmaanMcleod: > Not sure if errorid matters, but I'd probably use PesterTestSkipped to be consistent. This is a user-controlled skip just like Set-ItResult, only with a different trigger. That reason is shown in the error-message.

Yeah I've kept it as `PesterTestSkipped`, which now seems to also show the skipped message next to the test inside my update PR description screenshots. 

> $Context.Test.ExpandedPath might be a better reference to the failed test as it can help identity a specific testcase (if variables are used in name). I don't think File, StartPosition and Scriptblock provides any value in the record since they belong to tests that hasn't been executed. If not null, they could point at the file/startpos/line for the original error that caused the failed test, but that might be bad practice (referencing different code in a error). Thoughts @nohwnd ?

Also removed File,StartPosition and Scriptblock from the error record and populating with `$null` for now. Happy to put in more suitable values here if needed. 

> Should it output a single line at the end of the run in Detailed+Diagnostic to mention that one or more tests were skipped due to Run.Skip... ? Or is it unnecessary noise considering it's a explicit opt-in behavior?

I feel like that since this will be opt-in behavior, the user would probably want as little noise as possible. I'm still wondering what we should do with the skipped message that is added from `PesterTestSkipped` error record being added. I feel like if a 100 tests are running this is going to get really noisy real fast, making me think a message at the bottom for Detailed/Diagnostic is probably the most feasible option.

fflaten: > I also see that this message is always added to `PesterTestSkipped` error records: https://github.com/ArmaanMcleod/Pester/blob/main/src/functions/Output.ps1#L645-L653. I think its really useful to add this error record, especially when using `PassThru`. 

Is that actually executed? Will have to test later, but I wouldn't expect it since it's part of test teardown. The first test is marked as failed, the rest are marked to be skipped which shouldn't invoke any setup, test or teardown for them.

fflaten: You're right. TIL the plugin steps are are actually called even though the setup/teardowns/tests/blocks themselves will be skipped after. Makes sense considering the steps are called ex. EachTestSetup**Start**, EachTestTeardown**End** etc., but still not sure if it's intentional or just a result of calling EachTestSetupStart using the same `Invoke-ScriptBlock` as OneTimeTestSetupStart.

It's easy enough to exclude with `if(-not $_test.Skip -and $_test.ErrorRecord.FullyQualifiedErrorId -eq 'PesterTestSkipped') { ...` in the Output-plugin - or use a different errorid ofc. I'm curious to know what @nohwnd thinks about this. Currently:
- `It 't1' -Skip` will only output `[!] t1`
- `Set-ItResult` will include a message/reason
- What should this do? Since it's opt-in and "planned" like `-Skip`, then ignoring the message might be the enough. This also aligns with the recent change to reduce noise in Detailed and Diagnostic. Optionally include a single message somewhere (RunEnd, ContainerEnd, BlockEnd depending on option? Or just at the end?)

src/functions/Get-SkipRemainingOnFailurePlugin.ps1
```diff
@@ -0,0 +1,110 @@
+function New-SkippedTestMessage {
+    [OutputType([string])]
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory)]
+        [Pester.Test]
+        $Test
+    )
+    "Skipped due to previous failure at path:line '$($Test.ExpandedPath):$($Test.StartLine)' and Run.SkipRemainingOnFailure set to '$($PesterPreference.Run.SkipRemainingOnFailure.Value)'"
```
fflaten, line 9: `$Test.ExpandedPath` is DescribeName.ContextName.TestName so ExpandedPath:StartLine looks a bit weird:
`[!] c Skipped due to previous failure at path:line 'a.b myvarvalue:9' and Run.SkipRemainingOnFailure set to 'Block'`

src/functions/Get-SkipRemainingOnFailurePlugin.ps1
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    elseif ($PesterPreference.Run.SkipRemainingOnFailure.Value -eq 'Run') {
+        $p.EachTestSetupStart = {
+            param($Context)
+
+            # It test has failed at some point during the run
+            # Skip the test before it runs
+            # This handles skipping tests that failed from different containers in the same run
+            if ($script:failedTest) {
```
fflaten, line 85: This variable will persist across runs, so the next run will skip all tests. `$script` in Pester-state will be stored as a module variable until Pester is reimported or the variable is overwritten.

I believe this type of run-scoped state is what PluginConfiguration is created for. You can use it like: `$Context.Configuration.SkipRemainingFailedTest = $Context.Test`  +++.

src/functions/Get-SkipRemainingOnFailurePlugin.ps1
ArmaanMcleod: Yeah this is a good point.  I originally had it in there to make it easier to find the failing test, but don't think we need it since expanded path shows everything we need. Also makes tests easier to write since I don't need to worry about line numbers 👍

src/functions/Get-SkipRemainingOnFailurePlugin.ps1
ArmaanMcleod: Ah yeah forgot about this. I was in the habit of importing before a run so I didn't catch this. Will use pluginconfiguration to store this state instead.

src/functions/Output.ps1
ArmaanMcleod: I think excluding is probably easiest for now. I can also add a message to Detailed/Diagnostic output at the end if needed. Up to you guys 🙂 .

ArmaanMcleod:
I think this is ready for another review. Thanks @fflaten for reviewing this change up till now, really helpful feedback 🙂 . 

One of the conversations above regarding how to handle skipped messages has not been resolved, leaving this for the maintainer @nohwnd to comment on since we are not sure if a message should be included at the bottom for Detailed/Diagnostic view. 

Other than this, happy to hear more feedback for anything else that needs to be changed.

nohwnd:
@ArmaanMcleod  sorry I am slammed with work, and don't have any time for Pester nowadays. I will review as soon as I can.

ArmaanMcleod:
@nohwnd No worries 🙂. I am also getting slammed at work. Must be that busy time of year again 😄

src/functions/Output.ps1
nohwnd: > but still not sure if it's intentional or just a result of calling EachTestSetupStart using the same Invoke-ScriptBlock as OneTimeTestSetupStart
@fflaten  do you mean this part? That runs the script in two different scopes. I don't think I intentionally did it this way. But imho it is correct. The  Skipped state is just a short circuiting the execution of the test.

https://github.com/pester/Pester/blob/main/src/Pester.Runtime.ps1#L567-L579

nohwnd:
@ArmaanMcleod I think adding a message to the bottom of detailed + diag output would be nice. Much like when a block fails there is summary at the bottom. We can change the behavior later as we learn how useful or tragically oververbose (like my original filter output in diag) it is.

ArmaanMcleod:
> @ArmaanMcleod I think adding a message to the bottom of detailed + diag output would be nice. Much like when a block fails there is summary at the bottom. We can change the behavior later as we learn how useful or tragically oververbose (like my original filter output in diag) it is.

Would something like this be ok
skipped

src/functions/Output.ps1
fflaten: > @fflaten  do you mean this part? That runs the script in two different scopes. I don't think I intentionally did it this way. But imho it is correct. The  Skipped state is just a short circuiting the execution of the test.
> 
> https://github.com/pester/Pester/blob/main/src/Pester.Runtime.ps1#L567-L579

Yeah, EachTestSetupStart + EachTestTeardownEnd are invoked even though BeforeEach/AfterEach won't be executed (due to skip-detection in the next lines). Nothing we need to solve here, it was just unexpected for someone unexperienced with the plugin-system. 👍

I'll probably post a few issues with some questions + demo around this later considering the plugin-system will be made public at some point. As part of that initiative the code and docs needs to be clear about when to expect which step to be executed. IIRC there's even some steps that's currently not invoked at all, including RunEnd that I suggested above. 🙂

nohwnd:
LGTM.

PR #2031, opened 2021-07-20
Future improvement for coverage exclusion, fix some scan warnings

asears:
## PR Summary

This introduces a new property to the csharp project CodeCoverageConfiguration.ExcludePaths.  It can be used for further implementation of configuration property in the Powershell code to exclude paths from Code Coverage results.  

Example feature for implementation would  be a script in the folder is picked up for coverage, however it is a Pester test harness script or other Powershell script not under test and not requiring coverage.

A couple minor updates for excluding Write-Host check from the scanner, spelling in the PSScriptAnalyzer custom check, fixing null check and alias scan warning.

## PR Checklist

- [ ] PR has meaningful title

### DIFF
--- a/PSScriptAnalyzerSettings.psd1
+++ b/PSScriptAnalyzerSettings.psd1
@@ -6,5 +6,6 @@
         'PSUseShouldProcessForStateChangingFunctions'
         'PSUseApprovedVerbs'
         'Measure-SafeCommands'
+        'PSAvoidUsingWriteHost'
     )
 }
--- a/src/csharp/Pester/CodeCoverageConfiguration.cs
+++ b/src/csharp/Pester/CodeCoverageConfiguration.cs
@@ -26,6 +26,7 @@ namespace Pester
         private StringOption _outputPath;
         private StringOption _outputEncoding;
         private StringArrayOption _path;
+        private StringArrayOption _excludePaths;
         private BoolOption _excludeTests;
         private BoolOption _recursePaths;
         private BoolOption _useBps;
@@ -46,6 +47,7 @@ namespace Pester
             OutputPath = new StringOption("Path relative to the current directory where code coverage report is saved.", "coverage.xml");
             OutputEncoding = new StringOption("Encoding of the output file.", "UTF8");
             Path = new StringArrayOption("Directories or files to be used for codecoverage, by default the Path(s) from general settings are used, unless overridden here.", new string[0]);
+            ExcludePaths = new StringArrayOption("Directories or files to be excluded from codecoverage.", new string[0]);
             ExcludeTests = new BoolOption("Exclude tests from code coverage. This uses the TestFilter from general configuration.", true);
             RecursePaths = new BoolOption("Will recurse through directories in the Path option.", true);
             UseBreakpoints = new BoolOption("EXPERIMENTAL: When false, use Profiler based tracer to do CodeCoverage instead of using breakpoints.", true);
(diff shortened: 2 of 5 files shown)


src/functions/Coverage.ps1
```diff
@@ -87,7 +88,7 @@ function Exit-CoverageAnalysis {
     # to only get those that are not $null
     # (like if we did $breakpoints | where {$_ -ne $null})
     # so DON'T change this.
-    $breakpoints = @($CommandCoverage.Breakpoint) -ne $null
+    $breakpoints = null -ne @($CommandCoverage.Breakpoint)
```
ArmaanMcleod, line 91: null should be $null

src/csharp/Pester/Coverage.cs
```diff
@@ -23,6 +23,7 @@ public static CodeCoverage Create()
         public List<object> CommandsMissed = new List<object>();
         public List<object> CommandsExecuted = new List<object>();
         public List<object> FilesAnalyzed = new List<object>();
+        public List<object> FIlesExcluded = new List<object>();
```
ArmaanMcleod, line 26:
```suggestion
        public List<object> FilesExcluded = new List<object>();
```

src/functions/Coverage.ps1
ArmaanMcleod, line 91:
```suggestion
    $breakpoints = $null -ne @($CommandCoverage.Breakpoint)
```

src/csharp/Pester/CodeCoverageConfiguration.cs
```diff
@@ -46,6 +47,7 @@ public CodeCoverageConfiguration() : base("CodeCoverage configuration.")
             OutputPath = new StringOption("Path relative to the current directory where code coverage report is saved.", "coverage.xml");
             OutputEncoding = new StringOption("Encoding of the output file.", "UTF8");
             Path = new StringArrayOption("Directories or files to be used for codecoverage, by default the Path(s) from general settings are used, unless overridden here.", new string[0]);
+            ExcludePaths = new StringArrayOption("Directories or files to be excluded from codecoverage.", new string[0]);
```
ArmaanMcleod, line 50: Might be good to mention what default behavior is, just like Path option.

PSScriptAnalyzerSettings.psd1
```diff
@@ -6,5 +6,6 @@
         'PSUseShouldProcessForStateChangingFunctions'
         'PSUseApprovedVerbs'
         'Measure-SafeCommands'
+        'PSAvoidUsingWriteHost'
```
ArmaanMcleod, line 9: Wouldn't this raise a lot of PSScriptAnalyzer warnings all over the codebase? We couldn't simply replace `Write-Host` with `Write-Output` given there is no `-BackgroundColor/-ForegroundColor/-NoNewLine` options for that cmdlet. Probably best doing this work in another issue anyways, doesn't seem to belong here in my opinion.

fflaten: I guess that's why it's set under excludedrules. 🙂

ArmaanMcleod:
Added some comments. I can probably help with the powershell implementation if you want. 

From first glance I think we will need to filter the files in https://github.com/pester/Pester/blob/main/src/functions/Coverage.ps1#L206.

ArmaanMcleod:
Added some comments. I can probably help with the powershell implementation if you want. 

From first glance I think we will need to filter the files in https://github.com/pester/Pester/blob/main/src/functions/Coverage.ps1#L206.

PSScriptAnalyzerSettings.psd1
ArmaanMcleod: Oops, I guess thats what I get for looking at code late in the night 😄

asears: It would be great if someone could put in a PR to turn off that rule in PSScriptAnalyzer by default.

asears:
@ArmaanMcleod I would appreciate someone taking the powershell implementation on after this PR gets merged.

asears:
Not clear why the tests are failing with this PR...

src/functions/Coverage.ps1
fflaten: Did you read the comment above this line?

ArmaanMcleod:
> @ArmaanMcleod I would appreciate someone taking the PowerShell implementation on after this PR gets merged.

To me it makes sense to have the C# and PowerShell code changes in this PR, since then you can add tests to validate the option is working correctly. @nohwnd What do you think?

Issue #2033, opened 2021-07-20
Pester configuration 'TestResult.TestSuiteName' sets name in element 'test-results' instead of name in first element 'test-suite'


Marc013:
## General summary of the issue

When specifying the `TestSuiteName` in the Pester configuration (`New-PesterConfiguration`) the name is added to object _name_ in element _test-results_ instead of object _name_ the first element _test-suite_.

I use the following code to generate the Pester configuration

```PowerShell
    $Configuration = New-PesterConfiguration
    $Configuration.Run.Path = $TestsPath
    $Configuration.CodeCoverage.Enabled = $true
    $Configuration.CodeCoverage.OutputPath = $CodeCoverageOutputPath
    $Configuration.CodeCoverage.Path = $ScriptPath
    $Configuration.TestResult.Enabled = $true
    $Configuration.TestResult.OutputFormat = 'NUnitXml'
    $Configuration.TestResult.OutputPath = $TestResultOutputPath
    $Configuration.TestResult.TestSuiteName = 'MarvelIsBetterThanDC!'
    $Configuration.Should.ErrorAction = 'Continue'
    $Configuration.Output.Verbosity = 'Detailed'
```

## Describe your environment

> Pester version     : 5.2.2 C:\Users\Marc013\OneDrive\Documents\PowerShell\Modules\Pester\5.2.2\Pester.psm1
> PowerShell version : 7.1.3
> OS version         : Microsoft Windows NT 10.0.19043.0

## Steps to reproduce

Configure a Pester configuration including `TestResult.TestSuiteName`. Run Pester defining the configuration and see the results in the XML file.

## Expected Behavior

See `name` in element `test-suite`

```XML
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<test-results xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="nunit_schema_2.5.xsd" name="MarvelIsBetterThanDC!" total="5" errors="0" failures="1" not-run="0" inconclusive="0" ignored="0" skipped="0" invalid="0" date="2021-07-20" time="20:19:01">
  <environment machine-name="MYLAPTOP" cwd="C:\" user-domain="AzureAD" os-version="10.0.19043" user="Marc013" nunit-version="2.5.8.0" clr-version="Unknown" platform="Microsoft Windows 10 Enterprise|C:\WINDOWS|\Device\Harddisk0\Partition3" />
[...]

PR #2034, opened 2021-07-21
Set name in xml element testsuite

Marc013:
## PR Summary

Fixes [https://github.com/pester/Pester/issues/2033](https://github.com/pester/Pester/issues/2033)

_Pester configuration 'TestResult.TestSuiteName' sets name in element 'test-results' instead of name in first element 'test-suite'_

### DIFF
--- a/src/functions/TestResults.ps1
+++ b/src/functions/TestResults.ps1
@@ -299,7 +299,7 @@ function Write-NUnitReport($Result, [System.Xml.XmlWriter] $XmlWriter) {
     $xmlWriter.WriteStartElement('test-results')
 
     Write-NUnitTestResultAttributes @PSBoundParameters
-    Write-NUnitTestResultChildNodes @PSBoundParameters
+    Write-NUnitTestResultChildNodes -RunResult $Result -XmlWriter $XmlWriter
 
     $XmlWriter.WriteEndElement()
 }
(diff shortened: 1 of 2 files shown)


nohwnd:
Please add a test for this so we avoid regressions.

src/functions/TestResults.ps1
```diff
@@ -325,6 +325,7 @@ function Write-NUnitTestResultChildNodes($RunResult, [System.Xml.XmlWriter] $Xml
     Write-NUnitCultureInformation -Result $RunResult -XmlWriter $XmlWriter
 
     $suiteInfo = Get-TestSuiteInfo -TestSuite $Result -Path "Pester"
+    $suiteInfo.name = $RunResult.Configuration.TestResult.TestSuiteName.Value
```
fflaten, line 328: You can merge this if you want. `-Path` is used for name and description IIRC.

Marc013:
@nohwnd 
I would love to make a unit test however, I have no clue how I can unit test the change I have implemented in function ‘_Write-NUnitTestResultChildNodes_’.
Can you provide any help/guidance here?

nohwnd:
There are tests for this feature in https://github.com/pester/Pester/blob/main/tst/Pester.RSpec.Nunit.TestResults.ts.ps1 those are P tests, P is simplified test framework that we use to test some aspects of Pester. You should be able to simply run the whole file. If that does not work run .\test.ps1 from the root of the repo first.

In that file you add a test, they use i, b and t functions, i is for invoke, b is for block (like Describe in Pester) and t is for test (like It in Pester). You can focus a single test by using dt instead of t to make development easier. Hopefully reading few other tests in the file will make it clear what is happening. You run a Pester run, and then compare the resulting XML to ensure it was produced correctly.

Marc013:
@nohwnd,
The unit test is added. I hope it's to your satisfaction.

fflaten:
Thanks @Marc013 . I made a few adjustments to clean up the "RunResult"-parameters + simplified the test. Also, didn't want to scare any DC-fans from contributing 😄

Issue #2036, opened 2021-07-22
Wording can be more explicit about needing to enable TestResult and CodeCoverage to get output


simonsabin:
## Location
https://pester.dev/docs/commands/New-PesterConfiguration#edit-this-page

## General summary of the issue
When you call invoke-pester with  OutputFile the TestResulEnabled is set to true. However when you use the Configuration, even if you've set the outptupath you also need to set the Enabled flag. 

The doc could be more specific , and/or if the outputPath is set to a non default value then the enabled flag is set.

nohwnd:
This is resolved in Pester 6: output is auto-enabled when you set an OutputPath, so the confusing 'set OutputPath but also remember to set Enabled' gap is gone. Thanks for flagging it, @simonsabin. Closing.

PR #2037, opened 2021-07-22
making the language a bit more specific about the enabled flags

simonsabin:
## PR Summary
Makes the docs more explicit about the need to set the enabled flag.
#2036

If your pull request integrates Pester with another system, please tell us how the change can be tested. -->

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -332,7 +332,7 @@ function New-PesterConfiguration {
       Default value: @()
 
     CodeCoverage:
-      Enabled: Enable running of CodeCoverage.
+      Enabled: Enable CodeCoverage.
       Default value: $false
 
       OutputFormat: Format to use for code coverage report. Possible values: JaCoCo, CoverageGutters
@@ -363,13 +363,13 @@ function New-PesterConfiguration {
       Default value: $true
 
     TestResult:
-      Enabled: Enable export of test results to a file.
+      Enabled: Enable the exporting of TestResults to file specified in `OutputPath`.
       Default value: $false
 
-      OutputFormat: Format to use for test result report. Possible values: NUnitXml, NUnit2.5 or JUnitXml
+      OutputFormat: Format to use for test result report. Possible ves: NUnitXml, NUnit2.5 or JUnitXml
       Default value: 'NUnitXml'
 
-      OutputPath: Path relative to the current directory where test result report is saved. Enabled has to be set to true to have test results exported
+      OutputPath: Path relative to the current directory where test result report is saved.
       Default value: 'testResults.xml'
 
       OutputEncoding: Encoding of the output file.
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -332,7 +332,7 @@ function New-PesterConfiguration {
       Default value: @()
 
     CodeCoverage:
-      Enabled: Enable CodeCoverage.
+      Enabled: Enable running of CodeCoverage.
       Default value: $false
 
       OutputFormat: Format to use for code coverage report. Possible values: JaCoCo, CoverageGutters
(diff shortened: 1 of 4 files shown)


src/Pester.RSpec.ps1
```diff
@@ -363,13 +363,13 @@ function New-PesterConfiguration {
...
     TestResult:
-      Enabled: Enable TestResult.
+      Enabled: Enable export of test results to a file.
       Default value: $false
 
       OutputFormat: Format to use for test result report. Possible values: NUnitXml, NUnit2.5 or JUnitXml
       Default value: 'NUnitXml'
 
-      OutputPath: Path relative to the current directory where test result report is saved.
+      OutputPath: Path relative to the current directory where test result report is saved. Enabled has to be set to true to have test results exported
```
nohwnd, line 372:
```suggestion
      OutputPath: Path relative to the current directory where test result report is saved. Enabled has to be set to $true to have test results exported.
```

But maybe it would be better to write this next to the first option? It is on the top, and you don't need to repeat this on every option in this section.

nohwnd:
Those lines are generated from the C# code. You would need to change them there. https://github.com/pester/Pester/blob/main/src/csharp/Pester/CodeCoverageConfiguration.cs#L44

simonsabin:
I assume same applies to https://github.com/pester/Pester/blob/main/src/csharp/Pester/TestResultConfiguration.cs, and the RSpec needs to match.

simonsabin:
What are your thoughts on changing https://github.com/pester/Pester/blob/6285de4c82bf61065d80395b770cfd31fa37b631/src/csharp/Pester/TestResultConfiguration.cs#L90 to set Enabled if the value isn't the default?

nohwnd:
Yes all the entries in the options list on New-PesterConfiguration are generated from the C# source in the `build.ps1 -clean` build.

> What are your thoughts on changing to set Enabled if the value isn't the default?

Let's do it only if the same makes sense for all the other options that use the same pattern. E.g. Codecoverage. We should also check if use did not set the Enabled value explicitly to false by checking the isoriginalvalue property.

simonsabin:
Before I make broad changes, can you check the latest changes to see if that aligns with your thoughts.

src/csharp/Pester/TestResultConfiguration.cs
```diff
@@ -130,9 +138,20 @@ public StringOption TestSuiteName
...
+                    EnableIfOriginalValue(_testSuiteName);
+
                     _testSuiteName = new StringOption(_testSuiteName, value?.Value);
                 }
             }
         }
+
+        public void EnableIfOriginalValue(Option optionBeingSet)
+        {
+            if (optionBeingSet.IsOriginalValue() && Enabled.IsOriginalValue())
```
nohwnd, line 150: Why is the first check needed, I don't think you are testing it.

tst/Pester.RSpec.Configuration.ts.ps1
```diff
@@ -268,6 +268,38 @@ i -PassThru:$PassThru {
             $config.Run.Path = $config.Run.Path.Default
             $config.Run.Path.IsOriginalValue() | Verify-False
         }
+
+        t "TestResult.Enabled is set if TestResult.OutputPath not originalValue even with the Default value" {
+            $config = [PesterConfiguration]::Default
+            $config.TestResult.Enabled.Value | Verify-False
+            $config.TestResult.OutputPath = $config.TestResult.OutputPath.Default
```
nohwnd, line 275: This seems to be unnecessarily complex. Looking at this test in a month I feel like I will focus on the "default value" being used rather than on what you are actually testing. 

I would rename the test to: TestResult is automatically enabled when OutputPath is set

And below

TestResult is not automatically enabled  when user already set Enabled to $false

nohwnd: Making sure that IsOriginalValue works even with the default value, is also nice but imho not the main purpose of this test. And I think we can safely skip it.

fflaten:
Friendly reminder @simonsabin 😇

splatteredbits:
I was just coming to report/request that the `TestResult.Enabled` flag should be automatically set to `true` if *any* of the `TestResult` properties are set *and* `Enabled` hasn't been explicitly set.. I can't tell you how many times I've forgotten to set enabled and discover weeks later that we haven't been capturing test results.

Any chance this can get reviewed, completed, approved, and merged?

simonsabin:
will look at it next week when back in office

PR #2042, opened 2021-07-28
Add more logging to BP CC and fix case sensitive new CC

### LINKED ISSUE #2041: Make new CC case insensitive when checking for file
The dictionaries we use in Tracer code are case sensitive, and because we are passing the file beforehand we might fail to recognize the hit when we invoke the script with a different casing. 

This is further complicated by Windows file system being case insensitive, so we are able to invoke files like that, but on Linux with case sensitive file system, we could end up merging multiple results incorrectly for files that only differ in case. 

Check if we can resolve the file better, to get the correct casing, but fail when the FS is case insensitive, to make it less error prone, but also correct.

---

This also uncovered a bug in the new coverage, I am not checking the path case insensitively, I am not sure if this is a bug, or powershell simply passes the file as is. I need to decide if we should check the path case insensitively always (more convenient), or check case sensitively (will lead to weird issues like this on Windows / other non-case insensitive file systems). Imho having script files that only differ in case is edge case, that should be avoided, and checking the path case insensitively is the better way to go. On Linux user will fail to run the file (unless they actually have multiple that differ in case, and then they would get broken "merged" CC results).

_Originally posted by @nohwnd in https://github.com/pester/Pester/issues/2019#issuecomment-888199486_

nohwnd:
Make tracer based CC case insensitive for file paths, because PowerShell will report the path the way it is given, including any casing differences when you invoke the script using . or &.

Fix #2041

### DIFF
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -261,7 +261,7 @@ function Write-PesterDebugMessage {
     [CmdletBinding(DefaultParameterSetName = "Default")]
     param (
         [Parameter(Mandatory = $true, Position = 0)]
-        [ValidateSet("Filter", "Skip", "Runtime", "RuntimeCore", "Mock", "MockCore", "Discovery", "DiscoveryCore", "SessionState", "Timing", "TimingCore", "Plugin", "PluginCore", "CodeCoverage")]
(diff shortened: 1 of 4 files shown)


nohwnd:
I am just making it case insensitive in all cases now. If we run into problems we can add check if filesystem is also case insensitive, to adjust this.

Issue #2100, opened 2021-11-03
Casting hashtable to PesterConfiguration fails to convert ['Run']['Path'] when its an ArrayList


splatteredbits:
## General summary of the issue
When casting a hashtable to a `PesterConfiguration` object, `Collections.ArrayList` objects aren't converted. Properties retain their default value.

## Describe your environment
```
Pester version     : 5.3.1 C:\********\PSModules\Pester\5.3.1\Pester.psm1
PowerShell version : 7.1.5
OS version         : Microsoft Windows NT 10.0.19043.0
```

## Steps to reproduce
This code:
```
$paths = [Collections.ArrayList]::New();
[void]$paths.Add('one')
[void]$paths.Add('two')
[void]$paths.add('three')
([PesterConfiguration]@{ Run = @{ Path = $paths } }).Run.Path
```
outputs 

```

Default Description                                                                                 Value
------- -----------                                                                                 -----
{.}     Directories to be searched for tests, paths directly to test files, or combination of both. {.}
```

As you can see, the value is `{.}`. I expected the value to be `{one, two, three}`.

Now, normally, this wouldn't be a problem. Just case `$paths` to a `[String[]]` when assigning it to the `Paths` property. Unfortunately, this doesn't work if you're running Pester in a background job. PowerShell converts *all* arrays passed to a background job to a `Collections.ArrayList` object:

```
$array = @('one','two','three')
Start-Job { ($using:array).GetType() } | Receive-Job -Wait -AutoRemoveJob
```
returns
```

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     ArrayList                                System.Object
```

## Expected Behavior
I would expect that when converting a hashtable to a `PesterConfiguration` object, Pester handles when an object is an array list or maybe any other type of enumerable.

## Current Behavior
Values that are `Collections.ArrayList` are treated as not set.

## Possible Solution? (optional)

```
if( $value -is [IEnumerable] )
{
   $convertedValue = $value | Write-Output
}
```

fflaten:
Thanks. This isn't specific to configuration casted from dictionaries. StringArray configuration-options like `Run.Path` only support conversion/casting from values or arrays of `string` or `PathInfo` atm. Attempting to use `ArrayList`, `List` etc would've thrown if you'd tried it directly on a `PesterConfiguration`-object. Ex.

```powershell
> $l = [System.Collections.Generic.List[string]]::new(); $l.Add("a"); $l.Add("b")
[...]

PR #2101, opened 2021-11-04
Fixed: casting a hashtable to a PesterConfiguration object fails to convert Run.Path when it is an ArrayList

splatteredbits:
Fixes issue #2100.

### DIFF
--- a/tst/PesterConfiguration.Tests.ps1
+++ b/tst/PesterConfiguration.Tests.ps1
@@ -26,3 +26,12 @@ Describe "PesterConfiguration.Format.ps1xml" {
         }
     }
 }
+
+Describe 'PesterConfiguration' {
+    It 'should convert arraylists' {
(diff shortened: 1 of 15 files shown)


tst/PesterConfiguration.Tests.ps1
```diff
@@ -26,3 +26,12 @@ Describe "PesterConfiguration.Format.ps1xml" {
         }
     }
 }
+
+Describe 'PesterConfiguration' {
+    It 'should convert arraylists' {
+        $expectedPaths = @('one', 'two', 'three')
+        $pathList = [Collections.ArrayList]::New($expectedPaths)
```
nohwnd, line 33: This is not PowerShell 3 compatible syntax.
```suggestion
        $pathList = [Collections.ArrayList]$expectedPaths
```

PR #2164, opened 2022-04-22
Test on less legacy systems

nohwnd:


nohwnd:
Let's see if I can get the 1 minute build back, instead of 15 minute build because of PS5 on W2016.

PR #2189, opened 2022-06-20
Always return to original CWD after running Pester

### LINKED ISSUE #2076: Set-Location $null run slowly and High CPU usage
`Describe 'root' {it 'test' {Set-Location $null}}`

Run this script and wait for the execution to complete. Running this script again should also return the test results immediately, but the actual situation is that PowerShell will not be able to interact, and the CPU utilization begins to rise abnormally. Wait a few minutes before you can receive the return value, and may be accompanied by an error message.The same conclusion can be reached by testing in new virtual machines.

```PowerShell
Get-ChildItem: C:\Users\Andy\Documents\PowerShell\Modules\Pester\5.3.0\Pester.psm1:3512
Line |
3512 |              & $SafeCommands['Get-ChildItem'] -Recurse -Path $p -Filte …
     |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Could not find a part of the path 'C:\Users\Andy\Documents\Tencent
     | Files\11681258\FileRecv'.
```

```PowerShell
Pester version     : 5.3.0 C:\Users\Andy\Documents\PowerShell\Modules\Pester\5.3.0\Pester.psm1
PowerShell version : 7.1.3
```

fflaten:
Returns to the original PWD/CWD at end of `Invoke-Pester` in case test code has changed it.

Fix #2076

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1,4 +1,4 @@
-﻿function Assert-ValidAssertionName {
+function Assert-ValidAssertionName {
     param([string]$Name)
     if ($Name -notmatch '^\S+$') {
         throw "Assertion name '$name' is invalid, assertion name must be a single word."
@@ -633,6 +633,9 @@ function Invoke-Pester {
         $script:mockTable = @{}
         # todo: move mock cleanup to BeforeAllBlockContainer when there is any
         Remove-MockFunctionsAndAliases -SessionState $PSCmdlet.SessionState
+
+        # store CWD so we can revert any changes at the end
+        & $SafeCommands['Push-Location'] -StackName 'Pester'
     }
 
     end {
@@ -1213,6 +1216,9 @@ function Invoke-Pester {
                 exit -1
             }
         }
+
+        # go back to original CWD
+        & $SafeCommands['Pop-Location'] -StackName 'Pester'
 
         # exit with exit code if we fail and even if we succeed, otherwise we could inherit
         # exit code of some other app end exit with it's exit code instead with ours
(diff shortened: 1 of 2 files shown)


src/Main.ps1
```diff
@@ -633,6 +633,9 @@ function Invoke-Pester {
         $script:mockTable = @{}
         # todo: move mock cleanup to BeforeAllBlockContainer when there is any
         Remove-MockFunctionsAndAliases -SessionState $PSCmdlet.SessionState
+
+        # store CWD so we can revert any changes at the end
+        & $SafeCommands['Push-Location'] -StackName 'Pester'
```
nohwnd, line 638: Seems a bit heavy. Maybe `$script:currentPwd = $pwd`?

src/Main.ps1
```diff
@@ -1214,6 +1217,9 @@ function Invoke-Pester {
             }
         }
 
+        # go back to original CWD
+        & $SafeCommands['Pop-Location'] -StackName 'Pester'
```
nohwnd, line 1221: I think doing this in non-stateful way with a variable would be better. This is expensive, Pester-in-Pester can possibly push on that stack, and simply doing` if ($null -ne $script:currentPwd) { cd... }` seems like less problematic code.

fflaten: So I guess we got a choice between:
- Push/Pop, but change StackName to `Pester_<$state.Id>` and add a guid to `New-PesterState`. RunId might be useful in the future.
- Add `$state.InitialDirectory = $pwd.Path` and only `Set-Location` at the end

nohwnd: The last option is used in other places as well, where we read variable that was defined in parent scope, to e.g. figure out if we are running Pester in Pester. As long as the scope that returns to the final folder is at the same scope level or below, this is probably the best choice.

PR #2199, opened 2022-06-27
Add ANSI support for output

### LINKED ISSUE #1953: Using ANSI escape codes for output
Using ANSI escape codes would probably also fix output not being colored in CI, which is very unfortunate:

<details>
<summary>Travis CI</summary>

image

https://travis-ci.org/felixfbecker/PSKubectl/jobs/500233965

</details>
<details>
<summary>VSTS</summary>

image

https://felixfbecker.visualstudio.com/PowerGit/_build/results?buildId=179

</details>

_Originally posted by @felixfbecker in https://github.com/pester/Pester/issues/1265#issuecomment-495924708_

fflaten:
## PR Summary
Introduces a configuration option `Output.RenderMode` and a wrapper-function to replace `Write-Host` so we can output strings using ANSI-escaped sequences to get colors in CI or plaintext etc. The option supports the following levels:

- Auto (default):
  - Sets to Plaintext if $env:NO_COLOR
  - Sets to Ansi if $host.UI.SupportsVirtualTerminal
  - Else, Legacy (same as today)
- Ansi
  - To force it. PSv3/v4 doesn't support auto-detect but CI might. Ex. our Azure Pipeline CI in this repo
- Legacy (same as before this PR)
  - For those who want color in GUI, but not in CI or redirected output to logs etc.
- Plaintext (why not.. https://no-color.org/)

The new function `Write-PesterHostMessage` supports all parameters of `Write-Host` to make it an easy replacement.

Fix #1953

TODO:
- [x] Fix auto-detection
   - [x] V3/V4 compatible
   - [x] PS6_2 run in CI should not use Ansi
- [x] Add tests for `Write-PesterHostMessage`
- [x] Add tests for config option and auto-detection
- [x] Test Azure DevOps and GitHub Actions
  - [x] Fix multiline-strings
- [x] Test PSJobs (remoting)
- [x] Support custom PSHosts without UI
  - Works for PS3 and PS4 now. Output only available in PS5+ where Information-stream exists
- [x] Support NO_COLOR

## PR Checklist

- [x] PR has meaningful title
  - Website-update tracked in pester/docs#214


fflaten:
Alternative configuration using `StringArrayOption Output.RenderMode`:
- **Auto** (default):
  - Set to Plaintext if `$env:NO_COLOR`
  - Set to Ansi if `$host.UI.SupportsVirtualTerminal`
  - Else, default (ConsoleColor like today)
- **Ansi** (force it, because PSv3/v4 doesn't support auto-detect but CI might (like our Azure Pipeline checks in this PR)
- **ConsoleColor** (for those who want color in GUI, but not in CI or redirected output to logs etc)
- **Plaintext** (idk.. https://no-color.org/)

@nohwnd @beatcracker @briantist Thoughts? Overcomplicating it? 😄

briantist:
> Alternative configuration using `StringOption Output.RenderMode`:
> 
> * **Auto** (default):
>   
>   * Set to Plaintext if `$env:NO_COLOR`
>   * Set to Ansi if `$host.UI.SupportsVirtualTerminal`
>   * Else, default (ConsoleColor like today)
> * **Ansi** (force it, because PSv3/v4 doesn't support auto-detect but CI might (like our Azure Pipeline checks in this PR)
> * **ConsoleColor** (for those who want color in GUI, but not in CI or redirected output to logs etc)
> * **Plaintext** (idk.. https://no-color.org/)
> 
> @nohwnd @beatcracker @briantist Thoughts? Overcomplicating it? 😄

I like the options as you have them laid out here. It doesn't seem overcomplicated to me, it seems thorough and well thought out. I didn't know about no-color and I think it's a nice idea and should be supported.

nohwnd:
The options seem more straight forward to me, if we add a "conflicting" FQFQ (or whatever) output option in the future, we don't have to solve what happens where there are both UseAnsi and UseFQFQ.

The breaking change: 

Is writing via console directly necessary to implement this? Or is that just for perf?

fflaten:
Ready for review. Good thing we squash the merges here 😄

src/Main.ps1
```diff
@@ -922,6 +922,25 @@ function Invoke-Pester {
...
+            if ($PesterPreference.Output.RenderMode.Value -eq 'Auto') {
+                if ($null -ne $env:NO_COLOR) {
+                    # https://no-color.org/)
+                    $PesterPreference.Output.RenderMode = 'Plaintext'
+                }
+                elseif (($supportsVT = $host.UI.psobject.Properties['SupportsVirtualTerminal']) -and $supportsVT.Value) {
+                    $PesterPreference.Output.RenderMode = 'Ansi'
+                }
+                else {
+                    $PesterPreference.Output.RenderMode = 'Legacy'
```
nohwnd, line 940: Not a big fan of "Legacy" options over concrete option name, do we use similar naming in other places as well?  I can't remember.

src/functions/Output.ps1
```diff
@@ -64,6 +64,111 @@ $script:ReportTheme = DATA {
...
+        $NoNewLine,
+
+        $Separator = ' ',
+
+        [ValidateSet('Ansi', 'Legacy', 'Plaintext')]
+        [string]
+        $RenderMode = $PesterPreference.Output.RenderMode.Value
+    )
+
+    begin {
```
nohwnd, line 89: I am not using begin often enough, this will run on start of every call to the command right? Maybe it is worth moving this outside of the function so we don't re-create the hashtable for every write to output?

You could also store the color codes in Dictionary<ConsoleColor, string>, which is probably a bit faster than keying by string, or you could put the colors in string array, and index into that based on the enum value.

nohwnd:
Really nice, some small suggestions.

src/functions/Output.ps1
fflaten: > I am not using begin often enough, this will run on start of every call to the command right? 

Correct, once even if multiple items through pipeline. 

> Maybe it is worth moving this outside of the function so we don't re-create the hashtable for every write to output?

I originally did that, but didn't measure any notable difference (last year). Can test again, including some different collections. 👍

src/Main.ps1
fflaten: Only parameter set name of the top of my head.

Suggestions? `HostOnly` or `Console` felt like still could be ANSI and possible complaint. In case we add other styles in the future like bold, underline etc.

`ConsoleColor` was a bit technical, but maybe better? `Compatibility` maybe?

src/functions/Output.ps1
fflaten: We save a total of 20ms on 1000 runs by having it pre-defined. Hashtable with string-keys vs consolecolor-key vs compiled in c# (`readonly static Dictionary<consolecolor,string>`) have no impact (+/- 2ms). Tested in 5.1 and 7.2

I'd stick with this for now. I can clean it up with ConsoleColor-keys to avoid the `ToString()`-call

nohwnd: > Since there's no real perf benefit, you can decide based on code style.

Keep it as is please, since there is no benefit to get from putting it outside of the function.

> I'll switch to ConsoleColor-keys just to avoid the ToString()-call
👍

src/Main.ps1
nohwnd: I think ConsoleColor is the one to use, it is a bit technical, but we can explain in docs somewhere in 1 line. The meaning of "Legacy" will just deteriorate, and then soon there will be Legacy2, or Legacy will be used for 10 different purposes.

nohwnd (approved):

nohwnd:
Merged, thanks for that.

fflaten:
Thanks! Finally done with this conflict-magnet 😄 🎉

PR #2201, opened 2022-06-27
Replace IsOriginalValue() with IsModified and fix merging and casting of default option values

### LINKED ISSUE #2198: IsOriginalValue() always returns False after merging
### Checklist

### What is the issue?

After merging two configurations (or using `$PesterPreference`), `IsOriginalValue()` will return False even if both configurations had original values for an option.

### Expected Behavior

Should return True or False depending on whether its actually modified

### Steps To Reproduce

```powershell
$one = New-PesterConfiguration
$two = New-PesterConfiguration
$merge = [PesterConfiguration]::Merge($one,$two)
$merge.Output.CIFormat.IsOriginalValue()
False
```

### Describe your environment

_No response_

### Possible Solution?

_No response_

fflaten:
## PR Summary
Only modify configuration options that have actually been changed when casting and merging configurations.

`Option.IsOriginalValue()` is marked obsolete and is replaced by a new `Option.IsModified`-property that enables us to fix merging of serialized PesterConfiguration.

Fix #2198

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/csharp/Pester/GenericOption.cs
+++ b/src/csharp/Pester/GenericOption.cs
@@ -21,12 +21,12 @@ namespace Pester
     {
         public Option(Option<T> option, T value) : this(option.Description, option.Default, value)
         {
-            _isOriginalValue = false;
+            _isModified = true;
         }
 
         public Option(string description, T defaultValue, T value)
         {
-            _isOriginalValue = true;
+            _isModified = false;
             Default = defaultValue;
             Value = value;
             Description = description;
--- a/src/csharp/Pester/Option.cs
+++ b/src/csharp/Pester/Option.cs
@@ -21,15 +21,15 @@ namespace Pester
 {
     public abstract class Option
     {
-        protected bool _isOriginalValue;
+        protected bool _isModified;
 
         [Obsolete("IsOriginalValue() is obsolete and should no longer be used. Use the property IsModified instead.")]
         public bool IsOriginalValue()
         {
-            return _isOriginalValue;
+            return !_isModified;
         }
         public bool IsModified {
-            get { return !_isOriginalValue; }
+            get { return _isModified; }
         }
     }
 }


fflaten:
@nohwnd:
- Thoughts on this?
- Any reason not to replace `IsOriginalValue()` with `IsOriginalValue`-property with only a getter? That would expose the state when serializing and deserializing `PesterConfiguration`-objects which should let us keep the status after deserialization. The new formats will still keep it hidden in default view.
  - Or keep both if we believe someone's using the method outside runtime-code

src/csharp/Pester/CodeCoverageConfiguration.cs
```diff
@@ -57,17 +57,16 @@ public CodeCoverageConfiguration(IDictionary configuration) : this()
...
-                OutputPath = configuration.GetObjectOrNull<string>("OutputPath") ?? OutputPath;
-                OutputEncoding = configuration.GetObjectOrNull<string>("OutputEncoding") ?? OutputEncoding;
-                Path = configuration.GetArrayOrNull<string>("Path") ?? Path;
-                ExcludeTests = configuration.GetValueOrNull<bool>("ExcludeTests") ?? ExcludeTests;
-                RecursePaths = configuration.GetValueOrNull<bool>("RecursePaths") ?? RecursePaths;
-                CoveragePercentTarget = configuration.GetValueOrNull<decimal>("CoveragePercentTarget") ?? CoveragePercentTarget;
-
-                SingleHitBreakpoints = configuration.GetValueOrNull<bool>("SingleHitBreakpoints") ?? SingleHitBreakpoints;
-                UseBreakpoints = configuration.GetValueOrNull<bool>("UseBreakpoints") ?? UseBreakpoints;
+                if (configuration.GetValueOrNull<bool>(nameof(Enabled)) is bool enabled) { Enabled = enabled; };
```
nohwnd, line 60: This looks like a lot of repeated code that can be done wrong in one or two places. I would rather see this in a method, and using `is not null` or `!= null` rather than `is <type>`.

IMHO you are checking to see if the value is not null, because GetValueOrNull already guarantees the given T type is returned.

Nice to see return to nameof, I had that everywhere but for some reason that I don't remember it did not work, and I reverted to just names. Maybe I was trying to avoid shipping dll, and compile in-line.

```cs
configuration.AssignValueIfNotNull<bool>(nameof(Enabled), v => Enabled = v);

public static void AssignValueIfNotNull<T>(this IDictionary dictionary, string key, Action<T> assign) 
where T : struct
{
    var value = GetValueOrNull<T>(dictionary, key);
    if (value != null)
    {
        assign(value.Value);
    }
}

```

fflaten: > IMHO you are checking to see if the value is not null, because GetValueOrNull already guarantees the given T type is returned.

Fair point. Shortsighted in the heat of the moment. Had to get something to work for the Ansi-option which I'll now probably skip. Will still fix this though for future use.

Love the generic extension-method. 💯 will steal 😄

fflaten:
Ready for review.
I could change to `IsDefaultValue` if you'd like to avoid mini-breaking change and keep `IsOriginalValue()`.

nohwnd:
IsDefaultValue seems like better name. Please put IsOriginalValue() back and mark it deprecated (with Obsolete attribute). Just to be nice about semantic versioning.

fflaten:
> IsDefaultValue seems like better name.

After letting the name sink in, it could confuse someone when it's false while Default and Value are equal (but modified). 😅. Naming is hard. `IsModified`?

nohwnd (approved):

PR #2217, opened 2022-07-14
Start interactive execution from BeforeDiscovery

### LINKED ISSUE #2092: Test fails when no breakpoint is set, test passes with a breakpoint set
This is a bit of strange bug but it is consistent. When we run `Invoke-Pester` from the command line in  a new PowerShell terminal the test fails. However, when we place a breakpoint within vscode and run the test then with `F5` or `Invoke-Pester` the test passes.

## Describe your environment
```
Pester version     : 5.3.1 C:\Program Files\WindowsPowerShell\Modules\Pester\5.3.1\Pester.psm1
PowerShell version : 5.1.14393.4583
```

VS Code:
```
Version: 1.61.0 (system setup)
Commit: ee8c7def80afc00dd6e593ef12f37756d8f504ea
Date: 2021-10-07T18:13:09.652Z
Electron: 13.5.1
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.14393
```
## Steps to reproduce
1. Clone the module and its test file from [GitHub](https://github.com/DarkLite1/pester-test-module)
2. In VS Code right click the folder with the script files an choose `Open in Integrated Terminal`
3. Run `Invoke-Pester` and see the test fail:
image
4. Place a breakpoint in the file `Toolbox.PermissionMatrix.Tests.ps1` within the `It` clause and press `F5` to start the tests
5. Notice the test is now successful:
image

Side issue: the prompt is not correct and says `PS>` as visible in the screenshot. Hitting enter after the incorrect prompt does bring us back to the correct one:
image

## Expected Behavior
We expect the test to always pass or always fail, with or without breakpoints set.

## Current Behavior
The test fails when not using a breakpoint for the editor to stop. When there is a breakpoint set, and VS Code stops at the breakpoint, the test is always successful.

fflaten:
## PR Summary
Extends interactive execution to also being triggered from root-level `BeforeDiscovery`.

This is necessary to avoid leaking variables from a `BeforeDiscovery`-block to Run-phase during interactive execution when `BeforeDiscovery` is placed before `Context\Describe` which it usually is. The variables would've been made available in Run-phase from a parent scope.

`Before-/AfterEach` throw when used in root and `Before-/AfterAll` only works when in Discovery-phase, so they're not affected.

Fix #2092

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1635,10 +1635,6 @@ function BeforeDiscovery {
         . $ScriptBlock
     }
     else {
-        if ($invokedInteractively) {
-            return
-        }
-        $invokedInteractively = $true
         Invoke-Interactively -CommandUsed 'BeforeDiscovery' -ScriptName $PSCmdlet.MyInvocation.ScriptName -SessionState $PSCmdlet.SessionState -BoundParameters $PSCmdlet.MyInvocation.BoundParameters
     }
 }
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1631,7 +1631,16 @@ function BeforeDiscovery {
         [ScriptBlock]$ScriptBlock
     )
 
-    . $ScriptBlock
+    if ($ExecutionContext.SessionState.PSVariable.Get('invokedViaInvokePester')) {
+        . $ScriptBlock
+    }
+    else {
+        if ($invokedInteractively) {
+            return
+        }
+        $invokedInteractively = $true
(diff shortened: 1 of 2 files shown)


nohwnd:
Yeah that looks like it was early attempt to solve having multiple top-level blocks in the same file.

nohwnd (approved):

PR #2219, opened 2022-07-16
Fix NRE when assigning null to arrayoption in hashtable

### LINKED ISSUE #2026: New-PesterConfiguration throw an exception when the 'Run.Path' content $null (Object reference not set to an instance of an object)
New-PesterConfiguration throw an exception when the 'Run.Path' content $null

## Describe your environment
Pester version     : 5.2.2 C:\Program Files\WindowsPowerShell\Modules\Pester\5.2.2\Pester.psm1
PowerShell version : 5.1.19041.1023

## Steps to reproduce
```powershell
$PesterOptions=@{
    Run = @{
       Path = $NotExist # typo $PscriptRoot 
    }

    Output = @{
       Verbosity = 'Detailed'
    }
}

$Configuration=New-PesterConfiguration -Hashtable $PesterOptions
```

## Expected Behavior
An error message like "Variable XYZ not exist"  or 'The key 'Run.Path' do not be $null' ?

## Current Behavior
Exception:
Cannot convert value "System.Collections.Hashtable" to type "PesterConfiguration". Error: "Object reference not set to an instance of an object."

Maybe check the $null value and an empty string? 
```Powershell
$PesterOptions=@{
    Run = @{
       Path = "$NotExist"
    }

    Output = @{
       Verbosity = 'Detailed'
    }
}

$Configuration=New-PesterConfiguration -Hashtable $PesterOptions
[string]::Empty -eq $Configuration.run.Path.Value
#true
```

fflaten:
## PR Summary
`GetArrayOrNull()` throws a NRE when value is null, ex. when `Run.Path` was set to `$null` in a hashtable. Added null-check to return before any use of `.GetType()` etc.

Fixing this hashtable-scenario because Pester ignores null-values while casting from hashtable and the error didn't make it clear which option had the problem:
```
InvalidArgument: Cannot convert value "System.Collections.Hashtable" to type "PesterConfiguration". Error: "Object reference not set to an instance of an object."
```

Fix #2026

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/csharp/Pester/DictionaryExtensions.cs
+++ b/src/csharp/Pester/DictionaryExtensions.cs
@@ -58,7 +58,7 @@ namespace Pester
 
         public static T[] GetArrayOrNull<T>(this IDictionary dictionary, string key) where T : class
         {
-            if (!dictionary.Contains(key))
+            if (!dictionary.Contains(key) || dictionary[key] is null)
                 return null;
 
             var value = dictionary[key];
--- a/tst/Pester.RSpec.Configuration.ts.ps1
+++ b/tst/Pester.RSpec.Configuration.ts.ps1
@@ -268,6 +268,27 @@ i -PassThru:$PassThru {
             $config.Run.Path = $config.Run.Path.Default
             $config.Run.Path.IsOriginalValue() | Verify-False
         }
+
+        t "Assigning null to array option using hashtable does not throw" {
+            # https://github.com/pester/Pester/issues/2026
+            $config = [PesterConfiguration]@{ Run = @{ Path = $null } }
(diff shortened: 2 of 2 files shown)


fflaten:
There's a difference between assigning null in a hashtable and directly to the options in a `PesterConfiguration`-object. During casting from hashtable, null-values are ignored in general. Assigning to options in a configuration-object allows null in most cases, except for booloptions which throws an NRE.

- Setters for bool-options don't null-check the values. Intentional? 
- Should we check for null in direct assignments in advanced config in general so both hashtables and direct assignment behave equally? Ex. null-valued paths, scriptblocks etc. are only going to cause issues later on.

PR #2223, opened 2022-07-30
Allow -Mandatory:$false and -HasArgumentCompleter:$false in HaveParameter assertion

### LINKED ISSUE #2105: "Should -HaveParameter 'param' -Not -Mandatory" passes tests if parameter is not present
"Should -HaveParameter 'param' -Not -Mandatory" passes tests if parameter is not present

Would expect it to fail

## Describe your environment
Pester version     : 5.3.1 C:\Users\___\Documents\PowerShell\Modules\pester\5.3.1\Pester.psm1
PowerShell version : 7.2.0

## Steps to reproduce
```powershell
function hi {
    param(
        [parameter(mandatory)]
        [string]$name
    )
    "hi $name"
}

Context 'ctx' {
    It 'test missing param' {
        $fnc = Get-Command 'hi'
        $fnc | Should -HaveParameter 'anotherParam' -Not -Mandatory
    }
}
```

```
Context ctx
  [+] test missing param 6ms (3ms|3ms)
```
-->

## Expected Behavior
would expect test to fail

## Current Behavior
test succeeds

use two tests
```powershell
Context 'ctx' {
    It 'test missing param' {
        $fnc = Get-Command 'hi'
        $fnc | Should -HaveParameter 'anotherParam'
        $fnc | Should -HaveParameter 'anotherParam' -Not -Mandatory
    }
}
```

fflaten:
## PR Summary
Enables `Should -HaveParameter` to test that a parameter actually exists, but was **not** mandatory and/or did **not** have argument completer.

Currently this was only covered by `Should -Not -HaveParameter abc -Mandatory -HasArgumentCompleter` which would also pass if the parameter was missing, requiring a leading `Should -HaveParameter abc` assertion to avoid false positive.

The change does not break existing syntax and `-SwitchParam:$false` syntax is compatible a bool-parameter should it be changed in the future. So I don't believe this commits us to anything in the future that would require us to wait for new Should/Assert.

Fix #2105

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -17,8 +17,13 @@
 
         This test passes, because it expected the parameter URI to exist and to
         be mandatory.
+    .EXAMPLE
+        Get-Command "Invoke-WebRequest" | Should -HaveParameter Method -Mandatory:$false
+
+        This test passes, because it expected the parameter Method to exist and to
+        not be mandatory/required.
     .NOTES
-        The attribute [ArgumentCompleter] was added with PSv5. Previouse this
+        The attribute [ArgumentCompleter] was added with PSv5. Previous this
         assertion will not be able to use the -HasArgumentCompleter parameter
         if the attribute does not exist.
     #>
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -205,15 +205,26 @@
     else {
         $attributes = $ActualValue.Parameters[$ParameterName].Attributes
 
-        if ($Mandatory) {
+        if ($PSBoundParameters.ContainsKey('Mandatory')) {
             $testMandatory = $attributes | & $SafeCommands['Where-Object'] { $_ -is [System.Management.Automation.ParameterAttribute] -and $_.Mandatory }
-            $filters += "which is$(if ($Negate) {" not"}) mandatory"
+            $filters += "which is$(if ($Negate -or $Mandatory -eq $false) {' not'}) mandatory"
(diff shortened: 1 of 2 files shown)


nohwnd:
This is great, but seems inconsistent with the rest of our parameters. I understand it is consistent with PowerShell, but not with Pester syntax. 

I feel like there should be special `Should -HaveParameter abc -NonMandatory -HasNotArgumentCompleter` parameters. (better names?)

fflaten:
I agree with the syntax mismatch.  

We could add `-NotMandatory` etc, with custom logic to throw if used with `-Not` etc. Just felt it was messy due to intellisense/tabcomplete ++ so submitted this as a suggestion.

I'm fine with closing if ex. we'd like to wait for new assert operators where we can split function and parameter sets. 🙂

fflaten:
In case you read the comment above early, I've made some edits. 🙂

fflaten:
> We could use `-NotMandatory` etc with our own conflict-test to throw if used with `-Mandtory` and `-Not`. Might be a little confusing since intellisense/tabcomplete will suggest the conflicts, but at least the parameternames make the conflict obvious. Good enough?

@nohwnd Better? Keep current? Close and fix later?

nohwnd:
Sorry I've dropped the ball on this, do you want to finish it or should I close it?

fflaten:
Depends on how you feel with the syntax. I'm fine with closing this, but the use case is valid and makes sense to support. Can it be handled better through Assert?

nohwnd:
Yes there will be an alternative and I will be happy for your input on it :)

PR #2233, opened 2022-08-08
Improve error when -HaveParameter is used with mock or alias for local function

### LINKED ISSUE #1431: Should -HaveParameter yields a RunTime exception when used on a mocked function

Title sums it up.

To reproduce, use Invoke-Pester on the following tests.ps1 code:

```
    function TestFunction($Parameter1) {
    }

    Mock TestFunction {}

    It 'should have Parameter1' {
        Get-Command TestFunction | Should -HaveParameter "Parameter1"
    }
```

RuntimeException Result:

```
    [-] should have Parameter1 7ms
      RuntimeException: You cannot call a method on a null-valued expression.
      at Should-HaveParameter, C:\Program Files\WindowsPowerShell\Modules\pester\4.9.0\Functions\Assertions\HaveParameter.ps1: line 142
      at <ScriptBlock>, C:\Users\johnd\OneDrive\scripts\StateChangingDemo.tests.ps1: line 
```

Pester version     : 4.9.0 C:\Program Files\WindowsPowerShell\Modules\Pester\4.9.0\Pester.psd1
PowerShell version : 5.1.19041.1
OS version         : Microsoft Windows NT 10.0.19041.0

## 3. Expected Behavior

`Should` should either reject the use of a mock explicitly, or treat the mock as if it was the original function and pass the test.  A RunTime exception shouldn't leak through.

## 5. Fix consideration

The exception occurs in this line in HaveParameter.ps1 (line 142):

`    $hasKey = $ActualValue.Parameters.PSBase.ContainsKey($ParameterName)`

`$ActualValue.Parameters` at this point is `$null`, so the access to PSBase fails.  

The check for whether the function is mocked would need to take place earlier in the function, either to reject the use of Should, or to look up the underlying command and act on that.

## 6. Context

I was trying to write tests for a function.  It wasn't necessary to test parameters on a mocked version of the function, but the nature of the error makes it hard to figure out why it doesn't work.

fflaten:
## PR Summary
PowerShell doesn't resolve local functions properly for `AliasInfo` when function is defined in local scope in different session state. This affects use of `-HaveParameter` on a mocked command, ex. `Get-Command Should -HaveParameter mockedCommand`

Replace generic NRE with exception that suggests a workaround as this problem is unlikely to be fixed soon in PowerShell and probably not at all in Windows PowerShell.

Fix #1431

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -189,7 +189,21 @@
     $buts = @()
     $filters = @()
 
-    $null = $ActualValue.Parameters # necessary for PSv2
+    $null = $ActualValue.Parameters # necessary for PSv2. Keeping just in case
+    if ($null -eq $ActualValue.Parameters -and $ActualValue -is [System.Management.Automation.AliasInfo]) {
+        # PowerShell doesn't resolve alias parameters properly in Get-Command when function is defined in a local scope in a different session state.
+        # https://github.com/pester/Pester/issues/1431 and https://github.com/PowerShell/PowerShell/issues/17629
+        if ($ActualValue.Definition -match '^PesterMock_') {
+            $type = 'mock'
+            $suggestion = "'Get-Command $($ActualValue.Name) | Where-Object Parameters | Should -HaveParameter ...'"
+        } else {
+            $type = 'alias'
+            $suggestion = "using the actual command name, ex. 'Get-Command $($ActualValue.Definition) | Should -HaveParameter ...'"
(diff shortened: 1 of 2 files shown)


nohwnd:
Looks good, I would start from throwing, it is better to make the user correct their test, than us having to support yet another scenario where we "guess" for the user.

nohwnd:
Again pending comments. :/

fflaten:
Pending #2267 to fix CI

PR #2234, opened 2022-08-08
Support mocking and execution in manifest modules

### LINKED ISSUE #933: InModuleScope gives error "Module 'MyModule' is not a Script module." in "Manifest"-type modules

### LINKED ISSUE #1456: Add InModuleScope support for manifest module types

When using a module of type "Manifest", InModuleScope gives error "Module 'MyModule' is not a Script module.". This looks to have been fixed previously in #933 but has been re-introduced.

The workaround is to add an empty psm1 file as the `RootModule` in the manifest. This causes the module to be loaded as a `Script` type instead.

```plaintext
Pester version     : 4.8.1 C:\Users\dalee\Documents\WindowsPowerShell\Modules\Pester\4.8.1\Pester.psd1
PowerShell version : 5.1.18362.145
OS version         : Microsoft Windows NT 10.0.18362.0
```

## 3. Expected Behavior

InModuleScope should work with manifest module types.

## 4.Current Behavior

InModuleScope causes an error with manifest module types.

## 5. Possible Solution

Add support for manifest module types to https://github.com/pester/Pester/blob/master/Functions/InModuleScope.ps1#L72

## 6. Context

Had to use a workaround for manifest module types of adding an empty file as the root module. All other functionality seems to remain consistent.

fflaten:
## PR Summary

Adds support for manifest module to `Mock`, `Should -Invoke` and `InModuleScope`.

Fix #1456 
Fix #933 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -2,10 +2,10 @@
     <#
     .SYNOPSIS
     Allows you to execute parts of a test script within the
-    scope of a PowerShell script module.
+    scope of a PowerShell script or manifest module.
     .DESCRIPTION
     By injecting some test code into the scope of a PowerShell
-    script module, you can use non-exported functions, aliases
+    script or manifest module, you can use non-exported functions, aliases
     and variables inside that module, to perform unit tests on
     its internal implementation.
 
@@ -16,7 +16,7 @@
     injected. This module must already be loaded into the current
     PowerShell session.
     .PARAMETER ScriptBlock
-    The code to be executed within the script module.
+    The code to be executed within the script or manifest module.
     .PARAMETER Parameters
     A optional hashtable of parameters to be passed to the scriptblock.
     Parameters are automatically made available as variables in the scriptblock.
@@ -118,7 +118,7 @@
         $ArgumentList = @()
     )
 
-    $module = Get-ScriptModule -ModuleName $ModuleName -ErrorAction Stop
+    $module = Get-CompatibleModule -ModuleName $ModuleName -ErrorAction Stop
     $sessionState = Set-SessionStateHint -PassThru -Hint "Module - $($module.Name)" -SessionState $module.SessionState
 
     $wrapper = {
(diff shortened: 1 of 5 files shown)


fflaten:
Stumbled into an issue with force-importing manifest module with nested module, https://gist.github.com/fflaten/0df7afe3c2a781e2de29aeb55df0b8ff

This causes alias and mock to be placed in old import while public function will execute in new and as a result not find the mock.

Workaround: Always remove manifest-module before re-importing. Ex.
```powershell
BeforeAll {
    Get-Module ManifestModule -ErrorAction SilentlyContinue | Remove-Module
    Import-Module ManifestModule.psd1`
}

nohwnd (approved):

PR #2239, opened 2022-08-11
Shorten TestDrive directory name

### LINKED ISSUE #2238: Shorten length of temp directory name
### Checklist

As a Pester user on Windows, I sometimes get failures due to Windows' 260 character max path limitation. When Pester creates the temp directory for its test drive, it uses a Guid, which is 36 characters long, e.g. `C:\Users\username\AppData\Local\Temp\19718499-6c05-4f67-bc4c-344f5c47c7bd`. 

Please shorten this to a four character directory name. This is what `[IO.Path]::GetRandomFileName()` uses for uniqueness. This saves 32 characters in the path.

### How should it work?

Instead of using an entire GUID for the temp directory name, use just the first four characters. Change this (in TestDrive.ps1):

```powershell
$Path = [IO.Path]::Combine($tempPath, ([Guid]::NewGuid()))
```

to

```powershell
$Path = [IO.Path]::Combine($tempPath, ([Guid]::NewGuid().ToString().Substring(0, 4)))
```

splatteredbits:
## PR Summary
Pester currently uses a full GUID as its temp directory name, which uses 36 characters. This contributes to test errors when paths used in tests exceed the Windows max path of 260 characters.

The `[IO.Path]::GetTempFileName()` only uses 4 characters for the unique/random part of its filename. Pester should follow the same pattern and only use a temp directory with 4 characters. 

Let's not use a GUID anymore, either. A GUID, when converted to a string, only uses hex characters: 0-9 and a-f, which gives us 65,536 unique possible values. Instead, lets use the first four characters of the file name returned by `[IO.Path]::GetRandomFileName()`. It generates a name that uses all letters of the alphabet and the numbers 0-9, which gives us 36^4 = 1,679,616 possible values.

Fix #2238

## PR Checklist

- [x] PR has meaningful title


src/functions/TestDrive.ps1
```diff
@@ -71,7 +71,8 @@ function Clear-TestDrive ([String[]]$Exclude, [string]$TestDrivePath) {
 function New-RandomTempDirectory {
     do {
         $tempPath = Get-TempDirectory
-        $Path = [IO.Path]::Combine($tempPath, ([Guid]::NewGuid()));
+        $dirName = [IO.Path]::GetRandomFileName().Substring(0, 4)
```
pnstickne, line 74: Perhaps 6 or 8.

pnstickne (approved):

src/functions/TestDrive.ps1
splatteredbits: I chose 4 because that's what the `System.IO.Path.GetTempFileName()` static method uses. It returns a filename with four unique hex characters. But I'm not against six or eight characters, either.

nohwnd: Which implementation of GetTempFileName() are you looking at? .NET does not do it. And .NET Framework does that neither, by looking at the code, they both use cryptographically safe randomization and a file check.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L277-L289

nohwnd: That said, running this on my computer produces 1-10 conflicts. Which we would resolve by file check, so it is not likely to cause too many problems. 

`1..1000000  | % {  [IO.Path]::GetRandomFileName().Substring(0, 4) } | unique | Measure-Object`

splatteredbits: I was looking at `[System.IO.Path]::GetTempFileName()` as the basis for "four is enough":

Here's Windows PowerShell 5.1:

```powershell
$ 1..5 | % { [IO.Path]::GetTempFileName() } | split-Path -leaf
tmp7358.tmp
tmp7359.tmp
tmp735A.tmp
tmp735B.tmp
tmp735C.tmp
```

Notice the four unique hex digits in each filename.

Other versions of PowerShell/.NET behave the same.

PowerShell 7.2.6 on Windows

```powershell
$  1..5 | % { [IO.Path]::GetTempFileName() } | split-Path -leaf
tmp2643.tmp
tmp2644.tmp
tmp2645.tmp
tmp2646.tmp
tmp2647.tmp
```

Looks like PowerShell 7.2.3 on Linux uses random chars:

```powershell
> 1..10 | % { [IO.Path]::GetTempFileName() } | split-Path -leaf
tmpGLw1dM.tmp
tmp9BzMhL.tmp
tmpNTQ1qL.tmp
tmpENSaxM.tmp
tmp2dHGmL.tmp
```

nohwnd:
Added good first issue to this. It should not be complex to replace the way the name is generated with the shortened ID. But what needs to be kept in place is the check that the item does not already exist.

nohwnd:
@fflaten any opinion on shortening the names? 

Maybe we should also unify this with Test Registry?

fflaten:
> @fflaten any opinion on shortening the names? 

Sounds good, but 4 chars sounds a bit low without a Pester-prefix

> Maybe we should also unify this with Test Registry?

Yes

splatteredbits:
> > @fflaten any opinion on shortening the names?
> 
> Sounds good, but 4 chars sounds a bit low without a Pester-prefix

The current test directory name doesn't have a Pester-prefix, it uses an inscrutable Guid. The object is to use as short a name as possible to avoid path too long issues. Adding a prefix would partially defeat that.

nohwnd:
@fflaten  I think we put them in our folder Temp/Pester/<guid> ? And same in registry? HKCU/../Pester/<guid> and we double check that what we are creating does not exist yet, so imho it should be all okay. 

We should just make it the same for registry, if that did not happen yet?

fflaten:
> The current test directory name doesn't have a Pester-prefix, it uses an inscrutable Guid.

Currently if cleanup doesn't run as expected (ex. cancelled script), then those guid-folders will be unidentifiable. "What keeps filling up my temp and is it safe to delete?" By adding a source hint like `pester` somewhere in the path we'd tackle that while also reducing risk of conflict with as few chars as 4. But it's not a critical issue and was just a suggestion as we rarely touch this.

> The object is to use as short a name as possible to avoid path too long issues. Adding a prefix would partially defeat that.

10 vs 36 is still a solid improvement. If the last 6 chars breaks your solution then you're one long username from breaking. So this PR is mostly a quality improvement and not a "every char counts" critical fix imo.

src/functions/TestRegistry.ps1
```diff
@@ -73,7 +73,7 @@ function Get-TestRegistryChildItem ([string]$TestRegistryPath) {
 function New-RandomTempRegistry {
     do {
         $tempPath = Get-TempRegistry
-        $Path = & $SafeCommands['Join-Path'] -Path $tempPath -ChildPath ([Guid]::NewGuid())
+        $Path = & $SafeCommands['Join-Path'] -Path $tempPath -ChildPath ([IO.Path]::GetRandomFileName().Substring(0, 4))
```
fflaten, line 76: nit: requires docs update

fflaten (approved):

src/functions/TestRegistry.ps1
nohwnd: issue https://github.com/pester/Pester/issues/2459

PR #2247, opened 2022-10-22
Add support for multiple aliases (-Alias) to Should -HaveParameter

### LINKED ISSUE #2152: How to use "Should -HaveParameter ... -Alias" when function has multiple aliases?
## Location

- Pester version: 5.3.1
- Function: Should

The documentation for Pester's "Should" command does not make it 100% clear what to do if I am trying to prove all of the possible variant Aliases for a function-under-test have been defined where more than one should exist.

Given a function definition like this:
``` PowerShell
function Get-FakeOutput {
    [CmdletBinding()]
    [OutputType([string])]
    param(
        # Input_A
        [ValidateNotNullOrEmpty()]
        [Alias('In_A', 'A')]
        $Input_A
    )

    begin {}

    process {
        return ("FakeOutput for Input_A: {0}" -f $($Input_A))
    }

    end {}
}
```

The documentation for `Should -HaveParameter` indicates that the `-Alias` parameter expects a `System.Object` but that's it. There are no examples of usage here to guide the user either.

One would hope that you would be able to supply the `-Alias` parameter with an array containing all of the alias variants for the function, like so:

``` PowerShell
Describe "Get-FakeOutput Tests" {
    Context "" {
        It "Has a mandatory parameter called 'Input_A' of type 'System.String' with aliases ('In_a', 'A')" {
            $getFakeOutputCommand = Get-Command -Module "MyModule" -Verb "Get" -Noun "FakeOutput"
            $getFakeOutputCommand | Should -HaveParameter 'Input_A' -Type 'System.String' -Mandatory -Alias $('In_A', 'A')
        }
    }
}
```

However, this results in the error `Expected command Get-FakeOutput to have a parameter Input_A, which is mandatory, of type [System.String] and to have an alias 'In_A A', but it didn't have an alias 'In_A A'.`

The only way I've successfully tested all alias variants on a function parameter was to repeat the Should for each variant like so:

``` PowerShell
Describe "Get-FakeOutput Tests" {
    Context "" {
        It "Has a mandatory parameter called 'Input_A' of type 'System.String' with aliases ('In_a', 'A')" {
            $getFakeOutputCommand = Get-Command -Module "MyModule" -Verb "Get" -Noun "FakeOutput"
            $getFakeOutputCommand | Should -HaveParameter 'Input_A' -Type 'System.String' -Mandatory -Alias 'In_A'
            $getFakeOutputCommand | Should -HaveParameter 'Input_A' -Type 'System.String' -Mandatory -Alias 'A'
        }
    }
}
```

More clear documentation would help others in this situation.

dtewinkel:
## PR Summary

Add support for multiple mandatory aliases for `-HaveParameter`. `-Alias` parameter was changed from a `String` to a `String[]`.

Allows for `Should -HaveParameter Parameter -Alias Alias1, Alias2` to test if parameter `Parameter` has both the aliases `alias1` and `Alias2`. If either one is missing the assert fails.

Fix #2152

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -5,7 +5,7 @@
     [String]$DefaultValue,
     [Switch]$Mandatory,
     [Switch]$HasArgumentCompleter,
-    [String]$Alias,
+    [String[]]$Alias,
     [Switch]$Negate,
     [String]$Because ) {
     <#
@@ -264,14 +264,16 @@
         }
 
         if ($Alias) {
-            $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $Alias
-            $filters += "to$(if ($Negate) {" not"}) have an alias '$Alias'"
+            foreach ($AliasValue in $Alias) {
+                $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $AliasValue
+                $filters += "to$(if ($Negate) {" not"}) have an alias '$AliasValue'"
 
-            if (-not $Negate -and -not $testPresenceOfAlias) {
-                $buts += "it didn't have an alias '$Alias'"
-            }
-            elseif ($Negate -and $testPresenceOfAlias) {
-                $buts += "it had an alias '$Alias'"
+                if (-not $Negate -and -not $testPresenceOfAlias) {
+                    $buts += "it didn't have an alias '$AliasValue'"
+                }
+                elseif ($Negate -and $testPresenceOfAlias) {
+                    $buts += "it had an alias '$AliasValue'"
+                }
             }
         }
     }
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -264,15 +264,26 @@
         }
 
         if ($Alias) {
+
+            $filters += "with$(if ($Negate) {'out'}) alias$(if ($Alias.Count -ge 2) {'es'}) $(Join-And ($Alias -replace '^|$', "'"))"
+            $faultyAliases = @()
             foreach ($AliasValue in $Alias) {
                 $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $AliasValue
-                $filters += "to$(if ($Negate) {" not"}) have an alias '$AliasValue'"
-
                 if (-not $Negate -and -not $testPresenceOfAlias) {
-                    $buts += "it didn't have an alias '$AliasValue'"
+                    $faultyAliases += $AliasValue
                 }
                 elseif ($Negate -and $testPresenceOfAlias) {
-                    $buts += "it had an alias '$AliasValue'"
+                    $faultyAliases += $AliasValue
+                }
+            }
+            if($faultyAliases.Count -ge 1) {
+                $aliases = $(Join-And ($faultyAliases -replace '^|$', "'"))
+                $singular = $faultyAliases.Count -eq 1
+                if($Negate) {
+                    $buts += "it has $(if($singular) {"an alias"} else {"the aliases"} ) $aliases"
+                }
+                else {
+                    $buts += "it didn't have $(if($singular) {"an alias"} else {"the aliases"} ) $aliases"
                 }
             }
         }
--- a/tst/functions/assertions/HaveParameter.Tests.ps1
+++ b/tst/functions/assertions/HaveParameter.Tests.ps1
@@ -63,7 +63,7 @@ InPesterModuleScope {
             function Invoke-DummyFunction {
                 param(
                     [Parameter(Mandatory = $true)]
-                    [Alias('First')]
+                    [Alias('First', 'Another')]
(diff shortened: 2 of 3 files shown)


dtewinkel:
@fflaten, @nohwnd , I see the PR build breaks due to a previously introduced bug in file `tst\functions\Add-ShouldOperator.ts.ps1`, where it runs into the more than 32 assertion operators limitation. This builds fail since a merge to main on the 3rd of October.  
I added a fix, or work-around, to the tests to prevent running into the limitation.

src/functions/assertions/HaveParameter.ps1
```diff
@@ -264,14 +264,16 @@
         }
 
         if ($Alias) {
-            $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $Alias
-            $filters += "to$(if ($Negate) {" not"}) have an alias '$Alias'"
+            foreach ($AliasValue in $Alias) {
+                $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $AliasValue
+                $filters += "to$(if ($Negate) {" not"}) have an alias '$AliasValue'"
```
fflaten, line 269:
```suggestion
            $filters += "with$(if ($Negate) {'out'}) alias$(if ($Alias.Count -ge 2) {'es'}) $(Join-And ($Alias -replace '^|$', "'"))"
            foreach ($AliasValue in $Alias) {
                $testPresenceOfAlias = $ActualValue.Parameters[$ParameterName].Aliases -contains $AliasValue
```

Should we the filter-text outside foreach and to improve readability?
We can utilize helpers in `Format.ps1` like `Join-And` or `Format-Collection` to make it pretty.

**Current PR:** 

> Expected command Invoke-DummyFunction to have a parameter MandatoryParam, to have an alias 'First' and to have an alias 'Second', but it didn't have an alias 'Second'.

**Join-And (suggestion):** 

> Expected command Invoke-DummyFunction to have a parameter MandatoryParam, with aliases 'First' and 'Second', but it didn't have an alias 'Second'.

**Format-Collection:** 

> Expected command Invoke-DummyFunction to have a parameter MandatoryParam, with aliases @('First', 'Second'), but it didn't have an alias 'Second'.

`$filters += "with$(if ($Negate) {'out'}) alias$(if ($Alias.Count -ge 2) {'es'}) $(Format-Collection $Alias)"`

fflaten:
Thanks for submitting the PR!

Also thanks for pointing out the CI-issue! I've submitted a separate PR (#2248) to split the recent help-tests into its own block and to keep this PR clean. Would you mind reverting the workaround? 😇

src/functions/assertions/HaveParameter.ps1
dtewinkel: Thanks for the suggestions. They help to make the output more readable. I applied them, including to the formatting of the wrong aliases.

dtewinkel:
> Thanks for submitting the PR!
> 
> Also thanks for pointing out the CI-issue! I've submitted a separate PR (#2248) to split the recent help-tests into its own block and to keep this PR clean. Would you mind reverting the workaround? 😇

Sure, no problem :-). I'll revert it as soon as it is merged to master. For now, it helps me to show that PS3 and PS4 are failing, and I need to figure out why.

dtewinkel:
> Thanks for submitting the PR!
> 
> Also thanks for pointing out the CI-issue! I've submitted a separate PR (#2248) to split the recent help-tests into its own block and to keep this PR clean. Would you mind reverting the workaround? 😇

Revert is done.

fflaten:
> Revert is done.

Thanks! Closed/reopened to trigger a working CI-run.
LGTM 👏 Waiting on @nohwnd

tst/functions/assertions/HaveParameter.Tests.ps1
```diff
@@ -8,7 +8,7 @@ InPesterModuleScope {
             function Invoke-DummyFunction {
                 param(
                     [Parameter(Mandatory = $true)]
-                    [Alias('First')]
+                    [Alias('First', 'Another')]
```
nohwnd, line 11: My only concern here is that this changes "data" of another test, would it be possible to keep those functions as they are, and introduce a new function that has two aliases.

dtewinkel: All tests share two functions, one with all possible options that need to be validated, and an empty one, so that was the road I continued on. 
I validated that the tests around aliases were still valid for what they were testing. I believe that actually sharing the data in this case 'enhances' the tests without having to add a number of functions to validate all kinds of scenarios. It now also validates that if `-Alias First` is tested, that it succeeds even if there are multiple aliases. 

So, my question is, is your concern a general one and was the setup of the tests perhaps not right in the first place, because they all share the same two functions here, or is there a more specific concern here that you need me to address?

nohwnd: It was a general concern about sharing test data than eventually become so complicated that no-one knows what is going to break. This is not severe here so let's allow it.

nohwnd:
@dtewinkel Thank you :)

PR #2265, opened 2022-11-15
Fix Format-Collection for ICollection

### LINKED ISSUE #2264: Using Format-Collection with ICollection causes infinite loop
A helper-function `Format-Collection` tries to use index-based looping through the `KeyCollection` which doesn't implement `IList` (index-support). It ends up returning a `KeyCollection` on index zero which then end up in the same `Format-Collection` again +++

Easily explained by this sample:
```ps1
(@{ "foo" = $true; "bar" = $false }).Keys[0][0][0][0]
foo
bar
```

`Clone()` is just a shallow clone and apparently returns an array
```ps1
(@{ "foo" = $true; "bar" = $false }).Keys.Clone().Gettype()
IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array

(@{ "foo" = $true; "bar" = $false }).Keys.Clone()[0]
foo
```

_Originally posted by @fflaten in https://github.com/pester/Pester/discussions/2263#discussioncomment-4150239_

fflaten:
## PR Summary
Replace index-based `for`-loop with `foreach` in `Format-Collection` to support ICollections like `KeyCollection` (`@{}.Keys`).

Fix #2264

/cc @Kj3l

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Format.ps1
+++ b/src/Format.ps1
@@ -15,9 +15,13 @@ function Format-Collection ($Value, [switch]$Pretty) {
     $trimmed = $count -gt $Limit
 
     $formattedCollection = @()
-    for ($i = 0; $i -lt [System.Math]::Min($count, $Limit); $i++) {
-        $formattedValue = Format-Nicely -Value $Value[$i] -Pretty:$Pretty
+    # Using foreach to support ICollection
+    $i = 0
+    foreach ($v in $Value) {
+        if ($i -eq $Limit) { break }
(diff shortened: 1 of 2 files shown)


fflaten:
Pending #2267 to fix CI

nohwnd (approved):

PR #2267, opened 2022-11-16
Use encodedCommand in Invoke-InNewProcress for P-tests

fflaten:
## PR Summary
`Invoke-InNewProcress` helper-method in P-tests is broken in PowerShell 7.3 because 7.3 now keeps the backslash previously used to escape quotes in the provided scriptblock.

Replacing with encodedCommand (base64) to avoid version-specific logic.

Fixes broken CI mentioned in https://github.com/pester/Pester/issues/2264#issuecomment-1316554951 . Some runs use prerelease CI-images with PowerShell 7.3 which makes CI inconsistent atm.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/tst/Pester.RSpec.InNewProcess.ts.ps1
+++ b/tst/Pester.RSpec.InNewProcess.ts.ps1
@@ -28,9 +28,10 @@ function Invoke-InNewProcess ([ScriptBlock] $ScriptBlock) {
         . $ScriptBlock
     }.ToString()
 
-    # we need to escape " with \" because otherwise the " are eaten when the process we are starting recieves them
-    $cmd = "& { $command } -PesterPath ""$PesterPath"" -ScriptBlock { $($ScriptBlock -replace '"','\"') }"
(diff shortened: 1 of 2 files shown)


fflaten:
Likely caused by this https://learn.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.3#psnativecommandargumentpassing

@nohwnd Any issues with using `powershell -encodedCommand` for P-tests? Might be flagged by some security teams etc, but this only affects our tests so hopefully okay. Let me know if I should switch to a workaround in the link above.

fflaten:
Ubuntu 20.04 and Windows 2019 tests ran with the new image with 7.3 = all good.

fflaten:
Looks like MS/Github [pinned PS 7.2.7 in newer images](https://github.com/actions/runner-images/pull/6591) atm due to issue like we had, but we should still do something to support future images and local builds using PS7.3+

nohwnd (approved):

PR #2273, opened 2022-11-26
Add -InParameterSet in Should -HaveParameter

### LINKED ISSUE #2227: Evaluate parameter in a parameter set
### Checklist

As a user I would like to validate that a specific parameter set exist and that it contain the expect parameters.

### How should it work?

If `HaveParameter` would learn `CommandParameterSetInfo` we could use that to pass to `Should`.

```powershell
It 'Should have have correct parameters in parameters sets' {
    $parmetersSet = (Get-Command -Name 'New-SqlDscAudit').ParameterSets | Where-Object Name -EQ 'Set1'
    $parameterSet | Should -HaveParameter 'MyParameter'
    $parameterSet | Should -HaveParameter 'Set1Parameter'

    $parmetersSet = (Get-Command -Name 'New-SqlDscAudit').ParameterSets | Where-Object Name -EQ 'Set2'
    $parameterSet | Should -HaveParameter 'MyParameter'
    $parameterSet | Should -HaveParameter 'Set2Parameter'
}
```

Another option is to add `HaveParameterSet` with an optional parameter `WithParameter`.

```powershell
It 'Should have have correct parameters in parameters sets' {
    $commandInfo = Get-Command -Name 'Get-Something')

    $commandInfo | Should -HaveParameterSet 'MySetName1'
    $commandInfo | Should -HaveParameterSet 'MySetName2' -WithParameter 'MyParameter'
    $commandInfo | Should -HaveParameterSet 'MySetName3' -WithParameter @('MyParameter1', 'MyParameter2')
}
```

fflaten:
## PR Summary
Adds support for asserting parameter in a specific parameter set.
This only affects whether parameter exists (or not) and if exists in set and mandatory.

Fix #2227

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/tst/functions/assertions/HaveParameter.Tests.ps1
+++ b/tst/functions/assertions/HaveParameter.Tests.ps1
@@ -7,10 +7,11 @@ InPesterModuleScope {
         if ($PSVersionTable.PSVersion.Major -ge 5) {
             function Invoke-DummyFunction {
                 param(
-                    [Parameter(Mandatory = $true)]
+                    [Parameter(Mandatory = $true, ParameterSetName = 'PrimarySet')]
                     [Alias('First', 'Another')]
                     $MandatoryParam,
 
+                    [Parameter(ParameterSetName = 'PrimarySet')]
                     [ValidateNotNullOrEmpty()]
                     [DateTime]$ParamWithNotNullOrEmptyValidation = (Get-Date),
 
@@ -62,10 +63,11 @@ InPesterModuleScope {
         else {
             function Invoke-DummyFunction {
                 param(
-                    [Parameter(Mandatory = $true)]
+                    [Parameter(Mandatory = $true, ParameterSetName = 'PrimarySet')]
                     [Alias('First', 'Another')]
                     $MandatoryParam,
 
+                    [Parameter(ParameterSetName = 'PrimarySet')]
                     [ValidateNotNullOrEmpty()]
                     [DateTime]$ParamWithNotNullOrEmptyValidation = (Get-Date),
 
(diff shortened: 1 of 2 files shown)


fflaten:
Just to be clear, `Should -Not -HaveParameter abc -InParameterSet 'foo' -Mandatory` will still fail if parameter exists regardless of mandatory-requirement as `-Not`, see #2105

fflaten:
Pending https://github.com/pester/Pester/pull/2267 to fix CI

johlju (approved):
This looks good to me. The tests seems to cover all (that I can see) possible variants.

fflaten:
Not sure about the parameter name. Maybe just `-ParameterSet`?

johlju:
> Not sure about the parameter name. Maybe just `-ParameterSet`?

I think `-InParameterSet` is good enough for now. It makes sense when used together with `-HaveParameter`. It could be `HaveParameterSet` if it it should be able to be used standalone, but if it never should be used alone then as-is is good enough. 🙂 

Come to think of it we did not have a test case for what happens (should happen) when using `-InParameterSet` without `-HaveParameter`, did we?

tst/functions/assertions/HaveParameter.Tests.ps1
```diff
@@ -305,6 +307,38 @@ InPesterModuleScope {
                 $err.Exception.Message | Verify-Equal "Expected command Invoke-DummyFunction to have a parameter ParamWithNotNullOrEmptyValidation, which is mandatory, of type [System.TimeSpan], the default value to be 'wrong value' and has ArgumentCompletion, because of reasons, but it wasn't mandatory, it was of type [System.DateTime], the default value was '(Get-Date)' and has no ArgumentCompletion."
             }
         }
+
+        Context 'Using InParameterSet' {
```
johlju, line 311: We should add a test case when `InParameterSet` is used without `HaveParameter` (plus a test for negate with `Not`). Should it throw for invalid use, or pass when parameter set exist, and fail if parameter set does not exist?

johlju, line 311: We should add a test for `InParameterSet` on a function that does not have any parameter set at all (for example a function that does not have any parameters at all). To validate that it does not throw any exceptions.

fflaten:
Thanks. 

> It could be `HaveParameterSet` if it it should be able to be used standalone

> Come to think of it we did not have a test case for what happens (should happen) when using `-InParameterSet` without `-HaveParameter`, did we?

`-HaveParameter` is the assertion/operator name so it's mandatory (like `-BeTrue` etc).

The reason I'm unsure is that we normally use `-Mandatory` (not `-IsMandatoey`) and `-DefaultValue` (not `-HasDefaulrValue`). But unlike other operators, this one already uses `-HasArgumentCompleter` so already different. 

`-InParameterSet` felt more intuitive since it's used as a single-value filter for the operator. `-ParameterSet` could read like it accepts an array of set names (as an assertion).

tst/functions/assertions/HaveParameter.Tests.ps1
fflaten: `-HaveParameter` is the operator name. It's mandatory so `Should` triggers the right operator-code.

fflaten: See other comments. You have to specify a parameter to use this operator at all. 🙂

johlju:
> -InParameterSet felt more intuitive since it's used as a single-value filter for the operator. -ParameterSet could read like it accepts an array of set names (as an assertion).

I agree.

nohwnd (approved):

Issue #2062, opened 2021-08-21
Stacktrace is not filtered in non-english system languages


sergey-s-betke:
## General summary of the issue

Please, disable verbosity exception output (execution stack trace) for failed tests (https://github.com/pester/vscode-adapter/discussions/51). 

## Describe your environment

Pester version     : 5.3.0 C:\Users\sergei.s.betke\Documents\WindowsPowerShell\Modules\Pester\5.3.0\Pester.psm1
PowerShell version : 5.1.15063.674
OS version         : Microsoft Windows NT 10.0.15063.0

## Steps to reproduce

Test file:

````powershell
#Requires -Version 5.0
#Requires -Modules @{ ModuleName='Pester'; ModuleVersion='5.2.0' }

param(
	# путь к папке с генерируемыми файлами
	[System.String]
	$DestinationPath = ( Join-Path -Path ( ( Get-Location ).Path ) -ChildPath 'output' ),

	# путь к папке с .odt файлами
	[System.String]
	$DestinationDocumentsPath = ( Join-Path -Path $DestinationPath -ChildPath 'doc' ),

	# имя .odt шаблона
	[System.String]
	$DocumentsFilter = '*.odt',

	# пути к .odt файлам
	[System.String[]]
	$DestinationDocFile = @(
		$DestinationDocumentsPath | Where-Object { Test-Path -Path $_ } |
		Get-ChildItem -Filter $DocumentsFilter | Select-Object -ExpandProperty FullName
	),

	# путь к папке с инструментами для сборки
	[System.String]
	$ToolsPath = ( ( Resolve-Path -Path '.\tools' ).Path ),

	# путь к папке с инструментами для документов
	[System.String]
	$DocsToolsPath = ( Join-Path -Path $ToolsPath -ChildPath 'docs' ),

	[System.String]
	$ODFValidatorPath = ( Join-Path -Path $DocsToolsPath -ChildPath 'ODFValidator' ),

	# путь к ODF validator JAR файлу
	[System.String]
	$ODFValidatorJarPath = ( Join-Path -Path $ODFValidatorPath -ChildPath 'ODFValidator.jar' )
)

chcp 65001 > $null;

Describe 'Open Documents' {
	Describe '<Name>' -ForEach @(
		$DestinationDocFile | Get-Item |
		ForEach-Object { @{ Name = $_.Name; FullName = $_.FullName } }
	) {
		It 'is valid (by ODFValidator)' -Tag 'ODFValidator' {
[...]

PR #2276, opened 2022-12-15
Fix issue pester#2062 "Stacktrace is not filtered in non-english system languages"

Nacos:
## PR Summary
Fix issue #2062 "Stacktrace is not filtered in non-english system languages"

When using StackTraceverbosity = 'Filtered', pester internals are filtered from the stack trace using regex matching.
However, Windows stack trace are localized depending on the language of the OS.
The words "At" and "Line" which can be translated (or not) depending on the language of the system.

This PR introduce an heuristic-like solution in ConvertTo-FailureLines function :

We throw an exception on purpose, then catch it in order to detect the localized strings of the stack trace.
We then feed these localized strings back to the regex for filtering pester internals.

## PR Checklist

- [X] PR has meaningful title

### DIFF
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -150,7 +150,8 @@ function Write-PesterHostMessage {
             $message = "$($message -replace '(?m)^', "$fg$bg")$($ANSIcodes.ResetAll)"
 
             & $SafeCommands['Write-Host'] -Object $message -NoNewLine:$NoNewLine
-        } else {
+        }
+        else {
             if ($RenderMode -eq 'Plaintext') {
                 if ($PSBoundParameters.ContainsKey('ForegroundColor')) {
                     $null = $PSBoundParameters.Remove('ForegroundColor')
@@ -534,7 +535,6 @@ function ConvertTo-FailureLines {
                     throw 'Generate exception on purpose'
                 }
                 catch {
-                    $PSItem.ScriptStackTrace
                     $regex = "(?<At>.*)\s(?<ScriptBlockOrFunction>\<\w+\>),\s(?<FileName>\<.+\>)\s*:\s(?<Line>\w+)\s(?<LineNumber>\w+)\z"
                     if ($PSItem.ScriptStackTrace -match $regex) {
                         $LocalizedAt = $Matches["At"]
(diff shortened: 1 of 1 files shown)


Nacos:
FYI, this is a work in progress.

I do not know if this is the right approach to fix this issue.
Would love some feedback about my proposed change.

Also, i think building the module without -Inline will not work because of the code at [lines 540-548 ](https://github.com/Nacos/Pester/blob/66184c488553dbfc0b30649b9062fa3471380238/src/functions/Output.ps1#L540)

fflaten:
Thanks for looking into this!

Last year I intended to simply replace `at|line` with `\S+` as I assume all languages use a single word and the same placements. Unfortunately I had some issues getting multiple samples last year to verify that assumption and include as tests.

While this PR is clever and more precise, it follows the same assumptions. I wonder if it might be unnecessary complex?

_A bonus with your approach is that we could technically localize the added tracelines for PesterAssertionFailed, though I'm not sure if we should since the translation might not be great._

> Also, i think building the module without -Inline will not work because of the code at [lines 540-548](https://github.com/Nacos/Pester/blob/66184c488553dbfc0b30649b9062fa3471380238/src/functions/Output.ps1#L540)

I don't follow. Wouldn't the same word-replacements apply there as well? Could you elaborate?

Nacos:
Hi,
> Last year I intended to simply replace `at|line` with `\S+` as I assume all languages use a single word and the same placements. Unfortunately I had some issues getting multiple samples last year to verify that assumption and include as tests.
> 
> While this PR is clever and more precise, it follows the same assumptions. I wonder if it might be unnecessary complex?
> 
From my experience, localization of technical messages (such as Stack Traces) on Windows can be quite _inconsistent_ depending on the language/culture of the system.
And it's not only applied to words but also to punctuation according to local conventions.

For example, on a French(France) machine, the ":" is preceded by a non-breakable space...
(That's why I used `\s*` when trying to extract the `Filename` with the regex).

I may be able to gather stack traces for a few different languages and share them here if that can be useful to you.

> > Also, i think building the module without -Inline will not work because of the code at [lines 540-548](https://github.com/Nacos/Pester/blob/66184c488553dbfc0b30649b9062fa3471380238/src/functions/Output.ps1#L540)
> 
> I don't follow. Wouldn't the same word-replacements apply there as well? Could you elaborate?

I have pushed a new commit to my branch that modifies the code inside these lines.
However, unlike the code at [lines 556-557](https://github.com/Nacos/Pester/blob/localizedStackTrace/src/functions/Output.ps1#L556),  this `if($true)` "PESTER_BUILD" block does not contains special code related to non-Windows systems.

I currently don't have the means to test on non-Windows system so my concern may be a non-issue though.

fflaten:
👍

> I may be able to gather stack traces for a few different languages and share them here if that can be useful to you.

Would be nice to have as part of tests.

> this `if($true)` "PESTER_BUILD" block does not contains special code related to non-Windows systems.

No need. That part uses the folderpath in the pattern, so will always have the correct slash/backslash. 🙂

nohwnd:
There is another PR #2391 implementing this in hopefully the most reliable way, and we cannot repro the issue. If you know how to make the stack traces show up in different language or how to test this on en-us system that would be awesome contribution to the other PR.

Issue #2286, opened 2023-01-22
-HaveParameter: DefaultValue is always <empty> for binary cmdlets


fflaten:
It fails because `Get-ChildItem` is a binary cmdlet and DefaultValue isn't exposed in any way for those AFAIK (tips are welcome if possible). We should add a "-DefaultValue is not compatible with binary cmdlet" warning and ignore it in those cases.

_Originally posted by @fflaten in https://github.com/pester/Pester/issues/2285#issuecomment-1399497682_

fflaten:
We can only support `ExternalScriptInfo` (ex. `Get-Command ./build.ps1`) , `FunctionInfo` (function) and `ScriptInfo` (can't remember) as they expose `ScriptBlock` where we can access Ast.

PR #2288, opened 2023-01-24
HaveParameter is not dependent on format of param block

### LINKED ISSUE #2285: HaveParameter DefaultValue evaluates to '<empty>' when actual value is not empty
### Checklist

### What is the issue?

Should -HaveParameter -DefaultValue '<anythingOtherThanEmpty>' fails.  Checking for an empty string, when the actual default value is something else, passes instead of failing too.

### Expected Behavior

Should -HaveParameter -DefaultValue should be able to validate default values

### Steps To Reproduce

```powershell
Get-Command -Name Get-ChildItem | Should -HaveParameter -ParameterName 'Path' -DefaultValue '.'
```

### Describe your environment

Pester version     : 5.4.0 C:\Users\brwest\Documents\PowerShell\Modules\Pester\5.4.0\Pester.psm1
PowerShell version : 7.3.1
OS version         : Microsoft Windows NT 10.0.22621.0

### Possible Solution?

Get-ParameterInfo should leverage ASTs since v2 of PowerShell is no longer supported, according to https://pester.dev/docs/introduction/installation#compatibility.  Current implementation requires that param block open parenthesis is on the same line as param keyword for $paramblock to have a count higher than one, which is for the new line, and is still not able to retrieve the default value, unless the parameter name is on the same line as the datatype.

brianwest:
## PR Summary

Get-ParameterInfo wasn't leveraging ASTs because of previous support for PowerShell v2 and required that the opening parenthesis for the param block be on the same line as the keyword.  This change leverages ASTs for Get-ParameterInfo so that HaveParmeter is format agnostic.

Fix #2285

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -31,79 +31,42 @@
     }
 
     function Get-ParameterInfo {
-        param(
-            [Parameter( Mandatory = $true )]
+        param (
+            [Parameter(Mandatory = $true)]
             [Management.Automation.CommandInfo]$Command
         )
         <#
         .SYNOPSIS
-            Use Tokenize to get information about the parameter block of a command
+            Use AST to get information about the parameter block of a command
         .DESCRIPTION
             In order to get information about the parameter block of a command,
             several tools can be used (Get-Command, AST, etc).
             In order to get the default value of a parameter, AST is the easiest
-            way to go; but AST was only introduced with PSv3.
-            This function creates an object with information about parameters
-            using the Tokenize
+            way to go
         .NOTES
-            Author: Chris Dent
+            Author: Brian West
         #>
 
-        function Get-TokenGroup {
-            param(
-                [Parameter( Mandatory = $true )]
-                [System.Management.Automation.PSToken[]]$tokens
-            )
-            $i = $j = 0
-            do {
-                $token = $tokens[$i]
-                if ($token.Type -eq 'GroupStart') {
-                    $j++
-                }
-                if ($token.Type -eq 'GroupEnd') {
-                    $j--
-                }
-                if (-not $token.PSObject.Properties.Item('Depth')) {
-                    $token.PSObject.Properties.Add([Pester.Factory]::CreateNoteProperty("Depth", $j))
-                }
-                $token
-
-                $i++
-            } until ($j -eq 0 -or $i -ge $tokens.Count)
-        }
-
-        $errors = $null
-        $tokens = [System.Management.Automation.PSParser]::Tokenize($Command.Definition, [Ref]$errors)
-
         # Find param block
-        $start = $tokens.IndexOf(($tokens | & $SafeCommands['Where-Object'] { $_.Content -eq 'param' } | & $SafeCommands['Select-Object'] -First 1)) + 1
-        $paramBlock = Get-TokenGroup $tokens[$start..($tokens.Count - 1)]
+        $ast = (Get-Command -Name $Command).ScriptBlock.Ast
 
-        for ($i = 0; $i -lt $paramBlock.Count; $i++) {
-            $token = $paramBlock[$i]
+        $paramBlock = $ast.FindAll(
+            {
+                param($Item)
+                return ($Item -is [System.Management.Automation.Language.ParamBlockAst])
+            },
+            $true
+        )
 
-            if ($token.Depth -eq 1 -and $token.Type -eq 'Variable') {
-                $paramInfo = & $SafeCommands['New-Object'] PSObject -Property @{
-                    Name = $token.Content
-                } | & $SafeCommands['Select-Object'] Name, Type, DefaultValue, DefaultValueType
+        foreach ($parameter in $paramBlock.Parameters) {
+            $paramInfo = & $SafeCommands['New-Object'] PSObject -Property @{
+                Name = $parameter.Name.VariablePath.UserPath
+                DefaultValue = $parameter.DefaultValue.Value
+                DefaultValueType = $parameter.StaticType.Name
+                Type = "[$($parameter.StaticType.Name.ToLower())]"
+            } | & $SafeCommands['Select-Object'] Name, Type, DefaultValue, DefaultValueType
 
-                if ($paramBlock[$i + 1].Content -ne ',') {
-                    $value = $paramBlock[$i + 2]
-                    if ($value.Type -eq 'GroupStart') {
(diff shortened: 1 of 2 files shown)


src/functions/assertions/HaveParameter.ps1
```diff
@@ -235,7 +206,12 @@
 
         if ($PSBoundParameters.Keys -contains "DefaultValue") {
             $parameterMetadata = Get-ParameterInfo $ActualValue | & $SafeCommands['Where-Object'] { $_.Name -eq $ParameterName }
-            $actualDefault = if ($parameterMetadata.DefaultValue) { $parameterMetadata.DefaultValue } else { "" }
+            $actualDefault = if ($parameterMetadata.DefaultValue) {
+                $parameterMetadata.DefaultValue
```
fflaten, line 210: Should we write some type of warning (@nohwnd ?) if `Get-ParameterInfo` doesn't return anything? Like `-DefaultValue is only supported for functions and scripts` and ignore the rest of this block? Only confusing to get a `but the default value was <empty>` when we actually don't know.

src/functions/assertions/HaveParameter.ps1
```diff
@@ -31,79 +31,50 @@
...
-                        $tokenGroup = Get-TokenGroup $paramBlock[($i + 2)..($paramBlock.Count - 1)]
-                        $paramInfo.DefaultValue = [String]::Join('', ($tokenGroup | & $SafeCommands['ForEach-Object'] { $_.Content }))
-                        $paramInfo.DefaultValueType = 'Expression'
-                    }
-                    else {
-                        $paramInfo.DefaultValue = $value.Content
-                        $paramInfo.DefaultValueType = $value.Type
-                    }
+            if ($null -ne $parameter.DefaultValue) {
+                if ([bool](Get-Member -Name Value -InputObject $parameter.DefaultValue -MemberType Property)) {
```
fflaten, line 69: This should also use `& $SafeCommands['Get-Member'] -Name ...`, but we can avoid the call completely with this. 🙂 
```suggestion
                if ($parameter.DefaultValue.psobject.properties['Value']) {
```

fflaten (changes requested):
Thanks for taking the time!
This looks great - and so much simpler! I've added a few suggestions.

Would you mind also adding a test in `tst/functions/assertions/HaveParameter.Tests.ps1`?
Ex the one you provided in the [issue comment](https://github.com/pester/Pester/issues/2285#issuecomment-1399530513)?

src/functions/assertions/HaveParameter.ps1
brianwest: Please, let me know if you're okay with the updates I made to tst\functions\assertions\HaveParameter.Tests.ps1.  Before I added the Ast logic to Get-ParameterInfo, the opening parenthesis for the param block had to be on the same line and the datatype had to be on the same line as the parameter name.  If you could also let me know how you'd like me to update the return value, when $parameterMetadata.DefaultValue is $null, I'd appreciate it.

fflaten: > Please, let me know if you're okay with the updates I made to tst\functions\assertions\HaveParameter.Tests.ps1. Before I added the Ast logic to Get-ParameterInfo, the opening parenthesis for the param block had to be on the same line and the datatype had to be on the same line as the parameter name.

Please revert and add a specific `It`-test for this param-scenario with a minimal function included in the test. The current change would likely be lost in a future format-commit. Ex.

```ps1
    It 'works without paramblock with parantheses on different line' {
        function adv {
            param
            (
                $param1,
                $param2 = '123'
            )
        }

        Get-Command adv | Should -HaveParameter 'param2' -DefaultValue '123'
    }
```

> If you could also let me know how you'd like me to update the return value, when $parameterMetadata.DefaultValue is $null, I'd appreciate it.

If `$parameterMetadata` exists but `DefaultValue` is null then the parameter exists, but does not have a default value (or it's null). In that scenario it should behave like now. 

I was thinking about a warning when `Get-ParameterInfo $ActualValue` doesn't return anything, prior to filtering by parameter-name. 🙂 I suggest we fix that later in a different PR when this is merged. Issue is tracked in #2286.

src/functions/assertions/HaveParameter.ps1
```diff
@@ -31,77 +31,50 @@
...
         # Find param block
-        $start = $tokens.IndexOf(($tokens | & $SafeCommands['Where-Object'] { $_.Content -eq 'param' } | & $SafeCommands['Select-Object'] -First 1)) + 1
-        $paramBlock = Get-TokenGroup $tokens[$start..($tokens.Count - 1)]
-
-        for ($i = 0; $i -lt $paramBlock.Count; $i++) {
-            $token = $paramBlock[$i]
+        $ast = $Command.ScriptBlock.Ast
+
+        if ($null -ne $ast) {
+            $paramBlock = $ast.FindAll(
```
fflaten, line 54: For improved readability, we can return early to avoid unnecessary nesting of the remaining code.

```ps1
        $ast = $Command.ScriptBlock.Ast
        # Unsupported command-type. Unable to determine default value, so returning early
        if ($null -eq $ast) { return }

        $paramBlock = $ast.FindAll(
```

fflaten (changes requested):
Sorry for the late response. See updated comments 🙂

src/functions/assertions/HaveParameter.ps1
brianwest: I decided to return if not able to find the parameters.  I hope that's okay.

src/functions/assertions/HaveParameter.ps1
```diff
@@ -31,77 +31,50 @@
...
+                {
+                    param($Item)
+                    return ($Item -is [System.Management.Automation.Language.ParamBlockAst])
+                },
+                $true
+            )
 
-            if ($token.Depth -eq 1 -and $token.Type -eq 'Variable') {
+            foreach ($parameter in $paramBlock.Parameters) {
```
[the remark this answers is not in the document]
brianwest: Good catch!  Sorry about that.  I've updated the logic to handle override parameters, as well as nested functions.

src/functions/assertions/HaveParameter.ps1
brianwest: I've added a dedicated test for a paramblock with the opening parenthesis on a new line.

fflaten (approved):
LGTM!

nohwnd (approved):

PR #2303, opened 2023-02-11
Ignore PassThru when run fail while using Run.Exit

### LINKED ISSUE #2300: Using Run.Exit and Run.PassThru causes unexpected output to stdout
### Checklist

### What is the issue?

While using the Run.PassThru and Run.Exit configuration options, a Pester object is unexpectedly dropped to the output stream if an exception is raised. The object is not in the output stream if an exception is not raised.

See screenshot below:

image

- `test.ps1` contains some blocks with a single test. `throw` is used in `BeforeAll` but it can be used in any block to reproduce the issue
- The test script file is called from `testrun.ps1` and the output should be assigned to `$r` within the scope of the script file
- `$r` isn't dropped to the output stream, however we can see a Pester output object in the console which called `testun.ps1`

In order to reproduce, both PassThru and Exit must be true.

## test.ps1

```powershell
Describe 'Some test' {

    BeforeAll {
        Write-Host 'hello world from beforeall'
        throw 'some error'
    }

    AfterAll { 
        Write-Host 'hello world from afterall'
    }

    it 'should be true' {
        throw 'some error'
        Write-Host 'within a test'
        $true | Should -Be $true
    }
}
```

## testrun.ps1
```powershell
$Container = New-PesterContainer -Path 'C:\path\to\test.ps1'

$Configuration = New-PesterConfiguration -Hashtable @{
    Run = @{
        PassThru = $true
        Exit = $true
        Container = $Container
        SkipRemainingOnFailure = 'Container'
    }
    Output = @{
        Verbosity = 'None'
    }
    TestResult = @{
        Enabled = $false
        TestSuitName = 'abc'
    }
}

$r = Invoke-Pester -Configuration $Configuration
```

### Expected Behavior

I expected to not see any output in the given code example.

### Steps To Reproduce

1) Create `test.ps1` with the above code
2) Create `testrun.ps1` with the above code and update line 1 for path to `test.ps1`
3) Call `testrun.ps1`

### Describe your environment

```
Pester version     : 5.3.1 C:\Users\adamp\OneDrive - Adam Cook\Documents\PowerShell\Modules\Pester\5.3.1\Pester.psm1
PowerShell version : 7.3.2
OS version         : Microsoft Windows NT 10.0.22621.0
```

### Possible Solution?

@fflaten suggests is [discord](https://discord.com/channels/180528040881815552/447476910499299358/1073885781854261278) to:

> We should likely ignore passthru when run.exit is set and there is a failed test. It's returned as text so it wouldn't provide anything more than the normal colored output

fflaten:
## PR Summary
Result object was printed when run failed because it was sent to standard output, but never reached variable assignment before `Run.Exit` killed the process causing a text dump.

Fix #2300 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -1226,7 +1226,8 @@ function Invoke-Pester {
                 Remove-RSPecNonPublicProperties $run
             }
 
-            if ($PesterPreference.Run.PassThru.Value) {
+            $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
+            if ($PesterPreference.Run.PassThru.Value -and -not ($PesterPreference.Run.Exit.Value -and 0 -ne $failedCount)) {
                 $run
             }
 
@@ -1253,12 +1254,10 @@ function Invoke-Pester {
         # go back to original CWD
         if ($null -ne $initialPWD) { & $SafeCommands['Set-Location'] -Path $initialPWD }
 
-        # exit with exit code if we fail and even if we succeed, otherwise we could inherit
-        # exit code of some other app end exit with it's exit code instead with ours
-        $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
         # always set exit code. This both to:
-        # - prevent previous commands failing with non-zero exit code from failing the run
+        # - avoid inheriting a previous commands non-zero exit code
         # - setting the exit code when there were some failed tests, blocks, or containers
+        $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
         $global:LASTEXITCODE = $failedCount
 
         if ($PesterPreference.Run.Throw.Value -and 0 -ne $failedCount) {


src/Main.ps1
```diff
@@ -1253,12 +1254,10 @@ function Invoke-Pester {
...
         if ($null -ne $initialPWD) { & $SafeCommands['Set-Location'] -Path $initialPWD }
 
-        # exit with exit code if we fail and even if we succeed, otherwise we could inherit
-        # exit code of some other app end exit with it's exit code instead with ours
-        $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
         # always set exit code. This both to:
-        # - prevent previous commands failing with non-zero exit code from failing the run
+        # - avoid inheriting a previous commands non-zero exit code
         # - setting the exit code when there were some failed tests, blocks, or containers
+        $failedCount = $run.FailedCount + $run.FailedBlocksCount + $run.FailedContainersCount
```
fflaten, line 1260: Assigning this twice in case we didn't reach the previous one inside try/catch

fflaten:
Actually, this will break `$res = & ./test.ps1` because `exit` will exit the script. Any way to identify if Invoke-Pester was called inside a script?

If not, not worth it? The returned text is mostly a problem with the PowerShell  extension-terminal (PSIC) in VSCode because it disables `exit`.

nohwnd:
> But is this intended to work or just luck? I'd lean towards merging this, because Run.Exit says Exit with non-zero exit code when the test run fails. so I wouldn't expect to be able to post-process anything.

Exit in powershell is defined as exiting the script with the error code, not the engine. So $res = & ./test.ps1 should imho assign the value, rather than not. 

Or am I misinterpreting what you are saying here?

fflaten:
> Or am I misinterpreting what you are saying here?

I think we agree. There's two sides here and I'm not sure if fixing one is worth breaking the other, which is why I left this without ready to merge. The question is if anyone's gonna be missing the part being broken. Current behavior:

**Catching output from script**
Running `$res = & ./test.ps1` you'd keep the pipeline and result-object even on failure. But would anyone process it for a failed run while using `-CI/Run.Exit`? Or would they immediately check `$? -eq $false` and break? Personally I'd use `$res = Invoke-Pester` inside the `test.ps1` script (and that script would exit), not catch the script output itself. I can't say how others use Pester.

**Calling Pester directly**
Users invoking Pester directly would get their process terminated by `exit` and the result-object would be printed as stirng (noise) in the CI-logs etc. Unless they do called Pester through a child-process like `$a = pwsh -noprofile -command { Invoke-Pester ...`, which seems unlikely.

nohwnd:
Close or keep? I agree with your reasoning above that neither case is preferable, so maybe just keep the current one and close this?

fflaten:
I'd say merge in v6. Easy rollback if necessary.

I'm struggling to see when someone would combine Run.Exit with processing script output. If so they could just drop Run.Exit and add a check in their script to get the previous behavior.

Issue #1234, opened 2019-02-02
Problem using HaveCount on [hashtable]


johlju:
## 1. General summary of the issue

When using `HaveCount` on a hashtable, the count is always 1. The following snippet shows the problem (both in Windows PowerShell and PowerShell Core).

```powershell
Describe 'Get-HashTable' {
    Context 'When returning a hashtable' {
        It 'Test' {
            $getHashTableResult = @{
                Property1 = '1'
                Property2 = '3'
            }

            $getHashTableResult | Should -BeOfType [System.Collections.Hashtable]
            $getHashTableResult | Should -HaveCount 2
        }
    }
}

```

## 2. Describe Your Environment

```
Pester version     : 5.1.1 C:\source\SqlServerDsc\output\RequiredModules\Pester\5.1.1\Pester.psm1
PowerShell version : 7.1.0
OS version         : Microsoft Windows NT 10.0.19042.0
```

## 3. Expected Behavior

`HaveCount` should report the correct number of properties in the hashtable.

## 4.Current Behavior

Fails with an error even though there are two properties in the hashtable.

```
Starting discovery in 1 files.
Discovery finished in 22ms.
[-] Get-HashTable.When returning a hashtable.Test 8ms (7ms|1ms)
 Expected a collection with size 2, but got collection with size 1 @(System.Collections.Hashtable).
 at $getHashTableResult | Should -HaveCount 2, C:\Users\johan.ljunggren\Desktop\Bug_HaveCount.ps1:10
```

## 5. Possible Solution

>NOTE: The below screenshots are from Pester 4.x, but the same issue still exist in 5.1.1.

The problem seems that it compares the wrong array. Instead of checking `$ActualValue.Count`, it should check `$ActualValue[0].Count` (at least in this case).

![image](https://user-images.githubusercontent.com/7189721/52165165-64215c00-26fd-11e9-8d2e-6b33244b34c0.png)

```
 [DBG]> $ActualValue.Count
1
 [DBG]> $ActualValue[0].Count
2
```
![image](https://user-images.githubusercontent.com/7189721/52165203-e578ee80-26fd-11e9-85ca-391d0810278c.png)

## 6. Context

Very minor inconvenience. Having `HaveCount` would just make the code prettier. Workaround is using the following instead.

```
$getHashTableResult.Count | Should -Be 2
```

johlju:
I have updated the issue for the Pester 5.1.1 where the issue still exist. It is also not possible to use `ActualValue`.

[...]

PR #2315, opened 2023-03-01
Avoid nested array when using Should -ActualValue

### LINKED ISSUE #2314: Should does NOT assert correctly for certain calls when calling it in different ways
### Checklist

### Preface

PowerShell has the ability to preload Cmdlet parameters in Hashtables used to Splat the eventual Cmdlet Call

Pester is made up of Cmdlet so this tenant should hold true

So for instance 

`'10 | Should -Be 10`

Could be splatted like

```
$Should =
    @{
        ActualValue =
            10
        ExpectedValue =
            10
        Be =
            $true
    }
Should @Should
```

And this example works just fine

However something like in the steps to reproduce does not

### Expected Behavior

Yes Splatting might not be the traditional way to call Pester but a lot of places have coding style guides that heavily prioritize splatting and Pester's style flies in the face of that

I'm not here to argue the merrits of one coding style over another simply that

Regardless of the means to call the Cmdlet the result should be the same

The fact the error states it expected X and got X is a bit humorous to me

### Steps To Reproduce

```
$Should =
	@{
		ActualValue =
			@(1,2,3)
		ExpectedValue =
			@(1,2,3)
		Be =
			$true
	}
```

Works with 

`$Should.ActualValue | Should -Be $Should.ExpectedValue`

but NOT

`Should -ActualValue $Should.ActualValue -Be -ExpectedValue $Should.ExpectedValue`

or

Should @Should

The last two both happily report

```
Expected @(1, 2, 3), but got @(1, 2, 3).
At C:\Program Files (x86)\WindowsPowerShell\Modules\Pester\5.4.0\Pester.psm1:8130 char:13
+             throw $errorRecord
+             ~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidResult: (System.Collecti...,System.Object]:Dictionary`2) [], Exceptio
   n
```

### Describe your environment

Windows 10. 
Windows PowerShell, and PowerShell

### Possible Solution?

For the above example with any complicated object the solution is to have a Compare-Object call as the ActualValue and BeNullOrEmpty as the assertion so

```
$CompareObject =
	@{
		ReferenceObject =
			@(1,2,3)
		DifferenceObject =
			@(1,2,3)
	}
$Should =
	@{
		ActualValue =
			Compare-Object @CompareObject
		BeNullOrEmpty =
			$true
	}
Should @Should
```

Will work and you can add Property to the hashtable for complicated object comparisions

fflaten:
## PR Summary
**Possible breaking change!**

If an enumerable was provided with `Should -ActualValue ..` instead of pipeline, unroll to be consistent with pipeline-input.

Fix #2314
Might also solve this: https://github.com/pester/Pester/issues/1234#issuecomment-748627991

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -394,3 +394,10 @@ function Fold-Run {
         }
     }
 }
+
+function IsPSEnumerable($Object) {
+    # https://github.com/pester/Pester/issues/1200#issuecomment-493043683
+    # PowerShell doesn't consider all IEnumerable an enumerable, ex. string is excluded.
+    $enumerator = [System.Management.Automation.LanguagePrimitives]::GetEnumerator($Object)
+    return $null -ne $enumerator
+}
--- a/src/functions/assertions/Should.ps1
+++ b/src/functions/assertions/Should.ps1
@@ -120,7 +120,14 @@ function Should {
     }
 
     process {
-        $inputArray.Add($ActualValue)
+        # If array was provided using -ActualValue @(1,2,3), unroll like pipeline for consistent behaviour
+        if (-not $PSCmdlet.MyInvocation.ExpectingInput -and (IsPSEnumerable -Object $ActualValue)) {
+            foreach ($object in $ActualValue) {
+                $inputArray.Add($object)
+            }
+        } else {
+            $inputArray.Add($ActualValue)
+        }
     }
 
     end {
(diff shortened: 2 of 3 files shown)


fflaten:
This could be considered a breaking change for people currently using workarounds like suggested [here](https://github.com/pester/Pester/issues/2314#issuecomment-1450756690)

nohwnd (approved):

Issue #2304, opened 2023-02-13
RuntimeExceptions for invalid output configuration doesn't kill the run


fflaten:
### Checklist

- [X] Issue has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

When an invalid value is set in output-related configuration options like `Output.CIFormat`, an exception is called once per failed test/block.

![image](https://user-images.githubusercontent.com/3436158/218558709-76e73739-d07e-432d-b0a1-6604367db393.png)

### Expected Behavior

Even though the current behavior would allow a successful run to execute, we should move this validation earlier and kill the run as all errors will be unreadable anyways.

### Describe your environment

Pester version     : 5.4.0 /workspaces/Pester/bin/Pester.psm1                                                           
PowerShell version : 7.2.6
OS version         : Unix 5.15.79.1

### Possible Solution?

Move validation for output-options to `Start`-step?

PR #2317, opened 2023-03-04
WIP - Refactoring Invoke-Pester and early configuration-validation

fflaten:
## PR Summary
Playing around with multiple refactoring of `Invoke-Pester` to make it more readable + improving configuration validation. Ex.
- [ ] Extract parameter set migration for Simple and Legacy-set
- [ ] Call TestResult-export through plugin End-step
- [ ] Move CodeCoverage processing to plugin-steps Start, RunEnd and End
- [ ] Remove legacy comment help? Ex. hashtable-format for -Script, CodeCoverage etc.
- [ ] Validate configuration early to avoid running a long run just to loose it at the end when it throws due to invalid format for coverage, testresult etc. (ex. #2304)
  - Throwing in Start plugin-step creates an ugly and duplicate error atm. Likely due to being thrown through `Assert-Success`. Might be better to use own function like `Resolve-OutputConfiguration`

## PR Checklist

- [ ] PR has meaningful title


fflaten:
>   * Throwing in Start plugin-step creates an ugly and duplicate error atm. Likely due to being thrown through `Assert-Success`. Might be better to use own function like `Resolve-OutputConfiguration`

The duplicate was because `Assert-Success` printed the error while also throwing it. Suggest removing the host message.

Thoughts on where to validate config, especially string-options? Suggestions:
1. Start plugin-step. Removed the duplicate message, but still a bit verbose due to `Assert-Success`
image

2. Validation-function, ex. `Resolve-OutputConfiguration`
image

3. Validate in assembly for StringOption. Would not support public plugins integrating, but those plugins should probably not replace built-in options.
image

5. A new `ConfigurePlugin/SetupPlugin` step invoked without `Invoke-PluginStep` or a different branch to simplify error. Maybe invoked before `Find-File` call which can be slow. This might fit better with public plugins.

src/Pester.Runtime.ps1
```diff
@@ -1289,21 +1300,20 @@ function Assert-Success {
...
     $err = ""
     foreach ($r in $InvocationResult) {
+        $rc++
         $ec = 0
         if ($null -ne $r.ErrorRecord -and $r.ErrorRecord.Length -gt 0) {
-            $err += "Result $($rc++):"
             $anyFailed = $true
             foreach ($e in $r.ErrorRecord) {
-                $err += "Error $($ec++):"
+                $err += "$([Environment]::NewLine)Result $rc - Error $((++$ec)):"
```
fflaten, line 1308: Previous message didn't print `$ec` and `$rc` (only incremented them) due to missing parantheses.

src/Pester.Runtime.ps1
```diff
@@ -1216,7 +1227,7 @@ function Invoke-PluginStep {
     # this is end step, we should run all steps no matter if some failed, and we should run them in opposite direction
     # only do this if there is more than 1, to avoid the "expensive" -like check and reverse
     $isEndStep = 1 -lt $pluginsWithGivenStep.Count -and $Step -like "*End"
-    if (-not $isEndStep) {
+    if ($isEndStep) {
```
nohwnd, line 1230: So this did not reverse the steps before, did you see any bugs linked to that?

fflaten: No bugs I'm aware of. Just not behaving like comment says. We reverse on TearDown-steps as expected, but this currently reverses on everything else than *End.

Not a critical part of the PR, just testing it out while in draft.

fflaten:
> > * Throwing in Start plugin-step creates an ugly and duplicate error atm. Likely due to being thrown through `Assert-Success`. Might be better to use own function like `Resolve-OutputConfiguration`
> 
> The duplicate was because `Assert-Success` printed the error while also throwing it. Suggest removing the host message.
> 
> Thoughts on where to validate config, especially string-options? Options:
> 
> 1. Start plugin-step. Removed the duplicate message, but still a bit verbose due to `Assert-Success`
>    image
> 2. Validation-function, ex. `Resolve-OutputConfiguration`
>    image
> 3. Validate in assembly for StringOption. Would block third-party plugins from integrating with built-in options like `TestResult.OutputFormat`, but do we want to support that?
>    image
>    image
> 4. A new `ConfigurePlugin/SetupPlugin` step invoked without `Invoke-PluginStep` or a different branch to simplify error. Maybe invoked before `Find-File` call which can be slow. This might also benefit public plugin API.

Thoughts on this @nohwnd ? Use validation-functions for now and consider a plugin-step later as part of making plugins public?

fflaten:
Rebased and updated with validation-functions for now. Can revisit later when looking into public plugins API.

Also cleaned up the error message for invalid StringOption. It now list option and supported values just like CodeCoverage did.
image
image

src/Pester.Runtime.ps1
```diff
@@ -1292,21 +1303,20 @@ function Assert-Success {
...
+                $err += "$([Environment]::NewLine)Result $rc - Error $((++$ec)):"
                 $err += & $SafeCommands["Out-String"] -InputObject $e
                 $err += & $SafeCommands["Out-String"] -InputObject $e.ScriptStackTrace
             }
         }
     }
 
     if ($anyFailed) {
-        $Message = $Message + ":`n$err"
-        Write-PesterHostMessage -ForegroundColor Red $Message
```
fflaten, line 1309: Revert this or good? Not validating configuration in plugin-step now, so PR is not affected by the duplicate error from `Assert-Success`. Not sure if it's useful somewhere else

nohwnd (approved):

src/Pester.Runtime.ps1
nohwnd: good.

nohwnd:
@fflaten  Approved both this and your test output refactoring PRs, they seem like they might run into each other so I was not sure in which order you want them merged, please merge any way you like :)

fflaten:
@nohwnd Thanks! Let's start with this. It will also cause a conflict in the last PR for container data. Can fix both later today. 🙂

fflaten:
Btw. Don't have merge rights, so up to you :)

nohwnd:
@fflaten  you do now.

Issue #2138, opened 2022-02-09
`Should` throws exception for legacy syntax when parameters are broken across lines


snow-jlove:
## General summary of the issue

When `Should` parameters (with dashes!) are split across lines, a runtime exception gets thrown, stating "RuntimeException: Legacy Should syntax (without dashes) is not supported in Pester 5. Please refer to migration guide at: https://pester.dev/docs/migrations/v3-to-v4".

## Describe your environment

```
Pester version     : 5.3.1 C:\Users\jlove\Documents\WindowsPowerShell\Modules\Pester\5.3.1\Pester.psm1
PowerShell version : 5.1.19041.1320
OS version         : Microsoft Windows NT 10.0.19044.0
```

## Steps to reproduce

```powershell
Describe ShouldLegacySyntax {
    It "does not complain about syntax" {
        1 + 1 | Should -Be 2
    }

    It "complains about syntax" {
        1 + 1 | Should `
            -Be 2
    }
}
```

## Expected Behavior

Both of the above tests should be treated equally and pass.

## Current Behavior

The second above test fails with the syntax exception.

snow-jlove:
I ran into this issue when I tried splitting a long command across lines. Consider this line - when indented, it overran my editor's vertical ruler (120 chars):

```powershell
        Should -CommandName Install-Module -Exactly -Invoke -ModuleName $ModuleName -ParameterFilter $ParameterFilter1 -Times 2

# gets formatted to...

        Should `
            -CommandName Install-Module `
            -Exactly `
            -Invoke `
            -ModuleName $ModuleName `
            -ParameterFilter $ParameterFilter1 `
            -Times 2
```

nohwnd:
Yikes, that rule is nothing but trouble. :/

fflaten:
Not parameters in general, just the operator (Invoke\Be\...). I'd recommend using the operator as your first parameter for readability and as a bonus that would let you keep it on the first line and avoid this issue. Try:

```powershell
Should -Invoke `
    -CommandName Install-Module `
    -Exactly `
    -ModuleName $ModuleName `
    -ParameterFilter $ParameterFilter1 `
    -Times 2
```

nohwnd:
Upvote on what Frode said, that makes it way easier to understand what that `Should` does and avoids the issue as well.

ElvenSpellmaker:
@nohwnd That seems like a bad workaround, it's much cleaner to only have a single parameter per line and either all on newlines or none.

[...]

PR #2325, opened 2023-03-23
Don't throw legacy error when using Should operator after escaped linebreak

fflaten:
## PR Summary
Fix false positive exception when specifying Should-operator after linebreak using backtick/escape char, ex.

```ps1
1 | Should `
    -Be 1
```

Provided as an [alternative to removing the legacy-check](https://github.com/pester/Pester/issues/2138#issuecomment-1183070814) though I'm not against removing it.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assertions/Should.ps1
+++ b/src/functions/assertions/Should.ps1
@@ -106,7 +106,7 @@ function Should {
         }
 
         # A bit of Regex lets us know if the line used the old form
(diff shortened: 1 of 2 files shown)


nohwnd:
Added note to remove the whole syntax in 6.0.0 here https://github.com/pester/Pester/issues/2437 

Do you still want to fix this, or close?

PR #2329, opened 2023-03-28
Quote input string in BeNullOrEmpty error

fflaten:
## PR Summary
Fix missing quotes when single string is passed to `Should -BeNullOrEmpty`.

/cc @johlju 

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assertions/BeNullOrEmpty.ps1
+++ b/src/functions/assertions/BeNullOrEmpty.ps1
@@ -1,5 +1,5 @@
 ﻿
-function Should-BeNullOrEmpty([object[]] $ActualValue, [switch] $Negate, [string] $Because) {
+function Should-BeNullOrEmpty($ActualValue, [switch] $Negate, [string] $Because) {
     <#
     .SYNOPSIS
     Checks values for null or empty (strings).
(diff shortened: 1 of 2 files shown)


src/functions/assertions/BeNullOrEmpty.ps1
```diff
@@ -52,7 +53,8 @@ function Should-BeNullOrEmpty([object[]] $ActualValue, [switch] $Negate, [string
             $failureMessage = NotShouldBeNullOrEmptyFailureMessage -Because $Because
         }
         else {
-            $failureMessage = ShouldBeNullOrEmptyFailureMessage -ActualValue $ActualValue -Because $Because
+            $valueToFormat = if ($singleValue) { $expandedValue } else { $ActualValue }
```
johlju, line 56: This seems to affect non empty hashtables too. Shall we add a unit tests to verify the exception message when the hast table is not empty? So it too returns the expected message? There is an existing test at line 22-24 but it does not verify the exception message.

fflaten: We don't do any special formatting for hashtables. It only drops the outer array, which in a single-item case should be the array Pester wrapped pipeline-input in.

```ps1
@{ hello = "world" } | Should -BeNullOrEmpty
# Pester 5.3.3
Expected $null or empty, but got @(System.Collections.Hashtable).

# After PR
Expected $null or empty, but got System.Collections.Hashtable

johlju: Looks good.

nohwnd (approved):

PR #2332, opened 2023-03-30
Make initial Mock cleanup more performant

nohwnd:
## PR Summary

There is initial cleanup of orphaned mocks that happens in Invoke-Pester and makes sure that there are no Mock aliases in place coming from previously cancelled runs. In Pester v4 we only cleaned up the main session state, but in v5 we pick up every loaded module and clean up there. This adds up to the point where it can take several seconds (in docker container) to do this cleanup. 

This PR optimizes the cleanup to just query functions, rather than all cmdlets and filtering them.

## PR Checklist

- [x] PR has meaningful title
- [ ] Tests are added/update *(if required)* (there are no tests for this functionality in the first place imho)

### DIFF
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -1573,32 +1573,25 @@ function Test-IsClosure {
 }
 
 function Remove-MockFunctionsAndAliases ($SessionState) {
-    # when a test is terminated (e.g. by stopping at a breakpoint and then stoping the execution of the script)
-    # the aliases and bootstrap functions for the currently mocked functions will remain in place
+    # When a test is terminated (e.g. by stopping at a breakpoint and then stoping the execution of the script)
+    # the aliases and bootstrap functions for the currently mocked functions will remain in place.
     # Then on subsequent runs the bootstrap function will be picked up instead of the real command,
     # because there is still an alias associated with it, and the test will fail.
     # So before putting Pester state in place we should make sure that all Pester mocks are gone
-    # by deleting every alias pointing to a function that starts with PesterMock_. Then we also delete the
-    # bootstrap function.
+    # by deleting every alias pointing to a function that starts with PesterMock_.
+    # As a perf optimization we don't delete the bootstrap functions because running Get-Command
+    # in every loaded module has significant overhead (https://github.com/pester/Pester/discussions/2331)
+    # and the PesterMock_ function won't be picked up without an alias.
     $Get_Alias = $script:SafeCommands['Get-Alias']
-    $Get_Command = $script:SafeCommands['Get-Command']
     $Remove_Item = $script:SafeCommands['Remove-Item']
     foreach ($alias in (& $Get_Alias -Definition "PesterMock_*")) {
         & $Remove_Item "alias:/$($alias.Name)"
     }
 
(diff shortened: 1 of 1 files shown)


src/functions/Mock.ps1
```diff
@@ -1823,13 +1816,13 @@ function Repair-EnumParameters {
     # https://github.com/PowerShell/PowerShell/issues/17546
     $ast = [System.Management.Automation.Language.Parser]::ParseInput("param($ParamBlock)", [ref]$null, [ref]$null)
     $brokenValidateRange = $ast.FindAll({
```
nohwnd, line 1818: Autoformatter did this. It kinda makes sense.

nohwnd:
It is tested: [n] ERROR: - Invoke-Pester cleans up orphaned mock hooks and aliases in modules ->  

will need to fix.

fflaten (commented):
Agree with the idea, but are we able to mute these? Could cause errors in a follow-up run.

```ps1
> & (Get-Module PSReadLine) { dir function:\ }                                                                                                   
Exception: /workspaces/Pester/src/functions/Pester.SessionState.Mock.ps1:592:9
Line |
 592 |          throw "Mock data are not setup for this scope, what happened? …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Mock data are not setup for this scope, what happened?
```

Not sure which property being enumerated is causing this.

```ps1
> & (Get-Module PSReadLine) { Get-Command Pester* } | % parameters
InvalidOperation: 
Line |
  34 |  … icparam { & $MyInvocation.MyCommand.Mock.Get_MockDynamicParameter -Cm …
     |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.
```

Maybe silenced by checking for `Mock`-property before invoking it? Even better, are we able to self-delete based on similar condition there and/or begin?

nohwnd:
I've reverted the change, and remove the functions as before, but now just querying the item provider for functions rather than asking Get-Command to give me all publc commands (including cmdlets) that are available in all modules. This seems to take 7ms (and 1 ms after warmup), vs 55ms (and 30ms after warmup), and should not depend on the number of modules that are imported.

nohwnd:
@SQLDBAWithABeard could you try this before I merge please?

SQLDBAWithABeard:
> @SQLDBAWithABeard could you try this before I merge please?

Totally removes those calls from the top of my performance list and gives me other things to work on :-)

nohwnd:
Rob reported this comparison of calls to Get-Command -Name PesterMock_* vs Get-ChildItem function:/PesterMock_*

With 8 modules loaded
20 Milliseconds
2 Milliseconds
With 19 Modules loaded
314 Milliseconds
5 Milliseconds
With 21 Modules loaded inc dbatools/dbacheks
220 Milliseconds 
8 Milliseconds

Insignificant difference between 1st and following runs

PR #2364, opened 2023-06-07
Replace Id with GroupId in Block and Test objects

### LINKED ISSUE #2333: Rename Id-property for parameterized items
> Should we just rename the Id-property to `GroupId` and use a guid to make it more obvious? It's a public property, but I doubt anyone relies on it and the fact that it's startline atm.

_Originally posted by @fflaten in https://github.com/pester/Pester/pull/2208#discussion_r1145464399_

Keep Id-property as obsolete until next major release.

fflaten:
## PR Summary
All blocks and tests have an `Id`-property used to group tests/blocks generated using the `It/Context/Describe` using `-ForEach`.

This PR replaces with a new `GroupId` property to make it more explicit. It also changes the value from start line of `It/Context/Describe` to a generated guid. `Block.Id` and `Test.Id` as marked as obsolete for now and is used as aliases for the new `GroupId`-value. 

Fix #2333

## PR Checklist

- [x] PR has meaningful title


src/csharp/Pester/Block.cs
```diff
@@ -40,7 +40,9 @@ public Block()
         public List<object> ErrorRecord { get; set; }
         public TimeSpan Duration { get => DiscoveryDuration + FrameworkDuration + UserDuration; }
 
-        public string Id { get; set; }
+        [Obsolete("Id is obsolete and should no longer be used. Use GroupId instead.")]
+        public string Id { get => GroupId; }
```
fflaten, line 44: Necessary? Obsolete-attribute isn't visible when invoked in PowerShell iirc.

fflaten:
Tests failed because `Group-Object` sorts by name, so with guids the order is unpredictable in nunit-reports when we group the tests/blocks. Need to either filter by name in tests or reconsider value. TBC.

fflaten:
Changing Id/GroupId from StartLine to random value like a guid would have a minor breaking change.

Given `Describe 'd' -ForEach @(1,2) { It 'i' -ForEach @(1,2) { } }`, `$result.Tests | ? Id | Group-Object { "$($_.Block.BlockContainer.Item):$($_.Id)" }` would return:
- Before PR: A single group with count 4
- After PR: Two groups with count 2, one per block.

The after behavior is the intended behavior and the workaround would be to use `$result.Tests | ? Id | Group-Object { "$($_.Block.BlockContainer.Item):$($_.StartLine)" }`

nohwnd:
Is there a problem with line number as id? Guids are long and complicated, super hard to remember and dynamic. Line ids are mostly static, and short.

Also we use line number in line filter, I don't remember if this would be affected.

fflaten:
> Guids are long and complicated, super hard to remember and dynamic.

`Id/GroupId` doesn't have to be static/reproducible or something you'd reference, does it? Only used to group data-generated blocks/tests, or am I missing something? :)

> Also we use line number in line filter, I don't remember if this would be affected.

It uses the `StartLine`-property

> Is there a problem with line number as id? 

Conflict with multiple groups on same like, typically debug/testing oneliner containers. Felt resonable to fix that while modifying this part. Can't remember why I when with a guid, maybe to avoid making it look redundant due to having the same value as `StartLine`.

A value of `startline:startposition` should suffice though, as mentioned in the original comment and solves the test sorting issue.  Will update soon. 🙂

nohwnd:
> Id/GroupId doesn't have to be static/reproducible or something you'd reference, does it? 

It does not have to be, but if it is it's much easier to debug, especially if you need more attempts. So I would stick with line number, or line:column number if we can. I can fit 3 of these in my head, I cannot fit 3 guids that change in my head.

src/csharp/Pester/Block.cs
nohwnd: It's a good reminder for us internally. Keep.

nohwnd (approved):

fflaten:
Was refreshing this PR yesterday and got reminded why I chose the random id.

With only line:offset you're unable to identity tests coming from the same context (parent block/data), see https://github.com/pester/Pester/pull/2364#issuecomment-1915561774

A workaround would be `...Tests | ? GroupId | Group { $_.Block.GetHashCode() }`, but should a group id not indicate the exact group?

nohwnd:
The id/groupId is just to put together tests that were generated from the same call to Describe / It, so reports can know which tests are coming from the same logical "test group" but we don't have to wrap them into additional layer in the result object. The id needs to be unique only within the current block. It does not have to be globally unique.

```powershell
$result = invoke-pester -Container (New-PesterContainer -ScriptBlock { Describe 'd' -ForEach @(1,2) {  It 'i' -ForEach @(1,2) { };  It 'j' -ForEach @(1,2) { } }; Describe 'd1' -ForEach @(3,4) { It 'i' -ForEach @(3,4) { } } }) -passthru -output detailed

$result.Containers.Blocks | % { $_.GroupId ; $_.Tests | %{ "   $($_.GroupId)"} }
```
This single line test run shows: 

```
4:72
   4:104
   4:104
   4:133
   4:133
4:72
   4:104
   4:104
   4:133
   4:133
4:163
   4:195
   4:195
4:163
   4:195
   4:195
```

And when formatting the code on multiple lines:
```
5:1
   6:5
   6:5
   7:5
   7:5
5:1
   6:5
   6:5
   7:5
   7:5
9:1
   10:5
   10:5
9:1
   10:5
   10:5
```

We had 2 `Describe` blocks and have 2 unique ids on the first level. And we had 3 `It` blocks and have 3 different ids on the second level. So all is fine.

In the previous case we would have all just be 4: 

```
4
  4
  4
  4
  4
4
  etc..
```

So the first two It blocks would be wrongly merged into the same group in report.

Issue #1423, opened 2020-01-12
Way forward for Should?


nohwnd:
Should assertions have multiple quirks and pain points, here is a not-complete list:
- `-Be` does too much, so it is hard to understand what is happening
- Enumeration via pipeline complicates a lot of things and `-Be` has special logic to revert this, which does not always work
- It is difficult to apply seperate special behavior for different types of data (e.g. `Should -BeFalse -AllowNull -AllowTypeCast -FalseStringAsFalse`, `Should -BeString -CaseSensitive`, `Should -BeCollection -All { $_ -like "*file.ps1" }`) because parameter sets limited to 32 parameter sets per function

I researched the various limitations of Should by writing a separate module [Assert](https://github.com/nohwnd/Assert) which has nice functionality but is not used often because people are not aware of it. I think it would be wort it to merge the functionality into Pester and fix some of the quirks finally. But I am wondering what ways forward I actually have. 

There are few more limitations: 

- How to export the Should-* functions without triggering Verb warning? (Pester is getting around this by using single word functions at the moment). 
- How to ensure backward compatibility? (probably Should -Be has to remain in place while Should-Be is added)
- Should this change be done in v5? (if it is parallel then we can do it as non-breaking change).

asears:
Should Be that community use Assert module rather than Should?  Seems much friendlier, some Rules to Better Pester Tests would be good.

Could this be done by aliasing the commands?  By a configuration or feature toggle to enable in Pester v5?  

Some of the features are great in Assert (eg. compare), some are a bit odd naming and could be leveraging some built-in commands (eg. Format-Nicely?) but are useful too!

Is it possible for the assert module to be added to a Microsoft Trusted Repository rather than PSGallery? Or perhaps a wiki page could be added for Azure Devops integration of Pester + Assert.

https://docs.microsoft.com/en-us/azure/devops/artifacts/tutorials/private-powershell-library?view=azure-devops

nohwnd:
@asears Thanks for the suggestions. Assert was my playground to find what is possible, and to look at assertions from a different angle. In some cases I went too far, and then returned back (e.g. -Be<type> which was different per type, which get's pretty annoying especially for numeric types). I think that the biggest problem is getting it to people, because I presented about it multiple times, and I always get asked if it will become part of Pester, as if installing it from the gallery would be difficult. I would also like to see it in Pester, because the current assertions are a bit difficult sometimes. 

[...]

PR #2381, opened 2023-07-19
Feature Request: Expected, Actual, and Because as Test Property Fields

### LINKED ISSUE #1993: Feature Request: Expected, Actual, and Because as Test Property Fields

The vscode proposed test UI has a very nice view window showing expected and actual results including a diff
image

I would like to be able to display this similar information from Pester, but currently since the error is just a string, it would require a lot of parsing.

Ideally a failure would add 3 parameters either to the data of the exception or to the testresult itself:
Example with Should -Be: `Expected $false, but got $true`
1. Expected - $false
2. Actual - $true

Example with Should -BeOfType: `Expected the value to have type [int] or any of its subtypes, but got $true with type [bool].`
1. Expected: [int]
2. Actual: [bool]

Also support Because, which would probably be shown as a comment to Expected

As a workaround a regex parsing of "Expected x, but y" would probably suffice for most scenarios

JustinGrote:
Fixes #1993

## PR Checklist

- [ ] PR has meaningful title

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -693,6 +693,9 @@ function Invoke-TestItem {
                     $Test.Passed = $result.Success
                 }
 
+                $Test.ExpectedValue = $result.ErrorRecord.TargetObject.ExpectedValue
+                $Test.ActualValue = $result.ErrorRecord.TargetObject.ActualValue
+                $Test.BecauseValue = $result.ErrorRecord.TargetObject.BecauseValue
                 $Test.StandardOutput = $result.StandardOutput
                 $Test.ErrorRecord.AddRange($result.ErrorRecord)
             }
--- a/src/csharp/Pester/Factory.cs
+++ b/src/csharp/Pester/Factory.cs
@@ -34,12 +34,12 @@ namespace Pester
             return new List<object>();
         }
 
-        public static ErrorRecord CreateShouldErrorRecord(string message, string file, string line, string lineText, bool terminating)
+        public static ErrorRecord CreateShouldErrorRecord(string message, string file, string line, string lineText, bool terminating, string expectedValue = null, string actualValue = null, string becauseValue = null)
         {
-            return CreateErrorRecord("PesterAssertionFailed", message, file, line, lineText, terminating);
+            return CreateErrorRecord("PesterAssertionFailed", message, file, line, lineText, terminating, expectedValue, actualValue);
         }
 
-        public static ErrorRecord CreateErrorRecord(string errorId, string message, string file, string line, string lineText, bool terminating)
+        public static ErrorRecord CreateErrorRecord(string errorId, string message, string file, string line, string lineText, bool terminating, string expectedValue = null, string actualValue = null, string becauseValue = null)
         {
             var exception = new Exception(message);
             // we use ErrorRecord.TargetObject to pass structured information about the error to a reporting system.
@@ -50,6 +50,9 @@ namespace Pester
                 ["Line"] = line,
                 ["LineText"] = lineText,
                 ["Terminating"] = terminating,
+                ["ExpectedValue"] = expectedValue,
+                ["ActualValue"] = actualValue,
+                ["BecauseValue"] = becauseValue,
             };
             return new ErrorRecord(exception, errorId, ErrorCategory.InvalidResult, targetObject);
         }
--- a/src/csharp/Pester/Test.cs
+++ b/src/csharp/Pester/Test.cs
@@ -36,6 +36,9 @@ namespace Pester
         public string ExpandedPath { get; set; }
 
         public string Result { get; set; }
+        public string ExpectedValue { get; set; }
+        public string ActualValue { get; set; }
+        public string BecauseValue { get; set; }
         public List<object> ErrorRecord { get; set; }
         public object StandardOutput { get; set; }
         public TimeSpan Duration { get => UserDuration + FrameworkDuration; }
--- a/src/functions/assertions/Be.ps1
+++ b/src/functions/assertions/Be.ps1
@@ -38,6 +38,8 @@ function Should-Be ($ActualValue, $ExpectedValue, [switch] $Negate, [string] $Be
     return [PSCustomObject] @{
         Succeeded      = $succeeded
         FailureMessage = $failureMessage
+        ActualValue    = Format-Nicely $ActualValue
+        ExpectedValue  = Format-Nicely $ExpectedValue
     }
 }
 
@@ -111,6 +113,8 @@ function Should-BeExactly($ActualValue, $ExpectedValue, $Because) {
     return [PSCustomObject] @{
         Succeeded      = $succeeded
         FailureMessage = $failureMessage
+        ActualValue    = Format-Nicely $ActualValue
+        ExpectedValue  = Format-Nicely $ExpectedValue
     }
 }
 
--- a/src/functions/assertions/FileContentMatchMultilineExactly.ps1
+++ b/src/functions/assertions/FileContentMatchMultilineExactly.ps1
@@ -63,6 +63,9 @@
     return [PSCustomObject] @{
         Succeeded      = $succeeded
         FailureMessage = $failureMessage
+        ActualValue    = Format-Nicely $ActualValue
+        ExpectedValue  = Format-Nicely $ExpectedContent
+        BecauseValue   = $Because
     }
 }
 
--- a/src/functions/assertions/PesterThrow.ps1
+++ b/src/functions/assertions/PesterThrow.ps1
@@ -127,6 +127,9 @@
         return [PSCustomObject] @{
             Succeeded      = $false
             FailureMessage = $failureMessage
+            ActualValue    = Format-Nicely $actualExceptionMessage
+            ExpectedValue  = if ($filterOnExceptionType) { "type $(Format-Nicely $ExceptionType)" } else { "any exception" }
+            BecauseValue   = $Because
         }
     }
 
--- a/src/functions/assertions/Should.ps1
+++ b/src/functions/assertions/Should.ps1
@@ -258,8 +258,7 @@ function Invoke-Assertion {
     $testResult = & $AssertionEntry.Test -ActualValue $ValueToTest -Negate:$Negate -CallerSessionState $CallerSessionState @BoundParameters
 
     if (-not $testResult.Succeeded) {
-        $errorRecord = [Pester.Factory]::CreateShouldErrorRecord($testResult.FailureMessage, $file, $lineNumber, $lineText, $shouldThrow)
-
+        $errorRecord = [Pester.Factory]::CreateShouldErrorRecord($testResult.FailureMessage, $file, $lineNumber, $lineText, $shouldThrow, $testResult.ExpectedValue, $testResult.ActualValue, $testResult.BecauseValue)
 
         if ($null -eq $AddErrorCallback -or $ShouldThrow) {
             # throw this error to fail the test immediately
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -22,5 +22,6 @@
       "systemId": "https://raw.githubusercontent.com/PowerShell/PowerShell/master/src/Schemas/Types.xsd",
       "pattern": "**/*.Types.ps1xml"
     }
-  ]
+  ],
+  "dotnet.defaultSolution": "src/csharp/Pester.sln"
 }
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -287,6 +287,10 @@ function ConvertTo-PesterResult {
                 }
             }
         }
+
+        $testResult.ExpectedValue = $details.ExpectedValue
+        $testResult.ActualValue = $details.ActualValue
+        $testResult.BecauseValue = $details.BecauseValue
     }
     else {
         $failureMessage = $ErrorRecord.ToString()
--- a/src/functions/assertions/BeGreaterThan.ps1
+++ b/src/functions/assertions/BeGreaterThan.ps1
@@ -17,6 +17,9 @@
         return [PSCustomObject] @{
             Succeeded      = $false
             FailureMessage = "Expected the actual value to be greater than $(Format-Nicely $ExpectedValue),$(Format-Because $Because) but got $(Format-Nicely $ActualValue)."
+            ActualValue    = Format-Nicely $ActualValue
+            ExpectedValue  = Format-Nicely $ExpectedValue
+            BecauseValue  = $Because
         }
     }
 
@@ -50,6 +53,9 @@ function Should-BeLessOrEqual($ActualValue, $ExpectedValue, [switch] $Negate, [s
         return [PSCustomObject] @{
             Succeeded      = $false
             FailureMessage = "Expected the actual value to be less than or equal to $(Format-Nicely $ExpectedValue),$(Format-Because $Because) but got $(Format-Nicely $ActualValue)."
+            ActualValue          = Format-Nicely $ActualValue
+            ExpectedValue        = Format-Nicely $ExpectedValue
+            BecauseValue         = $Because
         }
     }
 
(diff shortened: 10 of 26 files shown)


JustinGrote:
@fflaten  @nohwnd please review

src/csharp/Pester/Test.cs
```diff
@@ -36,6 +36,9 @@ public Test()
         public string ExpandedPath { get; set; }
 
         public string Result { get; set; }
+        public string ExpectedValue { get; set; }
+        public string ActualValue { get; set; }
+        public string BecauseValue { get; set; }
```
fflaten, line 41: Could you provide an example of where this would be used? `-Because` values are likely specifically written to match the assertion error template, so unsure if it's useful by itself.

fflaten, line 41: Do they need to be publicly accessible? If so, we'd need something like a single `ShouldResults[]` property. A test can have multiple assertions and could return multiple errors with `Should.ErrorAction = 'Continue'`. The values can be accessed through `ErrorRecord[i].TargetObject` - good enough?

fflaten (changes requested):
Thanks again for looking into this 👏 
Can't test and fully review for a few days, but added some initial thoughts.

src/csharp/Pester/Test.cs
JustinGrote: It's mostly to pass through so that plug-ins like the vscode extension can surface them without having to parse the message.

JustinGrote: The targetobject can be sufficient but I wanted to surface them in an intellisense-friendly manner, and this is the same process the other properties are extracted.

fflaten: Github didn't highlight my selection too well, so just clarifying that this was specific to `BecauseValue`. :) Will the extension use that at all?

The other comment was about all three.

fflaten: True, but they're also test-specific (single valued) and mostly non-customizable. I guess a `AssertionError[] FailedAssertions`-property could work. Should include `TargetObject.Line` to identify which assertion failed.

`Should` only supports 5-6 custom assertions IIRC due to max parameter sets, but in the future we could consider a `Add-AssertionError` function for third-party assertion modules like [Assert](https://github.com/nohwnd/assert) to integrate with it. Let's see what @nohwnd thinks 🙂

JustinGrote: Apologies, I didn't fully understand what you were communicating, but yes, we could surface the multiple assertion errors as their own structured objects, but I don't think this happens today for failuremessages does it? I was just trying to follow the existing pattern.

JustinGrote: @fflaten I planned to have an option to add the "because" to the test result comparison to make it more surfaced, and it was very little effort to surface it in addition to the others so I went ahead and included it.

JustinGrote: I plan to reformat it to its own object per @fflaten's feedback so all the context will be preserved.

src/functions/assertions/Be.ps1
```diff
@@ -38,6 +38,9 @@ function Should-Be ($ActualValue, $ExpectedValue, [switch] $Negate, [string] $Be
     return [PSCustomObject] @{
         Succeeded      = $succeeded
         FailureMessage = $failureMessage
+        ActualValue    = Format-Nicely $ActualValue
```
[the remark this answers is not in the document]
JustinGrote: Agreed, for now the main purpose is for display of the information, but it can be more raw in the future that format-nicely might be cleaning up, my main goal was to just make sure it stringifies easily.

src/Pester.Runtime.ps1
```diff
@@ -693,6 +693,9 @@ function Invoke-TestItem {
                     $Test.Passed = $result.Success
                 }
 
+                $Test.ExpectedValue = $result.ErrorRecord.TargetObject.ExpectedValue
+                $Test.ActualValue = $result.ErrorRecord.TargetObject.ActualValue
+                $Test.BecauseValue = $result.ErrorRecord.TargetObject.BecauseValue
```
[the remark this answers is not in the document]
JustinGrote: Fixed [83d0352](https://github.com/pester/Pester/pull/2381/commits/83d03523eefd7d16881d251edd9f8166a0690621)

src/csharp/Pester/Test.cs
JustinGrote: Fixed 83d0352

JustinGrote: Fixed 83d0352

JustinGrote:
@fflaten @nohwnd ready for re-review.

I strong typed the result object for all Should results, and kept the info in the TargetObject without exposing it outwardly. Plugins should still be able to access the info as needed.

src/functions/assertions/PesterThrow.ps1
```diff
@@ -124,13 +124,21 @@
...
+            ExpectResult   = @{
+                Actual   = Format-Nicely $ActualValue
+                Expected = Format-Nicely $ExpectedValue
+                Because  = $Because
+            }
         }
     }
 
-    $result = [PSCustomObject] @{
+    $result = [Pester.ShouldResult] @{
```
nohwnd, line 141: Is this (not yours) forgotten code that is not used?

src/functions/assertions/PesterThrow.ps1
```diff
@@ -124,13 +124,21 @@
         $but = Join-And $buts
         $failureMessage = "Expected an exception$(if($filter) { " with $filter" }) to be thrown,$(Format-Because $Because) but $but. $actualExceptionLine".Trim()
 
-        return [PSCustomObject] @{
+    $ActualValue = $actualExceptionMessage
+    $ExpectedValue = if ($filterOnExceptionType) { "type $(Format-Nicely $ExceptionType)" } else { 'any exception' }
+
+        return [Pester.ShouldResult] @{
```
nohwnd, line 130: Will people be still able to register their own assertions without using our types? I think so, but are there tests for it?

src/functions/assertions/PesterThrow.ps1
```diff
@@ -124,13 +124,21 @@
...
             Succeeded      = $false
             FailureMessage = $failureMessage
+            ExpectResult   = @{
+                Actual   = Format-Nicely $ActualValue
+                Expected = Format-Nicely $ExpectedValue
```
nohwnd, line 135: Many of those calls just reformat a string that is already formatted, or do 2 calls of Format-Nicely separately on the same data. This looks like unnecessary overhead (calling functions is expensive, and Format-Nicely is also pretty expensive), that is done on every assertion, even if no-one will consume the data. Can this be optimized? Should we add a switch to even populate Actual and Expected?

src/functions/assertions/FileContentMatchMultilineExactly.ps1
```diff
@@ -60,9 +60,16 @@
         }
     }
 
-    return [PSCustomObject] @{
+    $ExpectedValue = $ExpectedContent
+
+    return [Pester.ShouldResult] @{
         Succeeded      = $succeeded
         FailureMessage = $failureMessage
+        ExpectResult   = @{
```
nohwnd, line 68: Here it populates the values no matter if the should fails, in some other places it only populates them when should fails. What is the desired way?

nohwnd:
Thanks for the PR, I am slightly worried about degrading perf by re-formatting the data where we don't need to, because the number of interactive runs where the functionality will be used is way smaller than the total of runs where this won't be used, and everyone will pay the price. 

Maybe conditioning it based on option, or env var, or both is the best way forward here. Some options (like CI output) already use this pattern. 

And then taking care to not reformat the same data in the assertions.

src/functions/assertions/PesterThrow.ps1
fflaten: The codepath is used by `{ throw 'omg' } | Should -Throw -PassThru`

src/functions/assertions/PesterThrow.ps1
fflaten: Agreed. `FailureMessage` usually has the same formatted values. Please extract to variable so we can reuse here.

> Should we add a switch to even populate Actual and Expected?
Sounds a bit overkill. If we reuse the formatted values as mentioned above then the performance hit is minimal. This is also exclusive to ailed assertions which you'd normally have few of.

src/functions/assertions/Should.ps1
```diff
@@ -258,8 +258,7 @@ function Invoke-Assertion {
     $testResult = & $AssertionEntry.Test -ActualValue $ValueToTest -Negate:$Negate -CallerSessionState $CallerSessionState @BoundParameters
 
     if (-not $testResult.Succeeded) {
-        $errorRecord = [Pester.Factory]::CreateShouldErrorRecord($testResult.FailureMessage, $file, $lineNumber, $lineText, $shouldThrow)
-
+        $errorRecord = [Pester.Factory]::CreateShouldErrorRecord($testResult.FailureMessage, $file, $lineNumber, $lineText, $shouldThrow, $testResult.ShouldResult)
```
fflaten, line 261:
```suggestion
        $errorRecord = [Pester.Factory]::CreateShouldErrorRecord($testResult.FailureMessage, $file, $lineNumber, $lineText, $shouldThrow, $testResult)
```

Typo. Currently doesn't work.

src/csharp/Pester/Factory.cs
```diff
@@ -50,6 +50,7 @@ public static ErrorRecord CreateErrorRecord(string errorId, string message, stri
                 ["Line"] = line,
                 ["LineText"] = lineText,
                 ["Terminating"] = terminating,
+                ["ShouldResult"] = shouldResult
```
fflaten, line 53: Should we set this when null? Not a change request, just looking to learn from others view on  scenarios like this. As it's just a dict key it wouldn't be breaking.

I doubt anyones using this factory method outside Pester, so 99% of the time `ShouldResult` will be provided.

src/functions/assertions/PesterThrow.ps1
fflaten: As long as they don't provide a object with properties/keys other than `FailureMessage` and `Succeeded`, including `ExpectResult` with wrong type/props. That would cause InvalidCastExpection

src/functions/assertions/Should.ps1
JustinGrote: Whoops good catch!

src/csharp/Pester/Factory.cs
JustinGrote: From a perf standpoint I don't think it makes that big of a difference, but there certainly can be a check if desired, that's not a big deal to implement.

src/functions/assertions/PesterThrow.ps1
JustinGrote: Honestly this was a bit of laziness on my part, I just wanted to get all of the various assertions refactored so I tried to get them into a uniform format, I can go through and optimize so that `Format-Nicely` is only called once and when a value is present.

I had considered consolidating a lot of this logic to a single function to keep it DRY, but that would require some dotsourcing or other handling that the build system might not be able to accomidate.

src/functions/assertions/FileContentMatchMultilineExactly.ps1
JustinGrote: Defer to @fflaten but seems to me it makes sense to not populate it if it's successful for minor performance benefit, the only use that info would be is troubleshooting (e.g. it said successful but it shouldn't have been)

src/csharp/Pester/Factory.cs
fflaten: Was thinking more about user experience. If you'd prefer consistency over only relevant when ex printing `ErrorRecord.TargetObject` to screen. No worries, no change needed :)

src/functions/assertions/PesterThrow.ps1
fflaten: > but that would require some dotsourcing or other

Just curious, what did you think about? As long as the helper would be placed in one of these, ex. Pester.Utility, you'd be fine

src/functions/assertions/PesterThrow.ps1
fflaten: `FailureMessage` and `Succeeded` are documented as [required properties in custom assertions](https://pester.dev/docs/assertions/custom-assertions#requirements--interface). As this only affects `Should`-registered operators, is documented and any additional properties/keys are inaccessible for the user, I don't think this is breaking in any way - even though validation will be stricter.

> I think so, but are there tests for it?

I don't think we have tests actually invoking custom assertions atm. Only `Add-ShouldOperator.ts.ps1` that makes sure they get registered properly. Should definitely add some P-tests.

JustinGrote:
@fflaten The strong typing already showed me that `Should -Invoke` is actually defined in Mock and not in its own assertion, and was emitting pscustomobjects from there. There's also somewhere that multiple "trues" along with the final object are being sent. I'll add some type guards to Invoke-Assertion for this and hunt down the bugs.

fflaten:
> The strong typing already showed me that Should -Invoke is actually defined in Mock and not in its own assertion

Yeah, not sure about the history there. I'm considering a PR to move those, `Invoke-Pester` and a couple others functions not used during import to separate files to make it easier to find them.

> There's also somewhere that multiple "trues" along with the final object

Sounds like a bug and another benefit of strongly typing 🙂 

It should've been typed already as the output requirements are documented and cast works. There's no way to access additional properties through `Should`, and **hopefully** users would stick to either `Should -MyCustomAssertion` (required output) or direct calls to function `MyCustomAssertion` (no required output besides throw on error) consistently.

However, I guess there's a slight chance someone leverages the loose mapping to include something for ex. unit-testing `MyCustomAssertion`-function directly. So probably smart to keep it loose until 6.0 and/or [Better Should](https://github.com/pester/Pester/milestone/9). Specifically #1423 might shakes things up either way.

JustinGrote:
I dropped the `Create` method back to object (34b47445f3dc2b826a786e74f4f6151c7de691b5) but all builtin items will still use the result type and I'll also fix up the shouldInvoke output as well to include actual/expected.

src/csharp/Pester/Factory.cs
JustinGrote: I made this conditional in 
34b4744

src/functions/assertions/FileContentMatchMultilineExactly.ps1
JustinGrote: All assertions should now be adjusted to not include this property on a success. Maybe we need tests for this?

src/functions/assertions/PesterThrow.ps1
JustinGrote: Mostly that a lot of the boilerplate around negate, expected, and actual could be consolidated into a common function for about 80% of the cases. I'll leave it for now as I've already fixed them in place but if I have to fix them again I'll definitely consider this option.

src/functions/Mock.ps1
```diff
@@ -452,43 +468,68 @@ function Should-InvokeInternal {
...
+                }
             }
         }
     }
 
-    return [PSCustomObject] @{
+    return [Pester.ShouldResult] @{
         Succeeded      = $true
-        FailureMessage = $null
+        FailureMessage = $null #TODO Pester 6: This should be ommitted for consistency but someone may have a strict check for null that may fail on the absence of this property.
```
fflaten, line 532: Should be okay to remove. This is internal function. Output-object not surfaced to user without calling `Should-Invoke/InvokeInternal` inside Pester-module which is unsupported

JustinGrote: 757b66043cce99e25eed591a52d7c46376000d57 done

nohwnd:
I've run it bunch of times, it does not seem to impact the speed negatively. Okay to merge imho. 

@fflaten wanna have one last look?

JustinGrote:
@nohwnd thank you!

nohwnd:
Thank you both :)

PR #2391, opened 2023-09-20
FixStacktraceLanguage

Akturos:
## PR Summary
"Stacktrace is not filtered in non-english system languages" I found the PR #2062 and completed the approach to fix the issue.
The new function get's the correct StackTracke language informations and passes them to necessary the regex filters.

## PR Checklist

- [X] PR has meaningful title


fflaten:
Thanks for looking into this. Will test and review later, but wanted to quickly mention this draft PR #2276. They differ in a couple ways:
- You gather language at run start vs on every call
  - Pro: Better performance 👍
- You use resource (smart!) vs throwing a real exception
  - Pro: Performance (?) and very clean
  - Con: Not compatible with PSv3 and 4 which we currently support. PS v3 and v4 are end-of-life so they're not critical, but atm. Pester supports them. So maybe throwing a real exception like the linked PR is worth it for now?

It would be nice to have a way to test this during our CI to avoid breaking it in the future. Ex. by accepting an optional `[string]$StackTrace` parameter so we could test the regex-pattern using a 2-3 language samples.

Akturos:
Hi @fflaten  

I have adapted the code according to your wishes and added a fallback scenario for all versions to the regex of PR #2276. I also separated this function and added a parameter [string]$StackTrace which allows you to test this function. But after the commit there was a bug in PS6_2 

https://github.com/pester/Pester/blob/7ca9c814cf32334303f7c506beaa6b1541554973/tst/Pester.RSpec.Demo.ts.ps1#L132C3-L132C68

which points to a runtime error. Can you please run the failed tests again, I think it's probably a onetime timing error.

src/Pester.Runtime.ps1
```diff
@@ -58,6 +58,59 @@ $flags = [System.Reflection.BindingFlags]'Instance,NonPublic'
...
+            else {
+                $StacktraceLanguage = [PSCustomObject]@{
+                    At   = $Matches["At"]
+                    Line = $Matches["Line"]
+                }
+            }
+        }
+        else {
+            #Prior to version 5 there are no ressources in the Assembly available (System.Management.Automation)
+            throw 'Fallback for PSVersion 3/4'
```
nohwnd, line 103: Using exceptions for control flow is not ideal, this will always throw for powershell 4, if you just reduce the scope of the `try` to be in the positive leg of the if it should make the code no more complicated, and avoid this issue.

nohwnd (commented):
Thanks for the PR :) 

As mentioned above, a test would be nice. I know there is Using-Culture either in this codebase or easily found on the internet, which should make this testable, and not dependent on the OS language. 

Also a nitpick: local variables and keywords start with lowercase in this codebase.

nohwnd:
@Akturos will you address that PR remarks, or should I finish it up so we can ship it? :)

Akturos:
@nohwnd 

When modifying the code, I noticed that without throwing a exception for PS version 3/4, the necessary values for the stacktracelanguage can no longer be read from the self-generated error. This raises the question of whether you really want it that way, as the error can also be used to generate the appropriate stacktracelanguage for PS version 3/4. If the error is removed, this will no longer work and only English will be the default for PS version 3 / 4.

I have adapted the variables to the default notation.

Regarding the test. I have tried to build a test with the cultures. However, this is not possible, or I have not managed to set the cultures in the PS session in such a way that I could use them for a test. 

`            & {
            $stackTraceRessourceName = 'DebuggerStrings'
            [System.Threading.Thread]::CurrentThread.CurrentCulture = 'fr-FR'
            [System.Threading.Thread]::CurrentThread.CurrentUICulture = 'fr-FR'
            $r = [System.Resources.ResourceManager]::new($stackTraceRessourceName, [System.Reflection.Assembly]::GetAssembly([System.Management.Automation.Breakpoint]))
            $r.GetString('StackTraceFormat') 
            }
`
Maybe you have a better idea.

Cheers 
Acturos

nohwnd:
> When modifying the code, I noticed that without throwing a exception for PS version 3/4, the necessary values for the stacktracelanguage can no longer be read from the self-generated error. 

We can probably throw once and get the elements from that? And then use normal control flow? Extensions are not super expensive, on my system it is 30ms to throw and catch 1000 of them, still it is IMHO unnecessary to throw and catch on every assertion fail.

claudiospizzi:
Hi all

I agree with @fflaten - the Idea was to store the stack trace language once in module scoped variable `$Script:stackTraceLanguage`. This is done in the script `Pester.Runtime.ps1`, which is loaded once during module import, right?

Throwing an exception once and only for rare cases (PSVersion < 5 or if there is any unexpected issue with the resource manager) sounds acceptable.

Would be cool, if we can merge that to have a better experience for non-English systems.

Thank you.

Akturos:
Hi all
I also agree with @fflaten. I have reset the branch to the commit ce24250aa54e9236879d873b7af868c6b4e3a272. I think this is the most useful solution. I have also adjusted the variables.

It would be great if we could ship this to prdouction

Akturos:
Happy new year all.

Is there an update from your side guys? Should anything else be adjusted or can we ship it. Thanks for your feedback.

src/Pester.Runtime.ps1
```diff
@@ -58,6 +58,59 @@ $flags = [System.Reflection.BindingFlags]'Instance,NonPublic'
...
+function Get-StackTraceLanguageFallBack {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$StackTrace
+    )
+    $StackTraceLanguage = [PSCustomObject]@{
+        At   = "at"
+        Line = "line"
+    }
+    $Regex = "(?<At>.*)\s(?<ScriptBlockOrFunction>\<\w+\>),\s(?<FileName>\<.+\>)\s*:\s(?<Line>\w+)\s(?<LineNumber>\w+)\z"
```
fflaten, line 71:
```suggestion
    $Regex = "(?<At>.*)\s(?<ScriptBlock>\<\w+\>),\s(?<NoFile>\<.+\>)\s*:\s(?<Line>.+)\s(?<LineNumber>\d+)\z"
```
Not sure if "line" could be multiple words, but regex should be consistent between this and `Get-StackTraceLanguage`.

Also consider comment to make it clear that this only matches final line in stacktrace with scriptblock-source, e.g. `at <ScriptBlock>, <No file>: line 1`

src/Pester.Runtime.ps1
fflaten: Second this. Please move fallback exception to be part of `Get-StackTraceLanguageFallBack`. That way we can remove the large try/catch here and just call the fallback if no result or PS <5.

src/Pester.Runtime.ps1
```diff
@@ -58,6 +58,59 @@ $flags = [System.Reflection.BindingFlags]'Instance,NonPublic'
...
+        $StackTraceLanguage.At = $Matches["At"]
+        $StackTraceLanguage.Line = $Matches["Line"]
+    }
+    Return $StackTraceLanguage
+}
+
+function Get-StackTraceLanguage {
+    #Full fallback scenario to the solution of PR #2276 in case of error
+    try {
+        if ($PSVersionTable.PSVersion.Major -ge 5) {
```
fflaten, line 82: Consider reversing this. Easier to read and less nesting.

```powershell
# Required string resource missing prior to version 5 - use fallback
if ($PSVersionTable.PSVersion.Major -lt 5) {
    return (Get-StackTraceLanguageFallBack)
}

if ($PSVersionTable.PSVersion.Major -gt 5) {
    $StackTraceRessourceName = 'System.Management.Automation.resources.DebuggerStrings'
    .... # rest of PS 5+ code.
```

fflaten (changes requested):

src/Pester.Runtime.ps1
fflaten: Second this. Please move fallback exception to be part of `Get-StackTraceLanguageFallBack`. That way we can remove the large try/catch here and just call the fallback if no result or PS <5.

fflaten:
Happy new year! 🙂 I've added a review with a few change requests.

I think @nohwnd's a bit busy, so final approval + merge might take a little time.

nohwnd:
I am here now and will be for few more weeks at least, so best time to resume work on this :))

src/Pester.Runtime.ps1
```diff
@@ -59,6 +59,20 @@ $script:SessionStateInternalProperty = [System.Management.Automation.SessionStat
...
+    $firstFrame = ($err.ScriptStackTrace -split [System.Environment]::NewLine, 2 )[0]
+    if ($firstFrame -match "(?<at>.[^G]?)\s+Get-StackTraceLanguage(?<separator>.).+:\s+(?<line>.\S+)\s\d+") {
+        @{
+            At        = $Matches["at"]
+            Separator = $Matches["separator"]
+            Line      = $Matches["line"]
+        }
+    }
+}
+#defining script variable
```
nohwnd, line 73:
```suggestion
```

src/Pester.Runtime.ps1
```diff
@@ -59,6 +59,20 @@ $script:SessionStateInternalProperty = [System.Management.Automation.SessionStat
...
+    $err = try { throw "err" }  catch { $_ }
+    $firstFrame = ($err.ScriptStackTrace -split [System.Environment]::NewLine, 2 )[0]
+    if ($firstFrame -match "(?<at>.[^G]?)\s+Get-StackTraceLanguage(?<separator>.).+:\s+(?<line>.\S+)\s\d+") {
+        @{
+            At        = $Matches["at"]
+            Separator = $Matches["separator"]
+            Line      = $Matches["line"]
+        }
+    }
+}
```
nohwnd, line 72: Changed your code to parse it from exception that we throw, as originally suggested. This should be cheaper and easier to replicate than relying on internal resource names because we rely on public api.

nohwnd:
Please figure out a way to test this. Changing current thread culture did not affect stack trace for me. And I don't think the proposed code was working because the regex was rewritten few lines below by the `if ($true)` code. So I would like a way to ensure this works and keeps working.

```powershell
[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo("cs-cz")
[System.Threading.Thread]::CurrentThread.CurrentUiCulture = [System.Globalization.CultureInfo]::GetCultureInfo("cs-cz")

1.5
function a {
    try { throw "a" } catch { $_.ScriptStackTrace }
}
```

nohwnd:
I would love to fix, but we don't have a repro for the original problem so it looks like it is not a problem anymore. If someone finds out how to repro, or even better, test this. We can re-open.

Issue #1888, opened 2021-04-01
Should -HaveParameter should check DefaultParameterType 


nohwnd:
Default parameter type is figured out in the code, but parameter for it is not surface to Should which is unfortunate when you check `$Path = (Get-DefaultPath)` because our assertion will pass for both `$Path = (Get-DefaultPath)`  and `$Path = "(Get-DefaultPath)"`. 

But if the DefaultValueType was surfaced then we could say the the value should be Expression, not a string, and we would be able to distinguish those two.

nohwnd:
@lipkau ^^^ would you consider PRing it? 🙂

fflaten:
To clarify, do we want to require `-DefaultValue { something }` when the `DefaultValueType` is `Expression`? Like this?

```powershell
Describe 'd' {
    It 'i' {
        function myDefault { 'my\default' }
        function myFunc {
            param(
                [string]$MyParam = '(myDefault)'
            )
        }

        # This should work
        Get-Command myFunc | Should -HaveParameter MyParam -DefaultValue {(myDefault)}

        # This shouldn't
        Get-Command myFunc | Should -HaveParameter MyParam -DefaultValue '(myDefault)'
    }
}
```

We still end up comparing them as strings, so it would mostly be for the optics right? If so, is it worth the breaking change? Am I missing something?

_Would likely have to trim the scriptblock-text since formatter in VSCode likes to add whitespace between curly-braces and content in `{content}`_

nohwnd:
I think I just meant that we will always compare them as strings, but we will have -DefaultValueType parameter which can would be the actual type of the default value: 

$Path = (Get-DefaultPath) == -DefaultValue "(Get-DefaultPath)" and -DefaultValueType ScriptBlock
$Path = (Get-DefaultPath) != -DefaultValue "(Get-DefaultPath)" and -DefaultValueType String

(or whatever is the name of the type on the default value, is that Expression?)

PR #2398, opened 2023-10-16
Fix Should -HaveParamter -DefaultValue

### LINKED ISSUE #2388: Should -HaveParameter -DefaultValue does not work for boolean types or integer value of 0 (regression)
### Checklist

### What is the issue?

Pester version `5.4.1` has a regression from `5.4.0` where `Should -HaveParameter -DefaultValue` no longer works for the following kinds of tests that previously passed:

- Function parameter of type `[int]` with a default value of `0`
- Function parameter of type `[bool]` with any default value set (both `$true` and `$false`)

## Sample test
```powershell
Describe 'Parameter default value tests' {
    BeforeAll {
        function Test-Parameter {
            param (
                [Parameter()]
                [int]$intParam = 0,

                [Parameter()]
                [bool]$boolTrueParam = $true,

                [Parameter()]
                [bool]$boolFalseParam = $false
            )
        }
    }

    It 'Default value when integer is zero' {
        Get-Command Test-Parameter | Should -HaveParameter 'intParam' -Type 'int' -DefaultValue 0
    }

    It 'Default value for true boolean' {
        Get-Command Test-Parameter | Should -HaveParameter 'boolTrueParam' -Type 'bool' -DefaultValue $true
    }

    It 'Default value for false boolean' {
        Get-Command Test-Parameter | Should -HaveParameter 'boolFalseParam' -Type 'bool' -DefaultValue $false
    }
}
```

In Pester `5.4.0`, these worked as expected:

```powershell
Running tests from 'C:\source\PesterBug\Test-Parameter.Tests.ps1'
Describing Parameter default value tests
  [+] Default value when integer is zero 36ms (6ms|30ms)
  [+] Default value for true boolean 6ms (5ms|2ms)
  [+] Default value for false boolean 14ms (6ms|9ms)
Tests completed in 217ms
Tests Passed: 3, Failed: 0, Skipped: 0 NotRun: 0
```

In Pester `5.4.1`, these fail with:
```powershell

[-] Parameter default value tests.Default value for integer 8ms (7ms|1ms)
 Expected command Test-Parameter to have a parameter intParam, of type [System.Int32] and the default value to be '0', but the default value was <empty>.
 at Get-Command Test-Parameter | Should -HaveParameter 'intParam' -Type 'int' -DefaultValue 0, C:\source\PesterBug\Test-Parameter.Tests.ps1:21
 at <ScriptBlock>, C:\source\PesterBug\Test-Parameter.Tests.ps1:21
[-] Parameter default value tests.Default value for true boolean 17ms (16ms|1ms)
 Expected command Test-Parameter to have a parameter boolTrueParam, of type [System.Boolean] and the default value to be 'True', but the default value was '$true'.
 at Get-Command Test-Parameter | Should -HaveParameter 'boolTrueParam' -Type 'bool' -DefaultValue $true, C:\source\PesterBug\Test-Parameter.Tests.ps1:25
 at <ScriptBlock>, C:\source\PesterBug\Test-Parameter.Tests.ps1:25
[-] Parameter default value tests.Default value for false boolean 19ms (18ms|1ms)
 Expected command Test-Parameter to have a parameter boolFalseParam, of type [System.Boolean] and the default value to be 'False', but the default value was '$false'.
 at Get-Command Test-Parameter | Should -HaveParameter 'boolFalseParam' -Type 'bool' -DefaultValue $false, C:\source\PesterBug\Test-Parameter.Tests.ps1:29
 at <ScriptBlock>, C:\source\PesterBug\Test-Parameter.Tests.ps1:29
```

## Regression

This appears to have been caused by https://github.com/pester/Pester/commit/ea7045190d5ed09194fb5ff4ddc16696b40aeaf8. I used a debugger to check the value of `$paramInfo` in `Get-ParameterInfo` and it looks different when the function was changed to use the AST.

### Boolean before and after

```powershell
# Pester 5.4.0
Name      Type      DefaultValue DefaultValueType
----      ----      ------------ ----------------
boolParam [bool]    true         Variable

# Pester 5.4.1
Name      Type      DefaultValue DefaultValueType
----      ----      ------------ ----------------
boolParam [boolean] $true        Boolean
```

### Integer before and after

```powershell
# Pester 5.4.0
Name     Type    DefaultValue DefaultValueType
----     ----    ------------ ----------------
intParam [int]   0            Number

# Pester 5.4.1
Name     Type    DefaultValue DefaultValueType
----     ----    ------------ ----------------
intParam [int32] 0            Int32
```

### Expected Behavior

These kind of tests with `-DefaultValue` should pass.

### Steps To Reproduce

_See working sample test above in issue description_

### Describe your environment

```
Pester version     : 5.4.1 C:\Users\phoerr\Documents\PowerShell\Modules\Pester\5.4.1\Pester.psm1
PowerShell version : 7.3.6
OS version         : Microsoft Windows NT 10.0.19045.0
```

### Possible Solution?

_No response_

nohwnd:
## PR Checklist

- [x] PR has meaningful title

Fix #2388

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -98,12 +98,38 @@
                 # When the value has a known fully realized value (indicated by .Value being on the DefaultValue object)
                 # we take that and use it, otherwise we take the extent (how it was written in code). This will make
                 # 1, 2, or "abc", appear as 1, 2, abc to the assertion, but (Get-Date) will be (Get-Date).
-                $paramInfo.DefaultValue = if ($parameter.DefaultValue.PSObject.Properties["Value"]) { $parameter.DefaultValue.Value } else { $parameter.DefaultValue.Extent.Text }
+                $paramInfo.DefaultValue = Get-DefaultValue $parameter.DefaultValue
             }
 
             $paramInfo
             break
         }
+    }
+
+    function Get-DefaultValue {
+        param($DefaultValue)
+
+        # This is a value like 1, or 0, return it direcly.
+        if ($DefaultValue.PSObject.Properties["Value"]) {
+            return $DefaultValue.Value
+        }
+
+        # This is for backwards compatibility with Pester v5.4.0.
+        # Existing assertions check for -DefaultValue "false", while the definition
+        # of the function says $MyParam = $false.
+        if ('$true' -eq $DefaultValue.Extent.Text -or '$false' -eq $DefaultValue.Extent.Text) {
+            # returns "true", or "false" without $ prefix
+            return $DefaultValue.VariablePath
+        }
+
+        # This is for backwards compatibility with Pester v5.4.0.
+        # Existing assertions check for -DefaultValue "", while the definition
+        # of the function says $MyParam = $null or $MyParam without any default value.
+        if ('$null' -eq $DefaultValue.Extent.Text) {
+            return ""
+        }
+
+        $DefaultValue.Extent.Text
     }
 
     function Get-ArgumentCompleter {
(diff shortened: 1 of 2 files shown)


tst/functions/assertions/HaveParameter.Tests.ps1
```diff
@@ -345,6 +345,46 @@ InPesterModuleScope {
...
+
+            Get-Command Test-Parameter | Should -HaveParameter 'objParam' -Type 'object' -DefaultValue $null
+        }
+
+        It 'fails when object parameter expects default parameter value $null, but there is none' {
+            function Test-Parameter {
+                param ( [Parameter()] [object] $objParam )
+            }
+
+            { Get-Command Test-Parameter | Should -HaveParameter 'objParam' -Type 'object' -DefaultValue $null } | Should -Throw "aaaa"
```
nohwnd, line 361: This test is incomplete. And I need more tests on negative side.

tst/functions/assertions/HaveParameter.Tests.ps1
```diff
@@ -345,6 +345,46 @@ InPesterModuleScope {
...
+            function Test-Parameter {
+                param ( [Parameter()] [int] $intParam = 0 )
+            }
+
+            Get-Command Test-Parameter | Should -HaveParameter 'intParam' -Type 'int' -DefaultValue 0
+        }
+
+        It 'passes when bool parameter has default parameter value $true' {
+            function Test-Parameter {
+                param ( [Parameter()] [bool] $boolParam = $true )
```
nohwnd, line 374: This test fails, because the param has value, but we don't lift the value from it properly because there is none, because $true is variable descriptor.

src/functions/assertions/HaveParameter.ps1
```diff
@@ -71,20 +73,30 @@
...
-                    $paramInfo.DefaultValue = $parameter.DefaultValue.Extent.Text
-                }
+                # The actual value of the default value can be falsy (e.g. $null, $false or 0)
+                # use this flag to communicate if default value was found in the AST or not,
+                # no matter if the actual default value is falsy.
+                # That is: param($param1 = $false) will set this to true for $param1
+                # but param($param1) will have this set to false, because there was no default value.
+                $paramInfo.HasDefaultValue = $true
+                # $null does not have a Value
+                $paramInfo.DefaultValue = if ($null -ne $parameter.DefaultValue.Value) { $parameter.DefaultValue.Value } else { $parameter.DefaulValue.Extent }
```
nohwnd, line 99: This is problematic, because we are lifting the value from AST it comes either as a literal when it is a constant (e.g. 0), but as semantic tree when it is not a literal, e.g. $true in the simplest case. Which is a variable descriptor. 

I think this will be problematic for variables, and also for other default values that need evaluation, we probably don't want to evaluate them ourselves.

nohwnd:
There is some hidden potential in this, but I don't think it is worth the breaking change. I've reverted to the previous behavior where we cannot explicitly determine whether or not the default value was specified at all or if it was $null. 

We also cannot check for the exact literal of the default type, to distinguish '(Get-Date)' string and (Get-Date) expression. 

BUT the first feature does not have enough potential to warrant the breaking change, because there is not much difference between providing null default, and not providing one at all. 

And the second one can be overcome by #1888 and still is an edge case.

nohwnd:
@fflaten  merged, let me know if you had any concerns :)

Issue #2400, opened 2023-10-31
Bring back 'InconclusiveCount' property (or alternative) on returned Pester.Run object


csandfeld:
### Checklist

- [X] Feature request has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### Summary of the feature request

In Pester v4 I have used `Set-ItResult -Skipped` and `Set-ItResult -Inconclusive` in tests, and distinguished between the two when generating reports of the results. In v4 it was easy because InconclusiveCount was returned, but while working on moving to v5, I noticed the property was missing, and found this [breaking changes](https://pester.dev/docs/migrations/breaking-changes-in-v5) note in the docs:

> Inconclusive and Pending states are currently no longer available, -Pending and -Inconclusive are translated to -Skip both on test blocks and when using Set-ItResult

What are the odds of this property - or some alternative means of reliably reporting on "Skipped reason" - to find it's way to the Pester.Run object?

### How should it work?

Surface properties similar to what is available in Pester v4.

Alternatively (might be easier to implement) include some kind of "reason data" in the exception thrown by the Set-ItResult command. One option could be to include the data in the [targetObject Dictionary](https://github.com/pester/Pester/blob/a50354e46344aa8d4b4460e55fd2137939d0a215/src/csharp/Pester/Factory.cs#L46) object that is already included in the exception. In theory that logic could also allow specifying custom reasons with a parameter for the `Set-ItResult` command.

fflaten:
You can still identify the tests skipped using `Set-ItResult -Inconclusive` by checking the errorrecord stored in the test. Not saying it's elegant, so just suggesting it as a workaround 🙂 

```powershell
$conf = New-PesterConfiguration
$conf.Output.Verbosity = 'Detailed'
$conf.Run.ScriptBlock = {
    Describe 'a1' { 
        Context 'b1' { 
            It 'i1' {
                Set-ItResult -Inconclusive -Because 'demo'
            }
            It 'i2' -Skip {
                1 | Should -Be 1
            }
        }
    }
}
$conf.Run.PassThru = $true

$run = Invoke-Pester -Configuration $conf
[...]

PR #2401, opened 2023-11-01
Set-ItResult: Return distinctive ErrorId depending on switch used

csandfeld:
## PR Summary

- Set-ItResult updated to return distinctive ErrorId depending on switch used (previously always returned `PesterTestSkipped`).
  - `-Inconclusive` now returns `PesterTestInconclusive`
  - `-Pending` now returns `PesterTestPending`
  - `Skipped` still returns `PesterTestSkipped`
- Invoke-TestItem in Pester.Runtime updated to handle additional ErrorId´s.
- Existing Set-ItResult Pester tests re-enabled.

I am not sure if these changes require additional tests and/or documentation updates - let me know if they do.

Related to issue #2400.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -680,7 +680,7 @@ function Invoke-TestItem {
 
                 $Test.FrameworkData.Runtime.ExecutionStep = 'Finished'
 
-                if ($Result.ErrorRecord.FullyQualifiedErrorId -eq 'PesterTestSkipped') {
+                if (@('PesterTestSkipped', 'PesterTestInconclusive', 'PesterTestPending') -contains $Result.ErrorRecord.FullyQualifiedErrorId) {
                     #Same logic as when setting a test block to skip
                     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
                         $path = $Test.Path -join '.'
--- a/src/functions/Output.ps1
+++ b/src/functions/Output.ps1
@@ -753,7 +753,7 @@ function Get-WriteScreenPlugin ($Verbosity) {
             $margin = $ReportStrings.Margin * ($level)
             $error_margin = $margin + $ReportStrings.Margin
             $out = $_test.ExpandedName
-            if (-not $_test.Skip -and $_test.ErrorRecord.FullyQualifiedErrorId -eq 'PesterTestSkipped') {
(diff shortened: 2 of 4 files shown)


nohwnd (approved):

src/Pester.Runtime.ps1
```diff
@@ -680,7 +680,7 @@ function Invoke-TestItem {
 
                 $Test.FrameworkData.Runtime.ExecutionStep = 'Finished'
 
-                if ($Result.ErrorRecord.FullyQualifiedErrorId -eq 'PesterTestSkipped') {
+                if (@('PesterTestSkipped', 'PesterTestInconclusive', 'PesterTestPending') -contains $Result.ErrorRecord.FullyQualifiedErrorId) {
```
nohwnd, line 683: @fflaten  what you commented on in the Issue happens here. The error record id is translated to .Skipped = $true on the test object. So all the accounting that is done above this is correct, but you can still see the actual reason for marking the test as skipped on the ErrorRecord itself on the Result object.

fflaten: Thanks 🙂  Didn't remember that `Add-RSpecTestObjectProperties` would set the correct result based on `Skipped = $true` during teardown.

fflaten:
LGTM as a first step. 🙂 

I'd prefer to change the output a little, but unfortunately it could be a minor breaking change since it was the only identifier for inconclusive/pending until now. E.g.
```
  # From
  [!] Test123 is skipped, because INCONCLUSIVE: test 245.6s (143.4s|102.2s)
  [!] Test456 is skipped, because PENDING: test 5ms (2ms|3ms)
  
  # To (until one or both a proper result values again)
  [!] Test123 is skipped (inconclusive), because test 245.6s (143.4s|102.2s)
  [!] Test456 is skipped (pending), because test 5ms (2ms|3ms)
```

nohwnd:
@fflaten  yeah let's not change the output, so we don't break the existing workarounds, while there is no complete solution in place.

nohwnd:
Merged, thanks guys :)

csandfeld:
> Merged, thanks guys :)

My pleasure :)

billgothacked:
Okay

Issue #2235, opened 2022-08-09
Support PSModuleInfo as argument to `-ModuleName` and `InModuleScope`


fflaten:
### Checklist

- [X] Feature request has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### Summary of the feature request

Pester currently reference modules by name. This requires modules to be imported and only one instance per name.

At the same time, it blocks us from executing code or injecting mocks inside un-imported dynamic modules and cause some inconsistency with nested modules (see https://github.com/pester/Pester/pull/2234#issuecomment-1209620140)

### How should it work?

```powershell
$someNestedModule = (Get-Module ManifestModule).NestedModules[0]
Mock SomeCommand -ModuleName $someNestedModule
```

fflaten:
Potential issues with mock history and cleanup in this scenario.
Is there anything identifying a module instances out-of-box? Or would we have to keep a copy of the PSModuleInfo to compare with?

nohwnd:
If we do allow it, will it be possible for the users to grasp the difference, or is that adding some additional complication to the already complicated world of Mock? 

I can see how this could be useful in a rare case where you have a nested module that you are able to access from a different module, and passing on the instance, but is it useful enough in other cases? 

Implementing this would possibly mean re-implementing mock searching and history, and maybe breaking some of the current behavior. 

What if we simply added a convention like Module|SubModule, and allowed only one level, and inspected NestedModules on behalf of the user? Then maybe all the functionality can remain the same, we will just change the way we resolve the modules.

There currently is some indirect reliance on module instances, e.g. InModuleScope attaches your ScriptBlock to a concrete instance of a module, and if you re-import it, then your module will no longer be the same instance. So in this case it would be easier to detect errors. On the other hand I am not sure if we are relying on the opposite behavior, where the name is universal identifier for any instance of that module, anywhere.

fflaten:
Thanks for the feedback. Closing for now as cost outweighs reward atm. Can reopen if it becomes a popular request in the future 🙂

PR #2412, opened 2023-12-11
Allow using 'RootModule/NestedModule' as ModuleName

fred-gagnon:
## PR Summary

Updated function `Get-CompatibleModule` to allow specifying a module name as `RootModule/NestedModule`. This change solves an issue where multiple modules with a similarly named nested module cause an exception `Multiple script or manifest modules named...` to be thrown when attempting to create a mock in a nested module.

## PR Checklist

- [X] PR has meaningful title
- [X] Tests are added/update *(if required)*

### DIFF
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -176,7 +176,12 @@ function Get-CompatibleModule {
         if ($PesterPreference.Debug.WriteDebugMessages.Value) {
             Write-PesterDebugMessage -Scope Runtime "Searching for a module $ModuleName."
         }
-        $modules = @(& $SafeCommands['Get-Module'] -Name $ModuleName -All -ErrorAction Stop)
+        if ($ModuleName -match '(?<RootModule>\w+)[/\\](?<NestedModule>\w+)') {
+            $modules = @(& $SafeCommands['Get-Module'] -Name $Matches['RootModule'] -All -ErrorAction Stop | & $SafeCommands['Select-Object'] -ExpandProperty NestedModules | & $SafeCommands['Where-Object'] { $_.Name -eq $Matches['NestedModule'] })
+        }
+        else {
+            $modules = @(& $SafeCommands['Get-Module'] -Name $ModuleName -All -ErrorAction Stop)
+        }
     }
     catch {
         throw "No modules named '$ModuleName' are currently loaded."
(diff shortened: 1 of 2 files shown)


fflaten:
Thanks for the PR. Will review this soon.
Related https://github.com/pester/Pester/issues/2235#issuecomment-1264617847

fred-gagnon:
> Thanks for the PR. Will review this soon. Related [#2235 (comment)](https://github.com/pester/Pester/issues/2235#issuecomment-1264617847)

Any idea when it will be reviewed ?

fflaten:
Sorry for the delay. I've been fully occupied at work for a while but will hopefully get some time for OSS this month.

I'm also waiting on @nohwnd to return (find time). He'll need to eventually merge and release the pending PRs once approved.

nohwnd:
Can you please add a test that uses InModuleScope with this syntax and proves that we ended up in the correct context? E.g. by checking `$ExecutionContext.SessionState.Module.Name` or a value of a variable.

This would also solve as an example usage.

fred-gagnon:
@nohwnd The tests you requested have been added

fflaten:
@nohwnd Won't this also allow mocking, while cleanup will fail?

Haven't had the time to test it yet, but that's a concern I had.

nohwnd:
Looks like it would impact Mock, because the function is common. @fred-gagnon can you please allow this only from InModuleScope (e.g. by bool parameter that needs to be provided)? I am not sure what the impact is on mocking, and I won't be able to deeply investigate this any time soon.

fred-gagnon:
@nohwnd The whole point of this PR is for use-cases I have where I need to inject a Mock in a sub-module and there is more than one sub-modules currently loaded which have the same name, for example when 2 REST clients are simultaneously loaded and each one has a Repository sub-module. If I am to limit the use of this change only for InModuleScope, then it is pointless to keep it open. Unless I can use Mock from within InModuleScope? I’ve never tried that and won’t be able to for a few days.

nohwnd:
I don't know why I was under the impression that this is for InModuleScope, sorry. 

To allow this change for Mocking you need to add ton of tests mocks. I am not trying to be difficult, but we rely on simple name to figure out location of mocks. We also don't inspect child modules to clean them etc. So basically no-one knows what the change would cause (see comment: https://github.com/pester/Pester/issues/2235#issuecomment-1264617847 ), and your scenario is so rare that it does not seem worth it to break the usage for everyone else.

If you want to take it all the way through then I am okay with it, but it won't be easy. 🙂

nohwnd:
Closing due to no activity on this PR, taking it to completion would require the implementation described above, if you want to commit to that please let me know.

Issue #2441, opened 2024-05-01
$configuration.CodeCoverage.UseBreakpoints = to false by default


nohwnd:
### Checklist

- [X] Feature request has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### Summary of the feature request

take this setting from experimental and make it the default, it is significantly faster

PR #2456, opened 2024-05-17
Set profiler based CodeCoverage as default

nohwnd:
Remove experimental flag from Profiler based Code Coverage and use it by default because it is way more performant.
 
Fix https://github.com/pester/Pester/issues/2441

### DIFF
--- a/src/Pester.RSpec.ps1
+++ b/src/Pester.RSpec.ps1
@@ -363,10 +363,10 @@ function New-PesterConfiguration {
       CoveragePercentTarget: Target percent of code coverage that you want to achieve, default 75%.
       Default value: 75
 
-      UseBreakpoints: EXPERIMENTAL: When false, use Profiler based tracer to do CodeCoverage instead of using breakpoints.
-      Default value: $true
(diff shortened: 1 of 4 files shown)


src/Pester.RSpec.ps1
```diff
@@ -363,10 +363,10 @@ function New-PesterConfiguration {
       CoveragePercentTarget: Target percent of code coverage that you want to achieve, default 75%.
       Default value: 75
 
-      UseBreakpoints: EXPERIMENTAL: When false, use Profiler based tracer to do CodeCoverage instead of using breakpoints.
-      Default value: $true
+      UseBreakpoints: When false, use Profiler based tracer to do CodeCoverage instead of using breakpoints.
+      Default value: $false
 
-      SingleHitBreakpoints: Remove breakpoint when it is hit.
+      SingleHitBreakpoints: Remove breakpoint when it is hit. This increases performance of breakpoint based CodeCoverage.
```
fflaten, line 369: Any downside to using this? Remove option and make it default?

PR #2478, opened 2024-05-21
Report code coverage that includes P tests

### LINKED ISSUE #2475: Measure Code Coverage for Pester
### Checklist

Pester is not enabling Code Coverage, probably because we are testing code coverage, and use P tests. Let's see if we can measure coverage somehow completely.

### How should it work?

_No response_

nohwnd:
Testing how we could measure CodeCoverage of Pester and P tests.  The code above runs, except for one test in mock, which is strange since we can Profile pester tests. Maybe it was some fluke. 

- [x] The coverage shows 28% which seems too low, most of our tests run in process, so we maybe are resetting something somewhere. Or not pointing the code at the right files.
- [x] This also needs to be made compatible with inlining the module.
- [x] And Frode pointed out that tracer might not be both the same type.
- [x] we also need finally to tear down
- [x] and print summary
- [x] We are probably not setting that one extra BP that we should be seetting, if all breakpoints are removed Set-Trace will turn off, check profiler what it does, so we can re-enable the test in Mock that removes just it's breakpoints.  https://github.com/nohwnd/Profiler/issues/67

- [x] ~Coverage tests fail but it is imho okay to skip~ these were mock tests, solved
- [x] coverage.ts tests break the coverage because they turn off tracing
- [x] this also higlights that when we profile Pester using PRofiler we are most likely not getting trace of all tests, so there are probably some other Perf wins hiding

image

Fix #2475

### DIFF
--- a/src/functions/Coverage.ps1
+++ b/src/functions/Coverage.ps1
@@ -1152,6 +1152,9 @@ function Stop-TraceScript {
     else {
         # Stop tracing so we don't record Unregister
         $corruptionAutodetectionVariable = Set-PSDebug -Trace 0
+        # This variable name is used to detect if the tracer was broken, we cannot use comments because they are not part of the ast Extent.Text
+        # Assigning it here to null, because otherwise it gives warning about not being used.
+        $null = $corruptionAutodetectionVariable
         # detect if profiler is imported, if yes, unregister us from Profiler (because we are profiling Pester)
         $profilerType = "Profiler.Tracer" -as [Type]
         if ($null -ne $profilerType) {
--- a/src/functions/Coverage.ps1
+++ b/src/functions/Coverage.ps1
@@ -1124,7 +1124,7 @@ function Start-TraceScript ($Breakpoints) {
     if (-not $registered) {
 
         # detect if code coverage is enabled throuh Pester tracer, and in that case just add us as a second tracer
-        if ([Pester.Tracing.Tracer]::ShouldRegisterTracer($tracer, <# overwrite: #> $false)) {
+        if (1 -eq $env:PESTER_CC_IN_CC -and [Pester.Tracing.Tracer]::ShouldRegisterTracer($tracer, <# overwrite: #> $false)) {
             $patched = $false
             $registered = $true
             [Pester.Tracing.Tracer]::Register($tracer)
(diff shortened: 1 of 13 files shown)


src/functions/Coverage.ps1
```diff
@@ -1122,9 +1122,18 @@ function Start-TraceScript ($Breakpoints) {
...
-        Set-PSDebug -Trace 1
+
+        # detect if code coverage is enabled throuh Pester tracer, and in that case just add us as a second tracer
+        if ([Pester.Tracing.Tracer]::ShouldRegisterTracer($tracer, <# overwrite: #> $false)) {
+            $patched = $false
+            $registered = $true
+            [Pester.Tracing.Tracer]::Register($tracer)
+        }
+        else {
+            $patched = $true
```
nohwnd, line 1133: > so I just needed to check in patch and unpatch not just for profiler tracer, but also for pester tracer.

@fflaten  this is what I meant by that. I should hide this under a Env variable probably, so we don't get into broken state after cancellation, as you pointed out.

nohwnd: See few lines of the unchanged code above, it then makes more sense imho.

nohwnd:
The error detection is sweet. Took me half evening yesterday to figure out where the code was stopping to properly trace, and this does it automatically: 

image

nohwnd:
test.ps1 -cc

image

nohwnd:
With inline. 

image

Not sure what is the difference, but it is small enough to ignore.

nohwnd:
shows summary in UI now as well, it is not super helpful, but it is there. 

image

johlju:
Fyi. The task `PublishCodeCoverageResults@2` does not (yet?) support line coverage, but if using v1 the `Pester.psm1` will be clickable and show coverage per line.

nohwnd:
I tried that and I still don't see the uncovered lines in the breakdown. 
image

@fhnaseer is there a way to make the new code coverage view (two posts above) more detailed than just summary per file? we are producing jacoco output.

johlju:
See example here where you can see it show hits and misses per line: https://dev.azure.com/dsccommunity/SqlServerDsc/_build/results?buildId=8834&view=codecoverage-tab

Issue #2477, opened 2024-05-20
replace `Get-EquivalencyOption` with parameters


nohwnd:
`Get-EquivalencyOption` is not exposed, so unable to customize/no docs on expected hashtable.

Just curious, why opt for object/hashtable vs parameters for this? Parameters are easier to document and splatting solves reusable parameters. If sticking with options-object, should it be typed?

_Originally posted by @fflaten in https://github.com/pester/Pester/pull/2428#discussion_r1606098213_

nohwnd:
Exported.

nohwnd:
get-equivalent needs more care, there is not support for Because. 

Strict Order is not exposed and might not even work.

PR #2488, opened 2024-06-09
Fix console warnings in Should-BeEquivalent.Tests.ps1

fflaten:
## PR Summary
Removes "Getting default equivalency options" warning spam in `tests.ps1` ouptut. Fixed by providing default options by... well, default in `Should-BeEquivalent.Tests.ps1`.

_Also replaces some calls to `New-Object`_

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/assert/Equivalence/Should-BeEquivalent.ps1
+++ b/src/functions/assert/Equivalence/Should-BeEquivalent.ps1
@@ -1,4 +1,4 @@
-﻿function Test-Same ($Expected, $Actual) {
+function Test-Same ($Expected, $Actual) {
     [object]::ReferenceEquals($Expected, $Actual)
 }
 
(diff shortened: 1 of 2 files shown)


nohwnd:
Clever. I would not think about doing it this way. Can the warning now be removed from the assertion and throw when the options are not provided?

fflaten:
Yes

nohwnd:
I won't, too many other things to fix before psconf :D

fflaten:
Changed to only check if parameter was not provided (allows null). PSDefaultParameters-trick didn't work when `Compare-Equivalent`, even if I set the option for e.g. `Compare-DictionaryEquivalent`.

Didn't bother to troubleshooting further after title change issue #2477 🙂

PR #2497, opened 2024-06-12
Fix security and support

### LINKED ISSUE #2494: Update SECURITY.md and SUPPORT.md?
@nohwnd Just noticed the Pester supported versions information in [SECURITY.md](https://github.com/pester/Pester/blob/main/SECURITY.md) has not been updated for many years.

_Originally posted by @xtqqczze in https://github.com/pester/Pester/issues/2324#issuecomment-2150578380_

Update SECURITY.md. Should the support matrix also move to SUPPORT.md?

nohwnd:
Fix #2494

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,21 +2,20 @@
 
 ## Supported Versions
 
-Any issues with Pester are generally fixed in the latest releases.  There has been no specific security patches to this date.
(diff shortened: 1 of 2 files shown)


fflaten:
Do we still support v4?

Otherwise LGTM

nohwnd:
Whole powershell repo is on it afaik. so I would keep it for security fixes. Re-releasing would require me to figure out the whole pipeline again probably and the certificate would change, but so far there was no need.

PR #2500, opened 2024-06-13
Add option to disable v5 syntax

### LINKED ISSUE #2495: Option to disable old assertions
### Checklist

When migrating to Assert assertions I want to disable them so I know that I did not migrate some of them. 

When I already migrated, I want to keep the old assertions disabled, and they should fail with informative message when used.

### How should it work?

Pester option is checked in all Shoud assertions and makes them fail when the option is disabled.

- [x] add option `$pesterConfig.should.DisableV5` (default false)
- [x] write error from `Should`
- [ ] write error from `Add-ShouldOperator Get-ShouldOperator` - skipping this is hard to do we don't have the settings at that point, and there is a lot of logic around it.

nohwnd:
Add option to disable Should -Be and other assertions shipped with Pester v5, so when you migrate you don't accidentally regress.

Fix #2495

### DIFF
--- a/src/functions/assertions/Should.ps1
+++ b/src/functions/assertions/Should.ps1
@@ -122,18 +122,23 @@ function Should {
             $shouldThrow = 'Stop' -eq $PSBoundParameters["ErrorAction"]
         }
 
+        # first check if we are in the context of Pester, if not we will always throw, and won't disable Should:
+        # This check is slightly hacky, here we are reaching out the the caller session state and
+        # look for $______parameters which we know we are using inside of the Pester runtime to
+        # keep the current invocation context, when we find it, we are able to add non-terminating
+        # errors without throwing and terminating the test.
+        $pesterRuntimeInvocationContext = $PSCmdlet.SessionState.PSVariable.GetValue('______parameters')
+        $isInsidePesterRuntime = $null -ne $pesterRuntimeInvocationContext
+        $pesterRuntimeInvocationContext.Configuration.Should.ErrorAction.Value
+
+        if ($isInsidePesterRuntime -and $pesterRuntimeInvocationContext.Configuration.Should.DisableV5.Value) {
+            throw "Pester Should -Be syntax is disabled. Use Should-Be (without space), or enable it by setting: `$PesterPreference.Should.DisableV5 = `$false"
+        }
+
         if ($null -eq $shouldThrow -or -not $shouldThrow) {
             # we are sure that we either:
             #    - should not throw because of explicit ErrorAction, and need to figure out a place where to collect the error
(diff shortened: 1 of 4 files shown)


src/functions/assertions/Should.ps1
```diff
@@ -122,18 +122,23 @@ function Should {
             $shouldThrow = 'Stop' -eq $PSBoundParameters["ErrorAction"]
         }
 
+        # first check if we are in the context of Pester, if not we will always throw, and won't disable Should:
+        # This check is slightly hacky, here we are reaching out the the caller session state and
```
fflaten, line 126:
```suggestion
        # This check is slightly hacky, here we are reaching out the caller session state and
```

src/functions/assertions/Should.ps1
```diff
@@ -122,18 +122,23 @@ function Should {
...
+        # look for $______parameters which we know we are using inside of the Pester runtime to
+        # keep the current invocation context, when we find it, we are able to add non-terminating
+        # errors without throwing and terminating the test.
+        $pesterRuntimeInvocationContext = $PSCmdlet.SessionState.PSVariable.GetValue('______parameters')
+        $isInsidePesterRuntime = $null -ne $pesterRuntimeInvocationContext
+        $pesterRuntimeInvocationContext.Configuration.Should.ErrorAction.Value
```
fflaten, line 132: leftover?

src/functions/assertions/Should.ps1
```diff
@@ -122,18 +122,23 @@ function Should {
...
+
+        if ($isInsidePesterRuntime -and $pesterRuntimeInvocationContext.Configuration.Should.DisableV5.Value) {
```
fflaten, line 134: Need to ignore `Invoke`, `InvokeVerifiable` and `HaveParameter` for now. Or delay merge for now.

src/functions/assertions/Should.ps1
nohwnd: yeah, removed

src/functions/assertions/Should.ps1
nohwnd: Why? What am I overlooking?

nohwnd: My logic was that they are all registered via `Should` function, so failing inside of it should fail for all registered assertions operators.

nohwnd:
I am stupid, why is this test writing error to the pipeline ???

src/functions/assertions/Should.ps1
fflaten: Won't this also throw on e.g. `Should -Invoke`? We don't have a replacement atm.

nohwnd:
It must be the CI format, because it formats the output to be recognized by azdo.

PR #2511, opened 2024-06-25
Fix a bug where the NUnit-Report fails to generate if the test output contains Virtual Terminal Sequences

ocalvo:
## PR Summary

The following test will fail to generate an NUnit report:

```
Describe 'Describe VT Sequences' {
  Write-Output "`e[32mHello World`e[0m"
  $true | Should -Be $true
}
```

This will fail because WriteCData does not support writting the ESC character, which is a valid [VT sequence](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences).
The fix is to replace the ESC and Bell characters for {ESC} and {BELL}. 
According to the [Xml CDATA spec](https://www.w3.org/TR/REC-xml/#NT-Char), there is no way to escape these characters, other than failing, but we still need to represent them in the NUnit report.
Afte this change, these characters will be represented as by its printable version in the unicode table: `␛` for the `ESC` (0x1B) and as `␇` for the `BELL` (0x07) in the resulting Xml CData for example.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/TestResults.NUnit3.ps1
+++ b/src/functions/TestResults.NUnit3.ps1
@@ -541,7 +541,18 @@ function Write-NUnit3OutputElement ($Output, [System.Xml.XmlWriter] $XmlWriter) 
         if ($null -eq $o) {
             [string]::Empty
         } else {
-            $o.ToString()
+            $result = ""
+            0..($o.Length-1) |% {
+                $char = $o[$_]
+                if ("`e" -eq $char) {
+                    $result += "{ESC}"
+                } elseif ("`a" -eq $char) {
+                    $result += "{BELL}"
+                } else {
+                    $result += $char
+                }
+            }
+            $result
         }
     }) -join [System.Environment]::NewLine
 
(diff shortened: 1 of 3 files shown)


nohwnd:
Hello, thanks for the PR, the test you added is failing. I assume you also want this backported to version 5.x.x ?

ocalvo:
> Hello, thanks for the PR, the test you added is failing. I assume you also want this backported to version 5.x.x ?

No, we can easily update our test suite to latest version.

ocalvo:
@nohwnd I have fixed the test, it was failling because WindowsPowerShell does not escape the ESC character. Fixed using the ASCII value explicitly.

src/functions/TestResults.NUnit3.ps1
```diff
@@ -537,11 +537,27 @@ function Write-NUnit3TestCaseAttributes {
...
+    # which is obtained by adding 0x2400
+    [int]$unicodeControlPictures = 0x2400
+    [int[]]$validChars = (0x09,0x0A,0x0D)
     $outputString = @(foreach ($o in $Output) {
         if ($null -eq $o) {
             [string]::Empty
         } else {
-            $o.ToString()
+            $result = ""
+            0..($o.Length-1) |% {
```
nohwnd, line 551: Could you try to replace this with regex `-replace` or similar code? This seems quite inefficient re-building the string char by char.

ocalvo:
Hi @nohwnd , we are blocked by this, we would appreciate if you could review at your earliest convenience.

PR #2535, opened 2024-07-08
Exclude duplicate files for Run.Path and CodeCoverage.Path

### LINKED ISSUE #2530: Deduplicate files provided using Run.Path and -Path?
> > Using "/tests", "/tests/.github" should be enough. You only need to specify the hidden folder, the rest will be found using recursion unless you want to include only a subset of folders under tests.
>
> I tried this initially but the `.github` tests ran twice on windows which is also not ideal.

_Originally posted by @tbutler-qontigo in https://github.com/pester/Pester/issues/2515#issuecomment-2202080286_

Should Pester deduplicate file paths provided using `-Path` / `Run.Path`? I don't see a reason why you'd want to run the same file multiple times without parameters.

Calling a file multiple times with different parameters (`-Data`) requires use of `-Container` / `Run.Container` which is unaffected by the suggested change to `-Path`.

fflaten:
## PR Summary
Updates `Find-File` and `Get-CodeCoverageFilePaths` internal functions to exclude duplicate files when provided overlapping `Path` values to search in. This affects:
- Test file paths provided using `-Path / Run.Path`
- Code Coverage source file paths provided using `-CodeCoverage / CodeCoverage.Path`
- Containers created using `New-PesterContainer -Path ..`
  - Mulitple containers can still call the same file, e.g. to invoke it with different `Data` (parameters)

PR also:
- Refactors `Get-CodeCoverageFilePaths` to use `Get-ChildItem -File -Recurse` for increased performance.
- Refactors `New-PesterContainer` to avoid multiple filesystem searches when provided array of Data for increased performance.

Fix #2530

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/functions/Coverage.ps1
+++ b/src/functions/Coverage.ps1
@@ -211,7 +211,7 @@ function Resolve-CoverageInfo {
 
     $filePaths = Get-CodeCoverageFilePaths -Paths $resolvedPaths -IncludeTests $includeTests -RecursePaths $recursePaths
 
-    $params = @{
+    $commonParams = @{
         StartLine = $UnresolvedCoverageInfo.StartLine
         EndLine   = $UnresolvedCoverageInfo.EndLine
         Class     = $UnresolvedCoverageInfo.Class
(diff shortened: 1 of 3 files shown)


src/functions/Coverage.ps1
```diff
@@ -211,54 +211,35 @@ function Resolve-CoverageInfo {
...
-            # todo: enable this warning for non wildcarded paths? otherwise it prints a ton of warnings for documentation and so on when using "folder/*" wildcard
-            # & $SafeCommands['Write-Warning'] "CodeCoverage path '$path' resolved to a non-PowerShell file '$($item.FullName)'; this path will not be part of the coverage report."
+    [string[]] $filteredFiles = @(foreach ($file in (& $SafeCommands['Get-ChildItem'] -LiteralPath $Paths -File -Recurse:$RecursePaths)) {
+        if (('.ps1', '.psm1') -contains $file.Extension -and ($IncludeTests -or $file.Name -notlike $testsPattern)) {
+            $file.FullName
         }
-    }
-
-    return $filePaths
+    })
```
fflaten, line 239: Include or revert refactor? AFAIK it's the same behavior but faster and cleaner.
If yes, ignore in v5 backport to play safe?

src/functions/Coverage.ps1
```diff
@@ -211,54 +211,35 @@ function Resolve-CoverageInfo {
...
-                elseif (-not $i.PSIsContainer) {
-                    $i.PSPath
-                }
-            }
-            Get-CodeCoverageFilePaths -Paths $children -IncludeTests $IncludeTests -RecursePaths $RecursePaths
-        }
-        elseif (-not $item.PsIsContainer) {
-            # todo: enable this warning for non wildcarded paths? otherwise it prints a ton of warnings for documentation and so on when using "folder/*" wildcard
-            # & $SafeCommands['Write-Warning'] "CodeCoverage path '$path' resolved to a non-PowerShell file '$($item.FullName)'; this path will not be part of the coverage report."
+    [string[]] $filteredFiles = @(foreach ($file in (& $SafeCommands['Get-ChildItem'] -LiteralPath $Paths -File -Recurse:$RecursePaths)) {
```
fflaten, line 235: Use of `-LiteralPath` is key here as `Get-ChildItem -Path c:\test.ps1 -Recurse -File` will recursively look for `test.ps1` while `-LiteralPath` avoids this.

nohwnd: I don't think I understand what the difference is. I experimented with it and it returned my child files in both cases.

src/functions/Coverage.ps1
nohwnd: Let's avoid if you think it might not be safe.

src/functions/Coverage.ps1
fflaten: image

src/functions/Coverage.ps1
fflaten: I'm fine with including in backport unless we find an issue. Free perf and easy revert in a patch release + same logic so it's easier to support both. I just thought you'd might want to avoid too many refactors in v5 so leaving decision to you 🙂

src/functions/Coverage.ps1
nohwnd: I get it now.

src/functions/Coverage.ps1
nohwnd: I've kicked off the backport, at least it will unblock OP. And I will have something to release in v5 to try the releasing from server (once I port it back :"D)

PR #2547, opened 2024-07-14
WIP - Block original command by default when parameterized mocks exist

### LINKED ISSUE #2166: Make `Mock` learn `Assert` parameter
As a contributor or maintainer of a project I'd like to be able to help other contributors to correctly mock functions by failing a test if a mock does not exist. This would be very useful in situations when there are very complex scripts that handle a lot of scenarios and using a lot of external helper functions.

The ask would be to add a new parameter to `Mock` that set up a default mock that fails the test if no other mock have overridden it. The paramete could be `Assert`, `MustMock` or `MustBeOverridden` (or any other name that is better).

```powershell
Mock -Assert -CommandName Get-Something2
```

Take this example as a "complex script", assume a contributor added the second `Context`-block for a new scenario. The contributor missed to add a mock for the function `Get-Something2`,  and the maintainer missed it in the review. The script luckily works, but when the mock was missed the test runs much slower than everyone expects, and does not easily show in the pipeline.

```powershell
BeforeAll {
    function Get-Something1
    {
        Start-Sleep -Seconds 3
    }

    function Get-Something2
    {
        Start-Sleep -Seconds 3
    }


    function BigScriptWithALotOfMocks
    {
        Get-Something1
        Get-Something2

        return $true
    }
}

Describe 'BigScriptWithALotOfMocks' {
    Context 'When testing scenario 1' {
        BeforeAll {
            Mock -CommandName Get-Something1
            Mock -CommandName Get-Something2
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }

    Context 'When testing scenario 2' {
        BeforeAll {
            Mock -CommandName Get-Something1
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }
}
```

This result in test that runs "slow" (3.44 seconds):

```
Discovery found 2 tests in 409ms.
Running tests.
[+] C:\Users\johlju\OneDrive\Desktop\PesterExampleMustMock.Tests.ps1 3.44s (3.13s|178ms)
Tests completed in 3.45s
Tests Passed: 2, Failed: 0, Skipped: 0 NotRun: 0
```

We could help prevent this today by adding a default mock for each command that needs to be mocked by contributors under the `Describe`-block. The default mock throws an error.

```powershell
BeforeAll {
    function Get-Something1
    {
        Start-Sleep -Seconds 3
    }

    function Get-Something2
    {
        Start-Sleep -Seconds 3
    }


    function BigScriptWithALotOfMocks
    {
        Get-Something1
        Get-Something2

        return $true
    }
}

Describe 'BigScriptWithALotOfMocks' {
    BeforeAll {
        Mock -CommandName Get-Something1 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }

        Mock -CommandName Get-Something2 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }
    }

    Context 'When testing scenario 1' {
        BeforeAll {
            Mock -CommandName Get-Something1
            Mock -CommandName Get-Something2
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }

    Context 'When testing scenario 2' {
        BeforeAll {
            Mock -CommandName Get-Something1
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }
}
```

This result in:

```
Discovery found 2 tests in 931ms.
Running tests.
[-] BigScriptWithALotOfMocks.When testing scenario 2.Should return $true 145ms (143ms|3ms)
 RuntimeException: The default mock for 'Get-Something1' was not overridden.
 at <ScriptBlock>, C:\Users\johlju\Desktop\PesterExampleMustMock.Tests.ps1:29
```

The contributor adds the missing mock and the tests passes and runs as quick one would expect.

```
BeforeAll {
    function Get-Something1
    {
        Start-Sleep -Seconds 3
    }

    function Get-Something2
    {
        Start-Sleep -Seconds 3
    }


    function BigScriptWithALotOfMocks
    {
        Get-Something1
        Get-Something2

        return $true
    }
}

Describe 'BigScriptWithALotOfMocks' {
    BeforeAll {
        Mock -CommandName Get-Something1 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }

        Mock -CommandName Get-Something2 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }
    }

    Context 'When testing scenario 1' {
        BeforeAll {
            Mock -CommandName Get-Something1
            Mock -CommandName Get-Something2
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }

    Context 'When testing scenario 2' {
        BeforeAll {
            Mock -CommandName Get-Something1
            Mock -CommandName Get-Something2
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }
}
```

```
Discovery found 2 tests in 499ms.
Running tests.
[+] C:\Users\johlju\OneDrive\Desktop\PesterExampleMustMock.Tests.ps1 906ms (456ms|242ms)
Tests completed in 1.01s
Tests Passed: 2, Failed: 0, Skipped: 0 NotRun: 0
```

Instead of setting a default mock that throws I would like to see this works natively with a command.
So instead of this:
```
    BeforeAll {
        Mock -CommandName Get-Something1 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }

        Mock -CommandName Get-Something2 -MockWith {
            throw "The default mock for 'Get-Something1' was not overridden."
        }
    }
```

This would look nicer and be simple:

```
    BeforeAll {
        Mock -Assert -CommandName Get-Something1
        Mock -Assert -CommandName Get-Something2
    }
```

### LINKED ISSUE #2178: [Pester 5] Original function silently executed if parameter validation fails in mock
When mock function fails parameter validation, the original function is executed with no warnings/errors reported.

## Describe your environment

Pester version     : 5.3.0 F:\Users\bozho\.powershell\modules\Pester\5.3.0\Pester.psm1
PowerShell version : 5.1.19041.1645
OS version         : Microsoft Windows NT 10.0.19044.0

## Steps to reproduce

I have a module function that looks something like this:
```powershell
function Get-SomeRestResource {
    [CmdletBinding(PositionalBinding = $false)]
    param(
        [parameter()]
        [ValidateNotNull()]
        [string] $Server = $script:DefaultServer
    )

    try {
        $headers = @{
            Authorization = "...."
            "Content-Type" = "application/json"
        }

        $response = Invoke-RestMethod -Uri "https://$Server/api/some/resource" -Method Get -Headers $headers
    }
    catch {
        ....
    }
}
```

`$script:DefaultServer` is a module variable.

I am mocking both `Get-SomeRestResource` and `Invoke-RestMethod`. `Invoke-RestMethod` simply returns a constant JSON object (doesn't reference `$Server` parameter at all).

I have a test where I simply execute `Get-SomeRestResource` with no parameters, so `$Server` is supposed to pick up the default server string.

I had a bug where `$script:DefaultServer` module variable was not defined (I forgot to initialise it in the module), which resulted in `$Server` parameter being an empty string. This in turn resulted in the original `Get-SomeRestResource` being executed - I've confirmed that both using VS Code's test debugging and the fact that the original `Invoke-RestMethod` was complaining about invalid URL (`https:///api/some/resource`).

The only change I had to make is to actually declare `$script:DefaultServer` module variable, with no other changes to the module or test code and everything worked as expected.

## Expected Behavior

Parameter validation error being raised.

fflaten:
## PR Summary
**WIP**

When a command is mocked by only parameterized mocks, throw for calls not matching any filters unless `-AllowFallback` is specified.

Fix #2166
Fix #2178

TODO
- [ ] Discuss desired behavior for precedence and multiple parameterized mocks etc. See review comments
- [ ] Discuss API
  - [ ] Do we need a global setting to disable for v5 backwards compatibility?
  - [ ] Change parameter and property name? `-AllowFallback` could reference fallback to default mock in module or script-scope
  - [ ] Is this the right approach? Having to add `-AllowFallback` is weird with multiple parameterized mocks in scope. Better to introduce `Mock -Throw` shortcut which can be used for both default mock and special parameterized tests?
- [x] Troubleshoot seemingly unaffected but failing test `Mocking with nested Pester runs.Mocking works in nested run`
- [ ] Update error thrown when original command is blocked. Currently a placeholder.

## PR Checklist

- [ ] PR has meaningful title

### DIFF
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -28,20 +28,22 @@ function New-MockBehavior {
         [Parameter(Mandatory)]
         $Hook,
         [string[]]$RemoveParameterType,
-        [string[]]$RemoveParameterValidation
+        [string[]]$RemoveParameterValidation,
+        [switch]$AllowFallback
     )
 
     [PSCustomObject] @{
-        CommandName = $ContextInfo.Command.Name
-        ModuleName  = $ContextInfo.TargetModule
-        Filter      = $ParameterFilter
-        IsDefault   = $null -eq $ParameterFilter
-        IsInModule  = -not [string]::IsNullOrEmpty($ContextInfo.TargetModule)
-        Verifiable  = $Verifiable
-        Executed    = $false
-        ScriptBlock = $MockWith
-        Hook        = $Hook
-        PSTypeName  = 'MockBehavior'
+        CommandName   = $ContextInfo.Command.Name
+        ModuleName    = $ContextInfo.TargetModule
+        Filter        = $ParameterFilter
+        AllowFallback = $AllowFallback
+        IsDefault     = $null -eq $ParameterFilter
+        IsInModule    = -not [string]::IsNullOrEmpty($ContextInfo.TargetModule)
+        Verifiable    = $Verifiable
+        Executed      = $false
+        ScriptBlock   = $MockWith
+        Hook          = $Hook
+        PSTypeName    = 'MockBehavior'
     }
 }
 
@@ -982,6 +984,10 @@ function FindMatchingBehavior {
 
     $foundDefaultBehavior = $false
     $defaultBehavior = $null
+    # A mock exists so we do not allow original command to be called. Either:
+    # - A parameterized mock explicitly allows it below
+    # - Or a default mock exists which takes precedence either way
+    $fallbackAllowed = $false
     foreach ($b in $Behaviors) {
 
         if ($b.IsDefault -and -not $foundDefaultBehavior) {
@@ -1005,6 +1011,8 @@ function FindMatchingBehavior {
                 }
                 return $b
             }
+
+            if ($b.AllowFallback) { $fallbackAllowed = $true }
         }
     }
 
@@ -1018,6 +1026,11 @@ function FindMatchingBehavior {
     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
         Write-PesterDebugMessage -Scope Mock "No parametrized or default behaviors were found filter."
     }
+
+    if (-not $fallbackAllowed) {
+        throw 'Not implemented. None of the parameterized mocks match the call or accept fallback to original command.'
+    }
+
     return $null
 }
 
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -984,10 +984,8 @@ function FindMatchingBehavior {
 
     $foundDefaultBehavior = $false
     $defaultBehavior = $null
-    # A mock exists so we do not allow original command to be called. Either:
-    # - A parameterized mock explicitly allows it below
-    # - Or a default mock exists which takes precedence either way
-    $fallbackAllowed = $false
+    # Allow fallback unless parameterized behaviors exist. Required to support nested runs where mock exists but no behaviors.
+    $blockOriginalCommand = @($Behaviors).IsDefault -contains $false
     foreach ($b in $Behaviors) {
 
         if ($b.IsDefault -and -not $foundDefaultBehavior) {
@@ -1012,7 +1010,7 @@ function FindMatchingBehavior {
                 return $b
             }
 
-            if ($b.AllowFallback) { $fallbackAllowed = $true }
+            if ($b.AllowFallback) { $blockOriginalCommand = $false }
         }
     }
 
@@ -1027,7 +1025,7 @@ function FindMatchingBehavior {
         Write-PesterDebugMessage -Scope Mock "No parametrized or default behaviors were found filter."
     }
 
-    if (-not $fallbackAllowed) {
+    if ($blockOriginalCommand) {
         throw 'Not implemented. None of the parameterized mocks match the call or accept fallback to original command.'
     }
 
--- a/src/functions/Pester.SessionState.Mock.ps1
+++ b/src/functions/Pester.SessionState.Mock.ps1
@@ -124,6 +124,10 @@ function Mock {
     validation requirements, and allows functions that are strict about their
     parameter validation to be mocked more easily.
 
+    .PARAMETER AllowFallback
+    Allows original command to execute when only mocks with -ParameterFilter exists and the call does not match any of them.
+    Creating a default mock (no -ParameterFilter specified) will take precedence over this.
+
     .EXAMPLE
     Mock Get-ChildItem { return @{FullName = "A_File.TXT"} }
 
@@ -219,15 +223,22 @@ function Mock {
     .LINK
     https://pester.dev/docs/usage/mocking
     #>
-    [CmdletBinding()]
+    [CmdletBinding(DefaultParameterSetName = 'Default')]
+    # TODO: Breaking change due to parameter sets. Previously had positions:
+    # CommandName, MockWith, ParameterFilter, ModuleName, RemoveParameterType, RemoveParameterValidation
     param(
+        [Parameter(Position = 0)]
         [string]$CommandName,
+        [Parameter(Position = 1)]
         [ScriptBlock]$MockWith = {},
         [switch]$Verifiable,
+        [Parameter(ParameterSetName = 'WithParameterFilter')]
         [ScriptBlock]$ParameterFilter,
         [string]$ModuleName,
         [string[]]$RemoveParameterType,
-        [string[]]$RemoveParameterValidation
+        [string[]]$RemoveParameterValidation,
+        [Parameter(ParameterSetName = 'WithParameterFilter')]
+        [switch]$AllowFallback
     )
     if (Is-Discovery) {
         # this is to allow mocks in between Describe and It which is discouraged but common
@@ -280,7 +291,7 @@ function Mock {
         $mockData.Behaviors[$contextInfo.Command.Name] = $behaviors
     }
 
-    $behavior = New-MockBehavior -ContextInfo $contextInfo -MockWith $MockWith -Verifiable:$Verifiable -ParameterFilter $ParameterFilter -Hook $hook
+    $behavior = New-MockBehavior -ContextInfo $contextInfo -MockWith $MockWith -Verifiable:$Verifiable -ParameterFilter $ParameterFilter -Hook $hook -AllowFallback:$AllowFallback
     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
         Write-PesterDebugMessage -Scope Mock -Message "Adding a new $(if ($behavior.IsDefault) {"default"} else {"parametrized"}) behavior to $(if ($behavior.ModuleName) { "$($behavior.ModuleName) - "})$($behavior.CommandName)."
     }
@@ -1045,7 +1056,7 @@ function Invoke-Mock {
             param ($b)
             "     Target module: $(if ($b.ModuleName) { $b.ModuleName } else { '$null' })`n"
             "    Body: { $($b.ScriptBlock.ToString().Trim()) }`n"
-            "    Filter: $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" } else { '$null' })`n"
+            "    Filter$(if ($b.AllowFallback) { ' (with fallback)' }): $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" }) else { '$null' })`n"
             "    Default: $(if ($b.IsDefault) { '$true' } else { '$false' })`n"
             "    Verifiable: $(if ($b.Verifiable) { '$true' } else { '$false' })"
         }
--- a/src/functions/Pester.SessionState.Mock.ps1
+++ b/src/functions/Pester.SessionState.Mock.ps1
@@ -1056,7 +1056,7 @@ function Invoke-Mock {
             param ($b)
             "     Target module: $(if ($b.ModuleName) { $b.ModuleName } else { '$null' })`n"
             "    Body: { $($b.ScriptBlock.ToString().Trim()) }`n"
-            "    Filter$(if ($b.AllowFallback) { ' (with fallback)' }): $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" }) else { '$null' })`n"
+            "    Filter$(if ($b.AllowFallback) { ' (with fallback)' }): $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" } else { '$null' })`n"
             "    Default: $(if ($b.IsDefault) { '$true' } else { '$false' })`n"
             "    Verifiable: $(if ($b.Verifiable) { '$true' } else { '$false' })"
         }
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -2798,6 +2798,66 @@ Describe 'Mocking using ParameterFilter' {
     }
 }
 
+Describe 'Calls not matching ParameterFilter' {
+    BeforeAll {
+        function demo ($name) {
+            "hello $name"
+        }
+    }
+
+    Context 'When default mock does not exist in parent scope' {
+        It 'Throws by default' {
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+            { demo -name 'you' } | Should -Throw
+        }
+
+        It 'Calls locally defined default mock' {
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+            Mock demo { 'default mock' }
+            demo -name 'you' | Should -Be 'default mock'
+        }
+
+        It 'Calls real function when -AllowFallback is used' {
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
+            demo -name 'you' | Should -Be 'hello you'
+        }
+    }
+
+    Context 'When default mock exists in parent scope' {
+        BeforeAll {
+            Mock demo -MockWith { 'default mock' }
+        }
+        It 'Calls default mock without -AllowFallback' {
+            # TODO: Is this expected? Should -AllowFallback only have effect when there's no default mock at any level?
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+            demo -name 'you' | Should -Be 'default mock'
+        }
+    }
+
+    Context 'When mock with ParameterFilter exists in parent scope' {
+        BeforeAll {
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+        }
+        It 'Throws by default' {
+            { demo -name 'you' } | Should -Throw
+        }
+
+        It 'Calls locally defined default mock' {
+            Mock demo { 'default mock' }
+            demo -name 'you' | Should -Be 'default mock'
+        }
+    }
+
+    Context 'When mock with ParameterFilter and -AllowFallback exists in parent scope' {
+        BeforeAll {
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
+        }
+        It 'Calls real function' {
+            demo -name 'you' | Should -Be 'hello you'
+        }
+    }
+}
+
 
 if ($PSVersionTable.PSVersion.Major -ge 3) {
     Describe "-RemoveParameterType" {
--- a/tst/functions/Mock.Tests.ps1
+++ b/tst/functions/Mock.Tests.ps1
@@ -374,7 +374,7 @@ Describe 'When calling Mock, StrictMode is enabled, and variables are used in th
 
 Describe "When calling Mock on existing function without matching bound params" {
     It "Should redirect to real function" {
-        Mock FunctionUnderTest { return "fake results" } -parameterFilter { $param1 -eq "test" }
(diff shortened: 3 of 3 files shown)


tst/functions/Mock.Tests.ps1
```diff
@@ -2798,6 +2798,79 @@ Describe 'Mocking using ParameterFilter' {
...
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
+        }
+        It 'Calls real function' {
+            demo -name 'you' | Should -Be 'hello you'
+        }
+
+        It 'Throws when a more local parameterized mock does not allow fallback' -Skip {
+            # TODO:Do we expect this? If so, we need to expose mock scope depth from Get-AllMockBehaviors
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+            { demo -name 'you' } | Should -Throw
```
fflaten, line 2869: How should this scenario work? Should a local parameterized mock be able to block fallback to original command when a parent mock allows it?

tst/functions/Mock.Tests.ps1
```diff
@@ -2798,6 +2798,79 @@ Describe 'Mocking using ParameterFilter' {
...
+
+    Context 'When default mock exists in parent scope' {
+        BeforeAll {
+            Mock demo -MockWith { 'default mock' }
+        }
+        It 'Calls default mock without -AllowFallback' {
+            # TODO: Is this expected? Should -AllowFallback only have effect when there's no default mock at any level?
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' }
+            demo -name 'you' | Should -Be 'default mock'
+        }
```
fflaten, line 2841: This is likely a common use case where a command returns a default result while some parameterized mocks returns special output. Good as is or any exceptions to this behavior?

tst/functions/Mock.Tests.ps1
```diff
@@ -2798,6 +2798,79 @@ Describe 'Mocking using ParameterFilter' {
...
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
+            demo -name 'you' | Should -Be 'hello you'
+        }
+
+        It 'Calls real function when at least one parameterized mock has -AllowFallback' {
+            # TODO: Is this expected? Or should all parameterized mocks allow fallback?
+            Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
+            Mock demo -ParameterFilter { $name -eq 'Wisconsin' } -MockWith { 'mocked2' }
+            demo -name 'you' | Should -Be 'hello you'
+        }
```
fflaten, line 2830: How should this work? This is where `-AllowFallback/AllowOriginal/AllowPassthrough` gets ugly compared to having a `Mock demo -Throw` shortcut to implement a guard default mock.

src/functions/Mock.ps1
```diff
@@ -28,20 +28,22 @@ function New-MockBehavior {
...
-        IsInModule  = -not [string]::IsNullOrEmpty($ContextInfo.TargetModule)
-        Verifiable  = $Verifiable
-        Executed    = $false
-        ScriptBlock = $MockWith
-        Hook        = $Hook
-        PSTypeName  = 'MockBehavior'
+        CommandName   = $ContextInfo.Command.Name
+        ModuleName    = $ContextInfo.TargetModule
+        Filter        = $ParameterFilter
+        AllowFallback = $AllowFallback
```
fflaten, line 39: Name suggestions for parameter and property? Fallback is also used for "fallback to default mock". `AllowOriginal` could give expectation of blocking stuff like `& (Get-Command Get-ChildItem -CommandType Cmdlet)`. `AllowPassthrough` is unclear. Other short suggestions?

Focus on other review comments first in case this design is discarded.

src/functions/Pester.SessionState.Mock.ps1
```diff
@@ -219,15 +223,22 @@ function Mock {
...
         [ScriptBlock]$MockWith = {},
         [switch]$Verifiable,
+        [Parameter(ParameterSetName = 'WithParameterFilter')]
         [ScriptBlock]$ParameterFilter,
         [string]$ModuleName,
         [string[]]$RemoveParameterType,
-        [string[]]$RemoveParameterValidation
+        [string[]]$RemoveParameterValidation,
+        [Parameter(ParameterSetName = 'WithParameterFilter')]
+        [switch]$AllowFallback
```
fflaten, line 241: Do we want to support all original positional parameters or are `CommandName` + `MockWith` enough? Wouldn't expect the rest defined at line 228 to be commonly used.

src/functions/Pester.SessionState.Mock.ps1
```diff
@@ -1045,7 +1056,7 @@ function Invoke-Mock {
             param ($b)
             "     Target module: $(if ($b.ModuleName) { $b.ModuleName } else { '$null' })`n"
             "    Body: { $($b.ScriptBlock.ToString().Trim()) }`n"
-            "    Filter: $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" } else { '$null' })`n"
+            "    Filter$(if ($b.AllowFallback) { ' (with fallback)' }): $(if (-not $b.IsDefault) { "{ $($b.Filter.ToString().Trim()) }" } else { '$null' })`n"
```
fflaten, line 1059: Should probably update text and/or use separate line.

fflaten (commented):
@nohwnd Experimented a little with this. Not sure if this is the most intuitive direction. See comments.  What do you think?

src/functions/Mock.ps1
```diff
@@ -1018,6 +1026,11 @@ function FindMatchingBehavior {
     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
         Write-PesterDebugMessage -Scope Mock "No parametrized or default behaviors were found filter."
     }
+
+    if (-not $fallbackAllowed) {
+        throw 'Not implemented. None of the parameterized mocks match the call or accept fallback to original command.'
```
fflaten, line 1031: For some reason this fails existing `Mocking with nested Pester runs.Mocking works in nested run` test found here https://github.com/pester/Pester/blob/0d3dd6681883f8db2c183a392952e5be383ba1e0/tst/functions/Mock.Tests.ps1#L3119-L3147

It fails because the inner run's `outer mock is not available` calls `Get-Date` which has an existing mock hock from the outer run with Debug logging enabled, while the inner run does not. Result: `Write-PesterDebugMessage` throws in the condition guard.

It tries to print "Mock bootstrap function Get-Date called from block Begin" which is the first line in the mock hook, way before the changed logic. What am I missing?

nohwnd:
I don't expect the original behavior to be useful to almost anyone, I would add option to configuration for the whole testbase and that is it.

If you really need the fallback only in a specific case you can still add a default mock that calls the original command via & (Get-Command ...).

At least that is my current view, and imho the easiest way to implement this change, and we can add the disabling on a different levels if we get enough good reasons to do that.

fflaten:
Oh... 😅 

I assumed it was common to e.g. mock `Invoke-RestMethod` for an external service you don't control while using real calls for the rest.

Not sure though as I rarely write tests depending on external data.

johlju:
> making users write the pass through mock manually is cumbersome

Yes I agree, this would just add another layer of complexity to an already hard thing to learn for folks and for me as a reviewer to grasp when reading code.

johlju:
This is what I'm looking for:

```powershell
function Get-SomeFile {
    $file = Get-Item -Path 'myFile.txt'
    Get-Content -Path $file.FullName -Raw
}

Describe 'Something' {
    BeforeAll {
        # Guard mock
        Mock Get-Item -Throw
    }

    Context 'When something happens' {
        It 'Should do something' {
            # FAILS: This fails because Get-Item is not mocked
            { Get-SomeFile } | Should-Be 'myFile content'
        }
    }

    Context 'When something else happens' {
        BeforeAll {
            Mock Get-Item -MockWith { @{ FullName = 'myFile.txt' } }
            Mock Get-Content -MockWith { 'myFile content' }
        }

        It 'Should do something' {
            # PASS: This passes because mocks are set up
            { Get-SomeFile } | Should-Be 'myFile content'
        }
    }
}
```

nohwnd:
Okay, thanks for checking my assumptions. I for myself never used the fall through to real command, and only occasionally I've called the real command from a mock when calling Write-Host or similar "utility" command. 

In this case I don't know what to do exactly, putting the allow parameter on all mocks in scope is complicated, and defining a default mock, just to let it call the real command sounds counterintuitive.

fflaten:
> Not sure I'm following what you are discussing, but if it is relevant here is my 2c...

That's the issue with trying to solve two separate issues with one PR 😄 
You'd basically like a built-in shortcut syntax for:
```powershell
# MockHelpers.ps1
function RequireMock {
    [CmdletBinding()]
    param(
        [Parameter(Mandatory)]
        [string] $CommandName,
        [ScriptBlock] $ParameterFilter,
        [string] $ModuleName
    )

    $mockScriptBlock = {
        throw "Missing mock for '$CommandName'."
    }.GetNewClosure()

    Pester\Mock @PSBoundParameters -MockWith $mockScriptBlock
}

# Demo.tests.ps1
BeforeAll {
    . "$PSScriptRoot/MockHelper.ps1"
}
Describe 'BigScriptWithALotOfMocks' {
    BeforeAll {
        RequireMock -CommandName Get-Something1
        RequireMock -CommandName Get-Something2
    }

    Context 'When testing scenario 1' {
        BeforeAll {
            Mock -CommandName Get-Something1
            Mock -CommandName Get-Something2
        }

        It 'Should return $true' {
            BigScriptWithALotOfMocks | Should -BeTrue
        }
    }
}
```

The other request wanted to make this implicit **when** any mock exists for the command. Implicit behavior would be a global setting, but IMO we'd also need a override doing the opposite as your design, which is a little more messy with multiple mocks. E.g.

```powershell
It 'Who wins? Do we allow original command or not?' {
    Mock demo -ParameterFilter { $name -eq 'world' } -MockWith { 'mocked' } -AllowFallback
    Mock demo -ParameterFilter { $name -eq 'Wisconsin' } -MockWith { 'mocked2' }
    demo -name 'you' | Should -Be 'hello you'
}

It 'Or maybe a separate parameter-set to enable it like a flag?' {
    Mock demo -AllowFallback
    Mock demo -ParameterFilter { $name -eq 'Wisconsin' } -MockWith { 'mocked2' }
    demo -name 'you' | Should -Be 'hello you'
}

It 'Or just a shortcut syntax to re-enable passthrough' {
    Mock demo -InvokeOriginal # Default mock will passthrough
    Mock demo -InvokeOriginal -ParameterFilter { $name -eq 'world' } # Passthrough
    Mock demo -ParameterFilter { $name -eq 'Wisconsin' } -MockWith { 'mocked2' }
    demo -name 'you' | Should -Be 'hello you'
}
```

I see the value of both though I'd probably wouldn't use either often. My suggestions:
**Alternative 1 - Support all**
- Add a `Mock.RequireDefaultMock` type of setting (default true? false for v5?)
- Add two shortcuts alternatives to `-MockWith` either direction:
  - `Mock -Throw -CommandName ... -ParameterFilter ...`
  - `Mock -InvokeOriginal -CommandName ... -ParameterFilter ...`

**Alternative 2 - Do nothing, require explicit code**
Update docs and/or publish a utility module to implement `RequireMock` and `InvokeOriginal` helpers.

johlju:
I think I'm following now, you trying to handle two scenarios where the new functionality is either implicitly enabled or disabled.

- Scenario implicitly Enabled:
  - Your suggestion is that we have `AllowFallback`, `InvokeOriginal`, `AllowOriginal` to allow original command to be called when there is one or more mocks for that command.
- Scenario implicitly Disabled:
  - Your suggestion is that we have `Throw` that prevents a command to be called when that command has one or more mocks.

If the above is alternative 1 then I vote for alternative 1. 🙂 

If alternativ 2 demands users to write even complex mocks to be able to call original command then I suggest we avoid that alternative.

fflaten:
> If the above is alternative 1 then I vote for alternative 1. 🙂

Correct

> If alternativ 2 demands users to write even complex mocks to be able to call original command then I suggest we avoid that alternative.

My alt. 2 was to not change the current behavior. So you'd only need to write a mock to call the original command if you already had a manual guard mock defined (e.g. in a root `BeforeAll`). As it's a explicit choice I'd assume you'd might want to place the guard mock in a more specific block instead.

> But would be okay for `Throw` part (`RequireMock` command) in alternative 1 to be in a utility module. Although that would be yet another module to maintain. 🤔

Yeah, publishing it as a module requires a little session state trickery to support all scenarios.

johlju:
The main reason for adding a guard mock in a test I made (and the reason for issue #2166) was so an initial author of a test (me in this example) in the Describe/BeforeAll-block can say "_This command we call must always be mocked in all context-blocks, otherwise bad things will happen to the dev machine_". Other contributors will not end up in the pit because of the guard mock. For that reason the `Mock ... -Throw` could be a solution, but I instead used a workaround by adding a "default mock" (a first mock that throws) that is then overridden by another mock. If that functionality to override an existing mock persist in Pester 6 then maybe it is not worth the effort to add `Throw` if it add complexity.

If we in Pester 6 simplify this PR to:

- default to not allow a call original command if there is a mock for that command
- only way to call the original command is to pass `InvokeOriginal` to the first mock of the command.
- if passing `InvokeOriginal` when there is already a mock present for the command result in error.
- (optionally) a setting to revert to Pester 5 functionality (the above no longer works)

This would be a breaking change with Pester 6 for existing tests unless it possible to disable it.

Then this would also hopefully still be allow to create a default (guard) mock:

```
function MySuperCommand { Do-PreSetup; Do-DevastatingWork }

Describe {
    BeforeAll {
        Mock -CommandName Do-DevastatingWork -MockWith { throw 'This must be mocked' }
    }

    Context 'Test pre setup' {
        It 'should call Do-PreSetup' {
            Mock -CommandName Do-PreSetup
            Mock -CommandName Do-DevastatingWork -MockWith { '1' }

            MySuperCommand
        }
    }

    Context 'Test pre setup' {
        It 'should call Do-PreSetup' {
            Mock -CommandName Do-PreSetup
            Mock -CommandName Do-DevastatingWork -MockWith { '2' }

            MySuperCommand
        }
    }
}

nohwnd:
getting back to this. I think the fallback should apply only to a default mock, and should not apply to parametrized mocks.

Implicitly if a parametrized mock does not match we would fallback to a default mock, that one will either be implicitly specified and will throw, or will be explicitly specified and will call the original command (this can be exposed as option, but personally I'd rather avoid it.


So this leaves these situations:

- mock is not defined at all, we throw
- parametrized mock is defined but it does not match - we fall back to "default" and throw.
- default mock that allows calling the original is defined via Mock MyCommand -FallThrough and we fall to original command
- default mock that throws is defined via Mock MyCommand -Throw and it throws - this is for situation when you want to define a mock that throws when other mocks don't match but you already have a default mock that does something.
- we might define a PesterMockedCommand variable to easily invoke the original command in mock body - this would be available for both parametrized and non-parametrized mocks.

JosephColvin:
Here is an idea... and one that might be valid testing wise.
`
Describe "Using '-InvokeOriginal' switch for both Mock and Should" {
 BeforeAll {
  function demo([string]$Name) { return("Hello $Name") }
  function demo2([string]$Name) { return("Goodbye $Name") }
  function demo3([string]$Name) { return("No more $Name") }
# The default behavior should be throwing a [NotImplemented] exception when a mock command has no defined behavior

  <# demo: We allow the original command to be invoked if name is "You" #>
  Mock demo -ParameterFilter {$Name -eq 'You'} -InvokeOriginal
  <# demo: Mock the command for when Name is Joe #>
  Mock demo -Parameterfilter {$Name -eq 'Joe'} { 'Hi' }
  
  <# demo2: Allow calling the original for any undefined behaviors #>
  Mock demo2 -InvokeOriginal
  <# demo2: Mock the command for when Name is Joe #>
  Mock demo2 -ParameterFilter {$Name -eq 'Joe'} { 'Not again' }

 <# Expected Behavior
   demo will throw an error when the behavior is undefined (name is 'You' or 'Joe')
   demo will invoke the original only when name is 'You'
   demo will invoke the mock only when the name is 'Joe'

   demo2 will not throw an error as the desired behavior is to invoke the original
   demo2 will invoke the mock for only when the name is 'Joe'
   demo2 will invoke the original only when the name is not 'Joe'
#>
 }
 It "Should know if the mocked function was called or not" {
  demo2 -Name 'John' | Should -InvokeOriginal
  demo2 -Name 'Joe' | Should -Not -InvokeOriginal
 }
 It "Should follow mock defined behaviors" {
  demo -Name 'You' | Should -BeExactly 'Hello You'
  demo -Name 'Joe' | Should -BeExactly 'Hi'
  demo2 -Name 'Joe' | Should -BeExactly 'Not again'
  demo2 -Name 'John' | Should -BeExactly 'Goodbye John'
 }
 It "Should still throw an error for undefined behaviors of mocked commands" {
  { demo -Name 'John' } | Should -Throw -Because 'the -InvokeOriginal is only applied to a parameter filter based mocked for demo'
  { demo2 -Name 'You' } | Should -Not -Throw -Because 'the behavior is defined to invoke the original'
 }
`

nohwnd:
Removed the fallback entirely. Let's see what it really blocks in practice.

PR #2572, opened 2024-10-30
[rel/5.x.x] Implement Cobertura coverage format

nohwnd:
Backport #2298 to 5.x.x
Backport #2553 as well

### DIFF
--- a/tst/functions/Coverage.Tests.ps1
+++ b/tst/functions/Coverage.Tests.ps1
@@ -145,7 +145,9 @@ InPesterModuleScope {
                 # Path deliberately duplicated to make sure the code doesn't produce multiple breakpoints for the same commands
                 $breakpoints = Enter-CoverageAnalysis -CodeCoverage $testScriptPath, $testScriptPath, $testScript2Path, $testScript3Path -UseBreakpoints $UseBreakpoints
 
-                @($breakpoints).Count | Should -Be 19 -Because 'it has the proper number of breakpoints defined (using profiler based code coverage: $UseBreakpoints)'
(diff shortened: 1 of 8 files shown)


nohwnd:
regarding your comments on the source PR @fflaten why do you say this one will need to use a different implementation? the api seems to exist in .NET Framework 462: https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimemilliseconds?view=netframework-4.6.2 which is the oldest one we support.

the comments: 
https://github.com/pester/Pester/pull/2298#discussion_r1676502351
https://github.com/pester/Pester/pull/2298#discussion_r1676499300

nohwnd:
Too tired to deal with the build errors right now.

fflaten:
> regarding your comments on the source PR @fflaten why do you say this one will need to use a different implementation? the api seems to exist in .NET Framework 462

Pester v5 targets .NET 4.5.2.

tst/functions/Coverage.Tests.ps1
```diff
@@ -79,11 +86,18 @@ InPesterModuleScope {
...
+                        'I am the constructor of base class.'
+                    }
+                #}
+
                 #class MyClass
                 #{
                     function MyClass
                     {
-                        'I am the constructor.'
+                        MyBaseClass # calls the base constructor
```
nohwnd, line 100: This was originally `MyBaseClass; "I am the constructor"` but that adds one more BP than needed, removed the string so we simulate the call to base constructor like the new class base code does it.

nohwnd: https://github.com/pester/Pester/pull/2577 Forward ported here.

PR #2614, opened 2025-01-22
Remove dead code for Incomplete.

davidmatson:
The block here appeared to me to be unreachable code, but please double-check my analysis.

- [x] PR has meaningful title


fflaten:
It should be dead, but if anything I'd consider throwing as an alternative to removing so we can catch future bugs.

davidmatson:
> It should be dead, but if anything I'd consider throwing as an alternative to removing so we can catch future bugs.

Sure, added a throw.

nohwnd (approved):

PR #2656, opened 2025-07-30
Fix standard output in BeforeAll

### LINKED ISSUE #2655: Pester crashes due to native command output in `BeforeAll`
### Checklist

### What is the issue?

Within `BeforeAll{}` if running a native command and it emits output to STDOUT, it can cause Pester to have an unhandled exception later.

### Expected Behavior

Test should run and Pester should not have unhandled exception

### Steps To Reproduce

In this [test script](https://github.com/PowerShell/DSC/blob/main/powershell-adapter/Tests/win_powershellgroup.tests.ps1#L8) within the `BeforeAll{}` it runs `winrm quickconfig -quiet -force`, despite `-quiet`, the command runs `net start winrm` which itself emits that `WinRM service is already running on this machine.` to STDOUT.  This is captured by Pester and during execution, at this [line](https://github.com/pester/Pester/blob/d08debe801361faf9a03621d2898a1ae35374acb/src/Main.ps1#L817) where it's adding a "container", `$i` happens to be the output string from above and Pester fails with:

```console
System.Management.Automation.MethodException: Cannot find an overload for "Add" and the argument count: "1".
at Invoke-Pester<End>, C:\Users\slee\OneDrive - Microsoft\Documents\PowerShell\Modules\pester\5.7.1\Pester.psm1: line 4985
```

Fix in the test was to redirect the output to `$null`, but this wasn't obvious and was working with previous version of Pester 5.

### Describe your environment

This is with Pester 5.7.1

### Possible Solution?

In the test, redirect the native command output to `$null`

nohwnd:
Tries to fix #2655, but cannot repro.

### DIFF
--- a/tst/Pester.RSpec.ts.ps1
+++ b/tst/Pester.RSpec.ts.ps1
@@ -2784,4 +2784,26 @@ i -PassThru:$PassThru {
             $pwd.Path | Verify-Equal $beforePWD
         }
     }
+
+    b 'Standard output in setup and teardown blocks does not crash tests' {
+        dt 'output in BeforeAll' {
+            $sb = {
+                Describe 'd' {
+                    BeforeAll {
+                        winrm quickconfig -quiet -force
+                        # Start powershell and output help to standard output
+                        $pwsh = Get-Process -Id $pid | Select-Object -ExpandProperty Path
+                        & $pwsh -help
+                    }
+
+                    It 'i' {
+                        1 | Should -Be 1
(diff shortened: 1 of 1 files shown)


nohwnd:
Closing - this never reproduced #2655 (the winrm + `pwsh -help` setup doesn't crash here), it's test-only with a `dt` left in and is Windows-only. Leaving #2655 open for a real repro and fix.

PR #2683, opened 2026-03-16
Remove coverageGutters, make coverage relative to repo root, add repo root

### LINKED ISSUE #2668: Filenames in JaCoCo report seem to be wrong
### Checklist

### What is the issue?

We create code coverage files and want to visualize the results in Jenkins (https://www.jenkins.io/) by using the coverage plugin (https://plugins.jenkins.io/coverage/). After an update of that plugin results could no longer be parsed by it and it seems that the xml file produced by Pester is wrong (Discussed here: https://issues.jenkins.io/browse/JENKINS-76044)

Results look e.g. like this:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE report PUBLIC "-//JACOCO//DTD Report 1.1//EN" "report.dtd"[]>
<report name="Pester (08/28/2025 05:08:10)">
  <sessioninfo id="this" start="1756357668799" dump="1756357690892" />
  <package name="myRepo/someDirectory/anotherDirectory">
    <class name="myRepo/someDirectory/anotherDirectory/foo" sourcefilename="someDirectory/anotherDirectory/foo.ps1">
      <method name="<script>" desc="()" line="16">
        <counter type="INSTRUCTION" missed="0" covered="31" />
        <counter type="LINE" missed="0" covered="22" />
        <counter type="METHOD" missed="0" covered="1" />
      </method>
      <counter type="INSTRUCTION" missed="0" covered="31" />
      <counter type="LINE" missed="0" covered="22" />
      <counter type="METHOD" missed="0" covered="1" />
      <counter type="CLASS" missed="0" covered="1" />
    </class>
    <sourcefile name="someDirectory/anotherDirectory/foo.ps1">
      <line nr="16" mi="0" ci="1" mb="0" cb="0" />
      <line nr="18" mi="0" ci="1" mb="0" cb="0" />
      <line nr="20" mi="0" ci="2" mb="0" cb="0" />
  // more lines
      <counter type="INSTRUCTION" missed="0" covered="31" />
      <counter type="LINE" missed="0" covered="22" />
      <counter type="METHOD" missed="0" covered="1" />
      <counter type="CLASS" missed="0" covered="1" />
    </sourcefile>
    <counter type="INSTRUCTION" missed="0" covered="31" />
    <counter type="LINE" missed="0" covered="22" />
    <counter type="METHOD" missed="0" covered="1" />
    <counter type="CLASS" missed="0" covered="1" />
  </package>
  // more packages
  
```

### Expected Behavior

I'm not an expert for JaCoCo but i assume that this line
`<class name="myRepo/someDirectory/anotherDirectory/foo" sourcefilename="someDirectory/anotherDirectory/foo.ps1">`

Should instead look like this
`<class name="myRepo/someDirectory/anotherDirectory/foo" sourcefilename="foo.ps1">`

Probably other occurences of `sourcefile` also need to be adapted

### Steps To Reproduce

1. Create 2 modules (or directories)
2. In each of those we need some code that does stuff
3. Then each needs a file to run tests on those
4. The tests are executed with a configuration like that:
```powershell
Import-Module Pester
$config = [PesterConfiguration]@{
	CodeCoverage = @{
	    Enabled = $true
		Path = @(
			".\module1\public\Start-Foo.ps1"
			".\module2\public\Start-Bar.ps1"
		)
		OutputPath = ".\Temp\coverage.xml"
	}
	Run = @{
		Path = @(
			".\module1\Module1.Unit.Tests.ps1"
			".\module2\Module2.Unit.Tests.ps1"
		)
		PassThru = $true
	}
	Output = @{
		Verbosity = "Detailed"
	}
}
```

### Describe your environment

Pester version     : 5.7.1 C:\Users\z003sk5z\Documents\PowerShell\Modules\Pester\5.7.1\Pester.psm1
PowerShell version : 7.5.4
OS version         : Microsoft Windows NT 10.0.22631.0

### Possible Solution?

Probably something like this is needed to get the file names:
`Split-Path $fileName -Leaf`

Probably more than this naive approach is needed.

nohwnd:
## PR Summary

Add detection of repository root, to find a common path for reports and potentially other stuff. Simply look for .git folder, which is a common way we do it in other repos that works well, and .git is basically used everywhere. If not found use the working directory, which is most often the repo root anyway. User can override the value if it does not work for them. It allows CI tools to work well, because they usually care about relative paths to the checkout directory, without forcing all paths to absolute.

Remove CoverageGutters format from coverage, because it was just JaCoCo with the paths correctly formed against coverage.xml location, which was in the repo root.

Removes the "get common folder" function which returns result dependent on the files you provided to the coverage in non-obvious way.

Fix https://github.com/pester/Pester/issues/2668

## PR Checklist

- [x] PR has meaningful title
- [ x Summary describes changes

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -65,6 +65,7 @@ param (
 )
 
 $ErrorActionPreference = 'Stop'
+$ErrorView = 'NormalView'
 Get-Module Pester | Remove-Module
 
 if ($Clean -and $PSVersionTable.PSVersion -lt [version]'5.1') {
@@ -105,6 +106,10 @@ if ($Clean) {
     }
 
     function Format-NicelyMini ($value) {
+        if ($null -eq $value) {
+            return '$null'
+        }
+
         if ($value -is [bool]) {
             if ($value) {
                 '$true'
@@ -148,7 +153,12 @@ if ($Clean) {
         foreach ($r in $section.PSObject.Properties.Name) {
             $option = $section.$r
             $default = Format-NicelyMini $option.Default
-            $type = $option.Default.GetType() -as [string]
+            if ("RepoRoot" -eq $r) {
+                # RepoRoot is a special case, because it is not actually an option, but a property that returns the value of the option. This is done to make it easier to use in scripts, but it causes issues when generating help, because it does not have a default value. So we need to get the default value from the actual option.
+                $default = Format-NicelyMini '<path of .git>'
+            }
+            # When the value is null this would throw if we checked the type of default value.
+            $type = $option.GetType().BaseType.GenericTypeArguments[0] -as [string]
             "    ${r}: $($option.Description)$eol    Type: ${type}$eol    Default value: ${default}$eol"
         }
     }
--- a/src/csharp/Pester/CodeCoverageConfiguration.cs
+++ b/src/csharp/Pester/CodeCoverageConfiguration.cs
@@ -31,6 +31,7 @@ namespace Pester
         private BoolOption _useBps;
         private BoolOption _singleHitBreakpoints;
         private DecimalOption _coveragePercentTarget;
+        private StringOption _reportRoot;
 
         public static CodeCoverageConfiguration Default { get { return new CodeCoverageConfiguration(); } }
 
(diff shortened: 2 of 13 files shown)


nohwnd:
There are 2 new options, run.repoRoot, and coverage.reportRoot, not sure if making the names similar is genius or stupid. But I like both names.

nohwnd:
finally green in case you want to have a look @fflaten

PR #2689, opened 2026-04-03
Fix #2555: Should-BeNull treats empty pipeline input as null

### LINKED ISSUE #2555: `Should-BeNull`: Strange behaviour when `$null`
### Checklist

### What is the issue?

I'm getting that the `$result` passed to `Should-BeNull` is not `$null`, but the verbose output I added to debug this do say it is not an array and it is indeed `$null`. 🤔  Not sure if I'm doing something wrong. I have seen this in another place as well but ignored it there, when I got it here I thought this must be an error somewhere. 🙂 

The `Out-Diff` currently only writes `Write-Information` messages and returns no values.
I output verbose message to determine that it is indeed $null and not an array,  still `Should-BeNull` thinks it is. 

```powershell
It 'Should output to console' {
    $expected = @(
        'My String that is longer than actual'
    )
    $actual = @(
        'My string that is shorter'
    )

    $result = Out-Diff -ExpectedString $expected -ActualString $actual
    Write-Verbose ($result -is [array]) -Verbose
    Write-Verbose ($null -eq $result) -Verbose
    $result | Should-BeNull -Debug -Verbose
}
```  

```plaintext
VERBOSE: False
VERBOSE: True
[-] Out-Diff.When output is made as informational message.Should output to console 124ms (122ms|2ms)
 Expected $null, but got [Object[]] '@()'.
 at $result | Should-BeNull -Debug -Verbose, /Users/johlju/source/Viscalyx.Common/tests/Unit/Public/Out-Diff.tests.ps1:94
```

### Expected Behavior

Should pass test as the actual value is $null.

### Steps To Reproduce

Don't have a reproducible example unfortunately, but if necessary I will try to make one. 

### Describe your environment

```
Pester version     : 6.0.0-alpha4 /Users/johlju/source/Viscalyx.Common/output/RequiredModules/Pester/6.0.0/Pester.psm1  
PowerShell version : 7.4.4
OS version         : Unix 14.5.0
```

### Possible Solution?

Hopefully I'm just doing something wrong. 🙂

nohwnd:
Fix #2555

When a function returns no output and the result is piped to Should-BeNull, PowerShell sends an empty array through the pipeline. Should-BeNull now recognizes this as no output and treats it as null.

- Pipeline empty array: passes (void function output)
- Explicit parameter empty array: still fails

Copilot-generated fix.

### DIFF
--- a/src/functions/assert/General/Should-BeNull.ps1
+++ b/src/functions/assert/General/Should-BeNull.ps1
@@ -32,6 +32,15 @@
 
     $collectedInput = Collect-Input -ParameterInput $Actual -PipelineInput $local:Input -IsPipelineInput $MyInvocation.ExpectingInput -UnrollInput
     $Actual = $collectedInput.Actual
+
+    # When a function returns no output and the result is piped to Should-BeNull,
+    # PowerShell sends an empty array @() through the pipeline. Treat empty pipeline
+    # input as $null since "no output" is effectively null.
+    # See https://github.com/pester/Pester/issues/2555
+    if ($collectedInput.IsPipelineInput -and $Actual -is [array] -and $Actual.Count -eq 0) {
+        $Actual = $null
+    }
+
     if ($null -ne $Actual) {
         $Message = Get-AssertionMessage -Expected $null -Actual $Actual -Because $Because -DefaultMessage "Expected `$null,<because> but got <actualType> '<actual>'."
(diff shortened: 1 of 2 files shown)


nohwnd:
Superseded by PR 2708, which fixes the same issue at the right layer: `Collect-Input -UnrollInput` now returns `$null` when the pipeline collected no items, so `@() | Should-BeNull` (and `Should-Be -Expected $null`, `Should-NotBeNull`, …) all do the right thing.

With #2708 merged, the patch on `Should-BeNull.ps1` here is dead code, and the two tests added in this PR are already covered (plus more cases for `@($null)`, `,$null`, multi-item arrays) on `main`. Closing.

PR #2690, opened 2026-04-03
Fix #2561: Escape control characters in assertion error messages

### LINKED ISSUE #2561: `Should-BeString`:  Doesn't handle escape character correctly in error message
### Checklist

### What is the issue?

If the actual value or expected value contain escape character (for example an ANSI sequence) the message are not shown correctly.

### Expected Behavior

Should show the special character like escape characters. Also ANSI sequences should not be hidden as they are part of the strings being compared.

### Steps To Reproduce

```powershell
Describe 'Something' {
    It 'Escape in expected value should be handled correctly' {
        '31;3;5m' | Should-BeString "`e[313;5m"
    }

    It 'Escape in actual value should be handled correctly' {
        "`e31;3;5m" | Should-BeString '[31;3;5m'
    }

    It 'Escape in both expected and actual value should be handled correctly' {
        "`e31;3;5m" | Should-BeString "`e[31;3;5m"
    }
}
```

Outputs:

```plaintext
Discovery found 11 tests in 30ms.
Running tests.
[-] Something.Escape in expected value should be handled correctly 10ms (9ms|2ms)
 Expected [string] '', but got [string] '31;3;5m'.
 at '31;3;5m' | Should-BeString "`e[313;5m", /Users/johlju/source/Viscalyx.Common/tests/Unit/Public/ConvertTo-AnsiSequence.tests.ps1:89
[-] Something.Escape in actual value should be handled correctly 24ms (22ms|2ms)
 Expected [string] '[31;3;5m', but got [string] '1;3;5m'.
 at "`e31;3;5m" | Should-BeString '[31;3;5m', /Users/johlju/source/Viscalyx.Common/tests/Unit/Public/ConvertTo-AnsiSequence.tests.ps1:93
[-] Something.Escape in both expected and actual value should be handled correctly 37ms (30ms|6ms)
 Expected [string] '', but got [string] '1;3;5m'.
 at "`e31;3;5m" | Should-BeString "`e[31;3;5m", /Users/johlju/source/Viscalyx.Common/tests/Unit/Public/ConvertTo-AnsiSequence.tests.ps1:97
Tests completed in 432ms
Tests Passed: 8, Failed: 3, Skipped: 0, Inconclusive: 0, NotRun: 0
```

### Describe your environment

```
Pester version     : 6.0.0-alpha4 /Users/johlju/source/Viscalyx.Common/output/RequiredModules/Pester/6.0.0/Pester.psm1  
PowerShell version : 7.4.4
OS version         : Unix 14.6.1
```

### Possible Solution?

Maybe convert special characters in actual and expected value to their unicode equivalent? 

```powershell
"`0`a`b`t`f`r`n`e" `
    -replace "`0", '␀' `
    -replace "`a", '␇' `
    -replace "`b", '␈' `
    -replace "`t", '␉' `
    -replace "`f", '␌' `
    -replace "`r", '␍' `
    -replace "`n", '␊' `
    -replace "`e", '␛'
```

Outputs:
```bash

␀␇␈␉␌␍␊␛
```

nohwnd:
Fix #2561

Format-String2 now escapes control characters as Unicode control pictures so they are visible in error messages. Characters like ESC (used in ANSI sequences) now display as their Unicode symbol instead of being invisible.

Copilot-generated fix.

### DIFF
--- a/src/Format2.ps1
+++ b/src/Format2.ps1
@@ -46,11 +46,17 @@ function Format-Object2 ($Value, $Property, [switch]$Pretty) {
 }
 
 function Format-String2 ($Value) {
-    if ('' -eq $Value) {
+    # Use .Length instead of '' -eq $Value because PowerShell's -eq operator
+    # considers some control characters (NUL, BEL, BS, ESC) equal to empty string.
+    if ($null -eq $Value -or $Value.Length -eq 0) {
         return '<empty>'
     }
 
-    "'$Value'"
+    # Escape ASCII control characters (0x00..0x1F) to the Unicode "Control
+    # Pictures" block (U+2400..U+241F) so they remain visible in error messages.
+    # See https://github.com/pester/Pester/issues/2561. Hot loop lives in C#
+    # (Pester.Formatter) for speed.
+    "'" + [Pester.Formatter]::EscapeControlChars($Value) + "'"
 }
 
 function Format-Null2 {
--- a/src/csharp/Pester/Formatter.cs
+++ b/src/csharp/Pester/Formatter.cs
@@ -1,0 +1,59 @@
+﻿using System;
+
+namespace Pester
+{
+    public static class Formatter
+    {
+        private static readonly char[] ControlChars = BuildControlChars();
+
+        private static char[] BuildControlChars()
+        {
+            var chars = new char[0x20];
+            for (int i = 0; i < 0x20; i++)
+            {
+                chars[i] = (char)i;
+            }
+            return chars;
+        }
+
+        /// <summary>
+        /// Replaces ASCII control characters (0x00..0x1F) in <paramref name="value"/>
+        /// with the matching Unicode "Control Pictures" code point
+        /// (U+2400..U+241F) so they are visible when printed. See
+        /// https://github.com/pester/Pester/issues/2561.
+        ///
+        /// Returns <paramref name="value"/> unchanged when it is null, empty, or
+        /// contains no control characters — common case, no allocation beyond
+        /// the IndexOfAny scan.
+        /// </summary>
+        public static string EscapeControlChars(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            int firstControl = value.IndexOfAny(ControlChars);
+            if (firstControl < 0)
+            {
+                return value;
+            }
+
+            int len = value.Length;
+            var buf = new char[len];
+            value.CopyTo(0, buf, 0, len);
+
+            // Everything before firstControl is known to be safe; start from there.
+            for (int i = firstControl; i < len; i++)
+            {
+                char c = buf[i];
+                if (c < 0x20)
+                {
+                    buf[i] = (char)(c + 0x2400);
+                }
+            }
+
+            return new string(buf);
+        }
+    }
+}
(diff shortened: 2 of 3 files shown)


src/Format2.ps1
```diff
@@ -50,6 +50,19 @@ function Format-String2 ($Value) {
...
+    # and make error output confusing. See https://github.com/pester/Pester/issues/2561
+    $Value = $Value `
+        -replace "`0", '␀' `
+        -replace "`a", '␇' `
+        -replace "`b", '␈' `
+        -replace "`t", '␉' `
+        -replace "`f", '␌' `
+        -replace "`r", '␍' `
+        -replace "`n", '␊' `
+        -replace "`e", '␛'
```
nohwnd: feels like we are re-constructing the same string so many times this way, we should have some similar code that does this by prepenting the code point with a bit that makes it visible. and does that via string builder and a loop. basically re-building the whole string just once.

nohwnd: special chars are uncommon, so this should be as cheap as possible.

nohwnd:
Rewrote at eb5917c per your comments — single StringBuilder pass, fast path for the common no-special-chars case:

- Module-load `[char[]] $script:Format2ControlChars` covering 0x00..0x1F.
- `String.IndexOfAny` for the cheap probe; if none, just `"'$Value'"` — no rebuild.
- Slow path: one StringBuilder, one foreach over `GetEnumerator()`, control chars get `+ 0x2400` (Unicode Control Pictures, same trick as Format-CDataString in NUnit3).

Micro-bench (50k calls, PS 7.6.2):
- fast path: ~1.4s
- slow path: ~6.8s

Side benefit: now escapes the entire 0x00..0x1F range instead of just 8 hand-listed chars — vertical tab, FS/GS/RS/US etc. also become visible.

Tests: 69/69 green on both PS 7.6 and PS 5.1.

nohwnd:
Moved the hot loop to `Pester.Formatter.EscapeControlChars` (`src/csharp/Pester/Formatter.cs`) at d89c54d. `Format-String2` is now a one-liner delegating to the C# static.

Correctness on 200,000-char input with 10k each of NUL/TAB/LF/ESC scattered through it — all 40k control chars converted to U+2400..U+241B, zero raw control chars left.

Bench (1k iterations, PS 7.6.2):

| length  | fast (no specials) | slow (1 ESC) | slow (all ESC) |
|--------:|-------------------:|-------------:|---------------:|
|     100 |              81 ms |        56 ms |          98 ms |
|   1,000 |              80 ms |       107 ms |          95 ms |
|  10,000 |             422 ms |       386 ms |         382 ms |
| 100,000 |           3,685 ms |     3,754 ms |       3,784 ms |

For a typical short error-message string fast and slow paths are both ~0.08 ms per call. 69/69 tests green on PS 7.6 and PS 5.1.

nohwnd:
Good call — pulled the `<empty>` check and quote-wrapping back out into `Format-String2` at 6f6e181. The C# helper now only does what its name says (escape control chars; returns input unchanged when no escaping is needed). 69/69 still green on PS 7.6 and PS 5.1.

```powershell
function Format-String2 ($Value) {
    if ($null -eq $Value -or $Value.Length -eq 0) {
        return '<empty>'
    }
    "'" + [Pester.Formatter]::EscapeControlChars($Value) + "'"
}
```

PR #2692, opened 2026-04-03
Fix #2474: DirectoryInfo/FileInfo display property map prevents infinite loop

### LINKED ISSUE #2474: Infinite loop for `FileInfo` and `DirectoryInfo` due to `Root` properties in `DirectoryInfo`. Should we restrict depth as a general fix in addition to specialcasing `DirectoryInfo` in `Get-DisplayProperty2`?
Infinite loop for `FileInfo` and `DirectoryInfo` due to `Root` properties in `DirectoryInfo`. Should we restrict depth as a general fix in addition to specialcasing `DirectoryInfo` in `Get-DisplayProperty2`?

_Originally posted by @fflaten in https://github.com/pester/Pester/pull/2428#discussion_r1606553664_

nohwnd:
Fix #2474

Added `System.IO.DirectoryInfo` and `System.IO.FileInfo` to the display property map in `Get-DisplayProperty2` to prevent infinite recursion from circular `Root`/`Directory` properties.

Copilot-generated fix.

### DIFF
--- a/src/Format2.ps1
+++ b/src/Format2.ps1
@@ -157,7 +157,10 @@ function Get-DisplayProperty2 ([Type]$Type) {
     # and for types that do not exist
 
     $propertyMap = @{
-        'System.Diagnostics.Process' = 'Id', 'Name'
+        'System.Diagnostics.Process'  = 'Id', 'Name'
+        # DirectoryInfo and FileInfo have circular references (Root, Directory) that cause infinite recursion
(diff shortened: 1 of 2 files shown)


nohwnd:
enforce this by adding tests. there should be already some for process, if not, add also for process.

nohwnd:
🤖 Tests added in latest push — includes DirectoryInfo, FileInfo, and Process display properties. Force-pushed.

PR #2693, opened 2026-04-03
Fix #2515: Include hidden folders in test discovery on Linux

### LINKED ISSUE #2515: Test discovery on linux fails to find tests in hidden folders
### Checklist

### What is the issue?

I am writing tests for github actions which are in the `.github` folder so my tests are in `tests/.github`
`Invoke-Pester` does not find these tests unless I explicitly specify `tests/.github`

This works fine in windows powershell core but not in linux powershell core

I am using pester v5.6.0

### Expected Behavior

`Invoke-Pester` finds all tests in all folders, including hidden folders, without having to explicitly specify the folder name
This should work in linux and in windows

### Steps To Reproduce

1. Place a tests in a folder named tests/.github
2. Call `Invoke-Pester` from the parent folder

Observe that the test is discovered in windows but not in linux

### Describe your environment

Working (in windows):
```
Pester version     : 5.6.0 C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.0\Pester.psm1
PowerShell version : 5.1.22621.3672
OS version         : Microsoft Windows NT 10.0.22631.0
```
Not working (in WSL2 Ubuntu)
```
Pester version     : 5.6.0 /home/XXXX/.local/share/powershell/Modules/Pester/5.6.0/Pester.psm1
PowerShell version : 7.4.3
OS version         : Unix 5.15.153.1
```

### Possible Solution?

_No response_

nohwnd:
Fix #2515

Added `-Force` to `Get-ChildItem` calls in `Find-File` so hidden (dot-prefixed) folders like `.github` are included in test discovery on Linux.

Copilot-generated fix.


nohwnd:
read the comments on the original issue, you need to make sure you exclude .git folder, and also write tests, you have test-drive exactly for that.

tst/Pester.RSpec.ts.ps1
```diff
@@ -2966,4 +2966,43 @@ i -PassThru:$PassThru {
             $ex.Exception.Message | Verify-Like '*Unbound scriptblock*'
         }
     }
+
+    # Regression test for https://github.com/pester/Pester/issues/2515
+    # Find-File should discover test files inside hidden (dot-prefixed) directories.
+    # Without -Force on Get-ChildItem, hidden folders are silently skipped.
+    b "Find-File discovers hidden folders" {
+        t "discovers test files in dot-prefixed (hidden) subdirectories" {
+            $tempDir = Join-Path ([IO.Path]::GetTempPath()) "PesterHidden_$([IO.Path]::GetRandomFileName().Substring(0,4))"
```
nohwnd: there is TestDrive you don't need to duplicate that functionality.

tst/Pester.RSpec.ts.ps1
```diff
@@ -2966,4 +2966,43 @@ i -PassThru:$PassThru {
...
+            $hiddenDir = Join-Path $tempDir '.hidden'
```
nohwnd: powershell is slash agnostic, just use "$onepath/$otherpath" join-path is just noise. You onnly need to pay attention when you use .net apis.

nohwnd:
I see no test for excluding .git folder, which has lot of files that we are almost 100% sure  to not care about.

nohwnd:
🤖 Good point on .git exclusion. Will add test for that. Tests for hidden folder discovery were added in latest push.

src/Pester.RSpec.ps1
```diff
@@ -60,13 +61,30 @@
...
         }
     }
 
+    # Exclude files found inside version-control metadata directories.
+    # These are almost never wanted and .git in particular can contain many files.
+    $vcsDirectories = @('.git', '.svn', '.hg')
+    $filteredFiles = foreach ($f in $files) {
+        # normalize backslashes for cross-platform ease of use
+        $normalizedPath = $f.FullName -replace "/", "\"
+        $dominated = $false
```
nohwnd: what does `dominated` mean in this context?

src/Pester.RSpec.ps1
```diff
@@ -60,13 +61,30 @@
         else {
             # this is a path that does not exist so let's hope it is
             # a wildcarded path that will resolve to some files
-            & $SafeCommands['Get-ChildItem'] -Recurse -Path $p -Filter "*$Extension" -File
+            # use -Force to include hidden items (e.g. dot-prefixed folders on Linux)
+            & $SafeCommands['Get-ChildItem'] -Recurse -Path $p -Filter "*$Extension" -File -Force
```
nohwnd: can you use exclude path to exclude .git? it is potentially  expensive to enumerate, so we don't want to enumerate it and then filter it out.

tst/Pester.RSpec.ts.ps1
nohwnd: Done in 0bd903b — moved the new tests into the existing `Describe 'Find-File'` block in `tst/Pester.Tests.ps1` so they can use TestDrive directly.

tst/Pester.RSpec.ts.ps1
nohwnd: Done — the new tests live in `tst/Pester.Tests.ps1` now and use `TestDrive:\...` strings directly, no Join-Path.

src/Pester.RSpec.ps1
nohwnd: Gone in 0bd903b — the whole post-filter loop is gone, so there's no more `dominated`. The walker decides at each subdirectory whether to descend, so files under `.git`/`.svn`/`.hg` are never enumerated in the first place.

src/Pester.RSpec.ps1
nohwnd: Done in 0bd903b. Rewrote the directory branch to walk the tree itself (new `Find-FileInDirectory` helper) so `.git`/`.svn`/`.hg` are skipped at the directory level — their contents are never enumerated.

The wildcard-fallback branch (when `Test-Path` returns false) still does a post-filter because the shape of those paths is harder to walk by hand, but that path is rare.

Added an equivalence test in `tst/Pester.Tests.ps1` that asserts the result equals `Get-ChildItem -Recurse -Filter '*.Tests.ps1' -File -Force` minus VCS folders, and verified locally that removing the skip makes that test fail.

PR #2699, opened 2026-04-21
Improve error message for array comparison in Should -Be

### LINKED ISSUE #1154: Should -Be gives confusing error message comparing arrays that are not enumerated.

In writing tests to ensure the accurate fixing of the issue resolved by [this PR](https://github.com/PowerShell/PowerShell/pull/8407), I discovered that `Should` _can and will_ differentiate between a complete array passed over the pipeline and an enumerated array, and it seems to expect the latter.

Simple repro:

```powershell
$a = 1, 2, 3
 ,$a |Should -Be $a
```

Operating System, Pester version, and PowerShell version:

```
Pester version     : 4.3.1 C:\Program Files\PowerShell\Modules\Pester\4.3.1\Pester.psd1
PowerShell version : 6.1.0
```

## 3. Expected Behavior

`Should` should more clearly indicate what is happening, with either symbolic representation of the different element that more closely reflects what's happened on the command line, or by explicitly calling out the fact that the two arrays are different in some other way.

Bonus: perhaps an optional switch or operator might be a good idea that _doesn't_ check for this, however this is getting picked up.

## 4.Current Behavior

Clearly, it is able to distinguish between what is a multidimensional array that has been enumerated, and a flat array. However, its error message is the polar opposite of clear. 😄 

Simple example:

```powershell
PS> $a = 1, 2, 3
PS> ,$a |Should -Be $a
Expected @(1, 2, 3), but got @(1, 2, 3).
At C:\Program Files\PowerShell\Modules\Pester\4.3.1\Functions\Assertions\Should.ps1:206 char:9
+         throw ( New-ShouldErrorRecord -Message $testResult.FailureMes ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidResult: (System.Collections.Hashtable:Hashtable) [], Exception
```
\

## 6. Context

https://github.com/PowerShell/PowerShell/pull/8407

It was noted that `Get-Variable` passes collections whole over the pipeline, thus similar to the original example, this too throws a strangely confusing error. This particular case is fixed in the linked PR, however, making the command properly enumerate collections when outputting.

```powershell
$v = 1, 2, 3
Get-Variable -Name v -ValueOnly | Should -Be $v
```

yezhi0:
This PR improves the error message when `Should -Be` compares two arrays that differ in length or type, but appear identical due to unrolling.

Changes:
- Add detection for collection length and type differences.
- Show a clear message with expected and actual lengths and types.
- Preserve existing behavior for string comparisons.

Fixes #1154

### DIFF
--- a/src/functions/assertions/Be.ps1
+++ b/src/functions/assertions/Be.ps1
@@ -47,22 +47,29 @@ function Should-BeAssertion ($ActualValue, $ExpectedValue, [switch] $Negate, [st
 }
 
 function ShouldBeFailureMessage($ActualValue, $ExpectedValue, $Because) {
-    # This looks odd; it's to unroll single-element arrays so the "-is [string]" expression works properly.
-    $ActualValue = $($ActualValue)
-    $ExpectedValue = $($ExpectedValue)
 
-    if (-not (($ExpectedValue -is [string]) -and ($ActualValue -is [string]))) {
-        return "Expected $(Format-Nicely $ExpectedValue),$(if ($null -ne $Because) { Format-Because $Because }) but got $(Format-Nicely $ActualValue)."
+    $actualType = if ($null -eq $ActualValue) { '<null>' } else { $ActualValue.GetType().Name }
+    $expectedType = if ($null -eq $ExpectedValue) { '<null>' } else { $ExpectedValue.GetType().Name }
+    
+    $actualCount = if ($ActualValue -is [Array]) { $ActualValue.Count } else { 1 }
+    $expectedCount = if ($ExpectedValue -is [Array]) { $ExpectedValue.Count } else { 1 }
+
+
+    $ActualValueUnrolled = $($ActualValue)
+    $ExpectedValueUnrolled = $($ExpectedValue)
+
+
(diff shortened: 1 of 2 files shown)


src/functions/assertions/Be.ps1
```diff
@@ -45,25 +45,39 @@ function Should-BeAssertion ($ActualValue, $ExpectedValue, [switch] $Negate, [st
         }
     }
 }
-
 function ShouldBeFailureMessage($ActualValue, $ExpectedValue, $Because) {
-    # This looks odd; it's to unroll single-element arrays so the "-is [string]" expression works properly.
-    $ActualValue = $($ActualValue)
-    $ExpectedValue = $($ExpectedValue)
+    # 获取类型和长度信息
```
nohwnd: can we get english comments

nohwnd (commented):
provide tests for the change you are making

nohwnd (approved):

PR #2700, opened 2026-05-05
Enhance module compatibility: support nested module resolution slash notation

wetwicky:
## PR Summary

Enable users to specify a nested module unambiguously in `-ModuleName` parameters across all Pester mocking and scoping commands (`Mock`, `InModuleScope`, `Should-Invoke`, `Should-NotInvoke`) using rooted nested-module notation (`'RootModuleName/NestedModuleName'` and `'RootModuleName\NestedModuleName'`). When this notation is detected, resolve the root module first, then locate the nested module within it and return its session state. Plain module names (without slash notation) must continue to work without any change in behavior.

## PR Checklist

- [x] PR has meaningful title

### Existing Concepts (from codebase)

- **ModuleName**: A string parameter accepted by `Mock`, `Should-Invoke`, `Should-NotInvoke`, and `InModuleScope` that identifies the target PowerShell module into which a mock is injected or within which test code is executed. Currently expects a simple module name (e.g., `"Pester"`, `"MyModule"`).
- **Get-CompatibleModule**: Internal Pester function (`src/functions/InModuleScope.ps1`) that resolves a module name string to a loaded `PSModuleInfo` object via `Get-Module -Name $ModuleName -All`. Used by both `InModuleScope` and `Resolve-Command` (in `Mock.ps1`). It is the single gateway for module resolution throughout the mocking and scoping system.
- **Resolve-Command**: Internal function (`src/functions/Mock.ps1`) that resolves the command to be mocked within a module's session state. Delegates module lookup entirely to `Get-CompatibleModule`.
- **TargetModule**: The resolved module name stored on `MockBehavior` and context info objects, used as part of the mock table key (`"$ModuleName||$CommandName"`) to track call history per module scope.
- **NestedModules**: A PowerShell module manifest concept (`psd1`) where a root/manifest module loads subordinate script modules. The nested modules share the root's session state when loaded via a manifest. Pester already has a test (`tst/functions/InModuleScope.Tests.ps1`, line ~325) demonstrating `InModuleScope` working against manifest modules that have nested modules.
- **RootModule (manifest key)**: In `Pester.psd1`, `RootModule = 'Pester.psm1'`. In the PowerShell module system, a manifest's `RootModule` is the primary script module; `NestedModules` lists additional modules loaded alongside it.

### New Concepts Required

- **ModuleName slash-notation**: A `RootModule/NestedModule` string format (e.g., `"Pester/Pester.Runtime"`) to be parsed by `Get-CompatibleModule` (and propagated through `Resolve-Command`) that uniquely identifies a nested module within a specific root module. This concept does not yet exist anywhere in the Pester codebase.
- **Module disambiguation path**: The resolution algorithm that, given `RootModule/NestedModule`, first finds the root module, then narrows the loaded `NestedModules` list to find the specific nested module, returning its `PSModuleInfo` — enabling disambiguation when the same nested module name appears under different root modules.

### Key Business Rules

- **Format rule**: When `ModuleName` contains exactly one `/`, the left side is the root/manifest module name and the right side is the nested module name. A plain name (no `/`) retains current behavior — no breaking change.
- **Root must exist**: The root module named by the left side must be currently loaded and be of type `Script` or `Manifest`; otherwise error with a clear message.
- **Nested must be a child**: The nested module named by the right side must be found in the root module's `NestedModules` list or otherwise be a module that was loaded as a consequence of importing the root; otherwise error.
- **Ambiguity resolution**: If multiple root modules share the same name, the slash notation still requires the root to be unambiguous (same rule as today for plain names). The value-add is when the *nested* module name is ambiguous across different roots.
- **Consistency**: The resolved `TargetModule` stored on `MockBehavior` must remain a plain module name (or the full slash path — TBD, see Key Design Decisions), so that internal comparisons such as `$ModuleName -eq $command.Mock.Hook.OriginalCommand.Module.Name` remain correct.

---

### Solution Direction

- Parse `ModuleName` for the `RootModule/NestedModule` separator in `Get-CompatibleModule` (the single module-resolution gateway). When detected, split on the first `/`, resolve the root module, then look up the nested module within it. Return the nested module's `PSModuleInfo`. All callers of `Get-CompatibleModule` — `InModuleScope` and `Resolve-Command` — automatically gain the capability.
- `InModuleScope` and external-API functions (`Mock`, `Should-Invoke`, `Should-NotInvoke`) accept `[string] $ModuleName` unchanged; no parameter type changes are needed.

### Key Design Decisions

- **Where to parse the slash notation** — Options: (a) in `Get-CompatibleModule` only, (b) at each call site. → **Recommendation**: Option (a). `Get-CompatibleModule` is the sole resolution gateway; centralising the logic there avoids duplication and ensures `InModuleScope`, `Mock`, and assert commands all behave consistently without touching call sites.

- **What to store as TargetModule on MockBehavior** — Options: (a) store the plain nested module name, (b) store the full `RootModule/NestedModule` path. The stored value is used in the mock call-history key (`"$ModuleName||$CommandName"`) and in comparisons like `$ModuleName -eq $command.Mock.Hook.OriginalCommand.Module.Name`. PowerShell's `Module.Name` is always a plain name. → **Recommendation**: Store the resolved nested module's plain `Name` (e.g., `"Pester.Runtime"`), not the slash path. This keeps internal comparisons correct. `Resolve-Command` already sets `TargetModule = $ModuleName` *after* calling `Get-CompatibleModule`; the resolved plain name must be propagated back to the callers, meaning `Resolve-Command` should capture `$module.Name` (the resolved nested module name) and use that as `TargetModule`.

- **Nested module lookup strategy** — Options: (a) iterate `$rootModule.NestedModules`, (b) call `Get-Module -Name $nestedModuleName` and filter to those where the `RootModule*` parent is the resolved root, (c) use `Get-Module -All` and filter by checking `$_.Path` starts with the root module's directory. → **Recommendation**: Option (a) — directly enumerate `$rootModule.NestedModules` and match by name. This is most precise and does not depend on path assumptions. It also mirrors how PowerShell itself tracks nested module relationships.

### Alternatives Considered

- **Add a separate `-NestedModuleName` parameter to `Mock`, `InModuleScope`, etc.**: Rejected because it requires changes to all public API surfaces, breaks existing scripts that splat parameters, increases complexity for consumers, and creates two mechanisms for the same concept.
- **Use `Get-Module -Name $nestedName` and disambiguate by checking parent**: Partially viable but fragile — PowerShell's module graph doesn't always expose a clear parent link from a nested module back to its manifest. Direct enumeration of `$rootModule.NestedModules` is more reliable.

---

### Requirement Ambiguities

- **What constitutes the "root"**: The left side of `/` — is it always the *manifest* module name, or can it also be any parent psm1 that imported nested modules explicitly? Modules imported via `Import-Module` with `-NestedModules` behave differently from manifest-defined nesting. Needs clarification.
- **Slash in module names**: PowerShell module names do not normally contain `/`, but this is not enforced by the module system. If a module is legitimately named `A/B`, the parsing would break. The probability is extremely low, but the edge case should be documented.

### Edge Cases

- **Single slash with no right side** (`"Pester/"`) or **no left side** (`"/Runtime"`): Parser must validate and throw a descriptive error rather than silently producing unexpected behavior.
- **Root module found but nested module not found within it**: Must throw an error that names both the root and the attempted nested module name, rather than a generic `Get-Module` error.
- **Nested module already loaded as a top-level module** (imported independently in the same session alongside being a nested module of a root): The slash notation disambiguates which instance to use. The algorithm must return the instance that is nested under the specified root, not any independently loaded copy.
- **Root module has not been imported yet**: `Get-Module -Name $rootName` returns nothing. Error message should guide the user to import it.
- **Manifest modules whose `RootModule` key is a nested module**: `Pester.psd1` sets `RootModule = 'Pester.psm1'`. When someone writes `Pester/Pester.psm1`, the right side matches the `RootModule` value, not a `NestedModules` entry — edge case that may need consideration.

### Technical Risks

- **TargetModule key mismatch in MockTable**: The mock call-history hashtable is keyed on `"$ModuleName||$CommandName"`. If the caller passes `"Pester/Pester.Runtime"` to `Mock` and the resolved TargetModule stored is `"Pester.Runtime"`, but `Should-Invoke` is later called with `ModuleName = "Pester/Pester.Runtime"`, the lookup must also resolve and use the plain name. `Resolve-Command` is called by `Should-Invoke` too; as long as it always normalises to the plain nested name, the key matches. Risk is low if the resolution always normalises — but needs a test.
- **`$ModuleName -eq $command.Mock.Hook.OriginalCommand.Module.Name` comparison** in `Resolve-Command` (lines 735, 755): This compares `TargetModule` to `OriginalCommand.Module.Name`. If `TargetModule` is stored as the slash path rather than the plain name, these comparisons silently fail, breaking `IsFromTargetModule`. Medium risk — mitigated by the decision to store the plain name.
- **`EscapeSingleQuotedStringContent $ModuleName`** used in the mock bootstrap script body (`Mock.ps1` line 170): The bootstrap script embeds `ModuleName` in a single-quoted string literal. The `/` character is safe in single-quoted strings, but if the slash notation is used as-is (rather than the resolved plain name), the embedded string literal would be `'Pester/Pester.Runtime'` — still valid PowerShell but would break the `Invoke_Mock` dispatch which uses `ModuleName` as a key. Low-medium risk — resolved by always normalising before embedding.

### Acceptance Criteria Coverage

| AC# | Description                                                                         | Addressable?  | Gaps/Notes                                                                              |
| --- | ----------------------------------------------------------------------------------- | ------------- | --------------------------------------------------------------------------------------- |
| 1   | `Mock -ModuleName 'Root/Nested' Get-Item {}` sets up a mock in the nested module    | Yes           | Requires `Get-CompatibleModule` to parse and resolve; `TargetModule` must be normalised |
| 2   | `InModuleScope 'Root/Nested' { ... }` executes in the nested module's session state | Yes           | Same `Get-CompatibleModule` change covers this                                          |
| 3   | `Should-Invoke -ModuleName 'Root/Nested' Get-Item` correctly finds call history     | Yes           | Depends on consistent normalisation of `TargetModule` to plain name                     |
| 4   | Plain `ModuleName` (no `/`) continues to work unchanged                             | Yes           | Conditional branch — no existing code path changes                                      |
| 5   | Clear error when root module not found                                              | Yes           | Must add explicit error message for split-notation path                                 |
| 6   | Clear error when nested module not found under root                                 | Yes           | Must add explicit error message — not covered by current `Get-Module` error             |
| —   | Multi-level nesting (`A/B/C`)                                                       | Yes           | Must support 1 to N level  |

### DIFF
--- a/src/functions/InModuleScope.ps1
+++ b/src/functions/InModuleScope.ps1
@@ -156,10 +156,10 @@
     Set-ScriptBlockScope -ScriptBlock $ScriptBlock -SessionState $sessionState
     Set-ScriptBlockScope -ScriptBlock $wrapper -SessionState $sessionState
     $splat = @{
-        ScriptBlock    = $ScriptBlock
-        Parameters     = $Parameters
-        ArgumentList   = $ArgumentList
-        SessionState   = $sessionState
+        ScriptBlock  = $ScriptBlock
+        Parameters   = $Parameters
+        ArgumentList = $ArgumentList
+        SessionState = $sessionState
     }
 
     Write-ScriptBlockInvocationHint -Hint "InModuleScope" -ScriptBlock $ScriptBlock
@@ -172,6 +172,84 @@ function Get-CompatibleModule {
         [Parameter(Mandatory = $true)]
         [string] $ModuleName
     )
+
+    # Slash/backslash notation: RootModule/NestedModule[/DeeperNestedModule...]
+    # Resolves the full nested path from left to right so deeply nested modules can be targeted.
+    if ($ModuleName -match '[/\\]') {
+        # Split on / or \ and trim whitespace from each segment.
+        $modulePathSegments = @($ModuleName -split '[/\\]')
+        $modulePathSegments = @($modulePathSegments | & $SafeCommands['ForEach-Object'] { $_.Trim() })
+        $hasEmptySegment = @($modulePathSegments | & $SafeCommands['Where-Object'] { [string]::IsNullOrEmpty($_) }).Count -gt 0
+
+        # Require at least two non-empty segments (root + one nested level).
+        if ($modulePathSegments.Count -lt 2 -or $hasEmptySegment) {
+            throw "Invalid ModuleName format '$ModuleName'. Expected format: 'RootModuleName/NestedModuleName[/DeeperNestedModuleName...]'."
+        }
+
+        # Seed the search with all copies of the root module (Get-Module -All covers reimports).
+        $rootModuleName = $modulePathSegments[0]
+        if ($PesterPreference.Debug.WriteDebugMessages.Value) {
+            Write-PesterDebugMessage -Scope Runtime "Nested path notation detected in ModuleName '$ModuleName'. Resolving from root module '$rootModuleName'."
+        }
+
+        $currentModules = @(& $SafeCommands['Get-Module'] -Name $rootModuleName -All -ErrorAction SilentlyContinue)
+        if ($currentModules.Count -eq 0) {
+            throw "No modules named '$rootModuleName' are currently loaded."
+        }
+
+        # Walk each path segment after the root, narrowing $currentModules to the matched nested module at each level.
+        $resolvedPath = $rootModuleName
+        for ($index = 1; $index -lt $modulePathSegments.Count; $index++) {
+            $nestedModuleName = $modulePathSegments[$index]
+            $nextModules = [System.Collections.Generic.List[object]]@()      # matches for this segment
+            $availableNested = [System.Collections.Generic.List[string]]@()  # all sibling names, for error messages
+
+            if ($PesterPreference.Debug.WriteDebugMessages.Value) {
+                Write-PesterDebugMessage -Scope Runtime "Resolving nested module segment '$nestedModuleName' under '$resolvedPath'."
+            }
+
+            # Scan NestedModules of every candidate at the current level.
+            foreach ($parentModule in $currentModules) {
+                foreach ($nestedModule in @($parentModule.NestedModules)) {
+                    # Collect unique sibling names so the error message can list available options.
+                    if (-not [string]::IsNullOrEmpty($nestedModule.Name) -and -not $availableNested.Contains($nestedModule.Name)) {
+                        $availableNested.Add($nestedModule.Name)
+                    }
+
+                    if ($nestedModule.Name -eq $nestedModuleName) {
+                        $nextModules.Add($nestedModule)
+                    }
+                }
+            }
+
+            # No match: segment name is wrong or the nested module is not loaded.
+            if ($nextModules.Count -eq 0) {
+                $availableList = if ($availableNested.Count -gt 0) { $availableNested -join ', ' } else { '(none)' }
+                throw "No nested module named '$nestedModuleName' was found under '$resolvedPath'. Available nested modules: $availableList."
+            }
+
+            # More than one match: ambiguous — multiple loaded copies of the same module exist.
+            if ($nextModules.Count -gt 1) {
+                throw "Multiple nested modules named '$nestedModuleName' were found under '$resolvedPath' across loaded module copies. Make sure to remove any extra copies of the module from your session before testing."
+            }
+
+            $resolvedNested = $nextModules[0]
+            # Only Script/Manifest modules expose a usable session state for InModuleScope.
+            if ($resolvedNested.ModuleType -notin 'Script', 'Manifest') {
+                throw "Nested module '$nestedModuleName' in path '$resolvedPath/$nestedModuleName' is not a Script or Manifest module. Detected module type: '$($resolvedNested.ModuleType)'."
+            }
+
+            # Narrow to the single resolved module and advance the path tracker for the next iteration.
+            $currentModules = @($resolvedNested)
+            $resolvedPath = "$resolvedPath/$nestedModuleName"
+        }
+
+        if ($PesterPreference.Debug.WriteDebugMessages.Value) {
+            Write-PesterDebugMessage -Scope Runtime "Found nested module $($currentModules[0].Name) version $($currentModules[0].Version) in path $resolvedPath."
+        }
+
+        return $currentModules[0]
+    }
 
     try {
         if ($PesterPreference.Debug.WriteDebugMessages.Value) {
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -662,6 +662,11 @@ function Resolve-Command {
                 Write-PesterDebugMessage -Scope Mock "Found module $($module.Name) version $($module.Version)."
             }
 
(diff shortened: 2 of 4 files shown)


nohwnd (approved):

nohwnd:
@fflaten  if you want to have a look as well :)

nohwnd (changes requested):
Took a proper look and pushed a commit that reworks the disambiguation tests first — the same-name ones were loading the nested modules under *different* names (`Repository_A` / `Repository_B`, `SharedRepository` / `B_SharedRepository`), so a plain `-ModuleName` would have resolved them just as well and the tests passed without ever exercising the feature. The new versions load two nested modules with the same base name from sibling folders, assert the ambiguity with `Get-Module -All`, and check the resolved copy by its content.

Doing that surfaced a real problem on the Mock side, and that's why I'm requesting changes.

`InModuleScope` resolution is fine — `Root/Nested` lands in the right session state. But `Resolve-Command` normalizes `$ModuleName = $module.Name` back to the plain nested name, and the mock table is keyed `"$ModuleName||$CommandName"`. So two nested modules that share a name collide on the same key. Mocking both clobbers, and `Should-Invoke` mixes their history:

```powershell
# A\Repository.psm1 and B\Repository.psm1 both load as 'Repository'
Mock Get-Data -ModuleName 'ClientA/Repository' -MockWith { 'mockedA' }
Mock Get-Data -ModuleName 'ClientB/Repository' -MockWith { 'mockedB' }

InModuleScope 'ClientA/Repository' { Invoke-Api }                  # -> mockedB  (expected mockedA)
Should -Invoke Get-Data -ModuleName 'ClientB/Repository' -Times 0  # fails: "called 1 times"
```

The injection itself is isolated when you mock only one of them, so the single-mock tests pass. It's the shared key that breaks once both are in play.

What makes this worse than today: with two same-named modules loaded the current code throws `Multiple script or manifest modules named 'Repository' are currently loaded...`. The slash notation walks past that guard and gives silent wrong results instead of the error. So for the exact scenario this feature exists to solve, mocking is currently unsafe.

I don't have a clean fix off the top of my head — the key has to stay comparable to `OriginalCommand.Module.Name` (always the plain name) for `IsFromTargetModule`, so we can't just key on the slash path. Maybe key on the module identity/path and compare names separately, or at least keep throwing when the resolved nested name is still ambiguous. Open to ideas.

Two smaller things:
- The checklist ticks "Documentation updated" but there's no docs change — the new `-ModuleName 'Root/Nested'` syntax should get a line in the Mock / `InModuleScope` docs.
- No milestone set — is this meant for 6.0.0?

Tests I pushed are green (InModuleScope 39, Mock 252). 🙂

nohwnd:
@wetwicky  if you want to have another look.

nohwnd:
The change requested status here might be confusing, it is the only way to prevent this from merging before you have another look that the changes I added are still following what you wanted to do. No changes needed if you are happy @wetwicky .

nohwnd (approved):

PR #2708, opened 2026-06-11
Collect-Input: empty pipeline unwraps to $null for value assertions

nohwnd:
`docs/assertion-types.md` says a value assertion should treat `@()` from the pipeline the same as `$null` (since there is no scalar to unwrap to). The existing TODO on line 46 flagged that we did not actually match that — `Collect-Input` returned `@()` for both value and collection assertions, which made these fail:

```powershell
@() | Should-Be -Expected $null
@() | Should-BeNull
```

and made this wrongly pass:

```powershell
@() | Should-NotBeNull
```

`Collect-Input` now returns `$null` when `-UnrollInput` is set and the pipeline collected no items. Collection assertions are unchanged: they still see `@()`.

Updated the `Should-BeNull` test that previously asserted the buggy "fails" behavior to assert the documented "passes" behavior, and added matching coverage on `Should-Be`, `Should-NotBeNull`, and `Collect-Input` itself. Dropped the TODO and a stray code fence in `assertion-types.md`.

Verified with the full Pester suite: 2174 passed, 0 failed.

Follow-up to #2707.

### DIFF
--- a/docs/assertion-types.md
+++ b/docs/assertion-types.md
@@ -43,13 +43,11 @@ Another special case is `@()`. A value assertion will handle it as `$null`, but 
 1 | Should-Be -Expected 1
 @(1) | Should-Be -Expected 1
 $null | Should-Be -Expected $null
-@() | Should-Be -Expected $null #< --- TODO: this is not the case right now, we special case this as empty array, but is that correct? it does not play well with the value and collection assertion, and we special case it just because we can.
-# $null | will give $local:input -> $null , and @() | will give $local:input -> @(), is that distinction important when we know that we will only check against values?
+@() | Should-Be -Expected $null
 
 # This fails, because -Expected does not allow collections.
 @() | Should-Be -Expected @()
-
-
+```
 
 ```powershell
 # Should-BeCollection is a collection assertion:
--- a/src/functions/assert/Common/Collect-Input.ps1
+++ b/src/functions/assert/Common/Collect-Input.ps1
@@ -22,14 +22,22 @@
     if ($IsPipelineInput) {
         # We are called like this: 1 | Assert-Equal -Expected 1, we will get $local:Input in $PipelineInput and $true in $IsPipelineInput (coming from $MyInvocation.ExpectingInput).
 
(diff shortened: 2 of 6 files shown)


docs/assertion-types.md
```diff
@@ -43,13 +43,11 @@ Another special case is `@()`. A value assertion will handle it as `$null`, but
 1 | Should-Be -Expected 1
 @(1) | Should-Be -Expected 1
 $null | Should-Be -Expected $null
-@() | Should-Be -Expected $null #< --- TODO: this is not the case right now, we special case this as empty array, but is that correct? it does not play well with the value and collection assertion, and we special case it just because we can.
-# $null | will give $local:input -> $null , and @() | will give $local:input -> @(), is that distinction important when we know that we will only check against values?
+@() | Should-Be -Expected $null
```
nohwnd, line 46: also what about null in array? @($null), or ,$null or @(@($null))

nohwnd: Single-item pipeline (PowerShell collapses these to one `$null` item, so they all unwrap to `$null` and pass):

```powershell
@($null)        | Should-BeNull            # passes
,$null          | Should-BeNull            # passes
@(@($null))     | Should-BeNull            # passes  (@() unwraps the outer single-item array)
@($null)        | Should-Be -Expected $null # passes
@($null)        | Should-NotBeNull         # fails (unwrap → $null)
```

Multi-item pipeline (stays as a collection, so it is not `$null`):

```powershell
@($null, $null) | Should-BeNull            # fails  ([Object[]] '@($null, $null)')
1, $null        | Should-BeNull            # fails
@($null, $null) | Should-NotBeNull         # passes
```

I.e. the rule is just "how many items did the pipeline actually carry": 0 or 1 → unwraps to `$null` for value assertions; ≥2 → stays a collection. The presence of `$null` elements in a multi-item array does not make the collection itself null.

Pushed `7dd3340` adding tests for all of these in `Should-Be`, `Should-BeNull`, `Should-NotBeNull`, and the underlying `Collect-Input`.

PR #2725, opened 2026-06-17
Add per-file discovery and run mode (fixes #2723)

### LINKED ISSUE #2723: per file  discovery & run
run discovery in file, then run, then discover, avoids issues with scoping for modules, and we can clean up more and faster.

nohwnd:
Adds a `Run.PerFileDiscovery` option (default `true`) that discovers and runs each test container one at a time instead of discovering all containers first, then running all.

**Why:** The current batch model (discover all → run all) accumulates module state and scope pollution across containers during discovery. Per-file mode isolates each container: discover it, run it, release it, move to the next.

**What changed:**

- New `Run.PerFileDiscovery` boolean option (default `true`)
- Refactored `Invoke-Test` into `Invoke-TestBatch` (original flow) and `Invoke-TestPerFile` (new flow)
- `Invoke-ContainerDiscovery` extracted as a single-container discovery helper
- Global plugin hooks (`DiscoveryStart`/`End`, `RunStart`/`End`) fire once around the full run
- Per-container hooks fire per container as before
- Block tree references released after each container run to reduce peak memory

**Fallback:** Set `Run.PerFileDiscovery = $false` to revert to the original batch behavior.

All 2285 tests pass in both modes.

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -1925,12 +1925,23 @@ function Invoke-Test {
     $state.PluginData = $PluginData
     $state.Configuration = $Configuration
 
-    # # TODO: this it potentially unreliable, because suppressed errors are written to Error as well. And the errors are captured only from the caller state. So let's use it only as a useful indicator during migration and see how it works in production code.
+    if ($PesterPreference.Run.PerFileDiscovery.Value) {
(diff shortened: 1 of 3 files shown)


nohwnd:
abandoning because I don't think it is worth it, no-one asked for it. and further complicates how test can run in pester.

PR #2736, opened 2026-06-23
Throw on duplicate BeforeAll/BeforeEach/AfterAll/AfterEach

### LINKED ISSUE #2706: Multiple BeforeAll/BeforeEach/AfterEach/AfterAll in the same container - only the last one runs
Found while migrating PowerShell/PowerShell tests to Pester 5 (PR PowerShell/PowerShell#27290). If a container has more than one BeforeAll (or BeforeEach / AfterEach / AfterAll), only the **last** one runs. The earlier ones are discovered and silently dropped.

Repro on 5.7.1:

```powershell
Describe "multi-hooks" {
    BeforeAll  { Write-Host "BA1" }
    BeforeAll  { Write-Host "BA2" }
    BeforeEach { Write-Host "  BE1" }
    BeforeEach { Write-Host "  BE2" }
    AfterEach  { Write-Host "  AE1" }
    AfterEach  { Write-Host "  AE2" }
    AfterAll   { Write-Host "AA1" }
    AfterAll   { Write-Host "AA2" }
    It "i" { Write-Host "    It" }
}
```

Prints `BA2`, `BE2`, `It`, `AE2`, `AA2`. The `*1` blocks never run.

Mechanism is plain assignment in [`src/Pester.Runtime.ps1`](https://github.com/pester/Pester/blob/main/src/Pester.Runtime.ps1#L783) - `New-OneTimeTestSetup` does `.CurrentBlock.OneTimeTestSetup = `, same for `New-OneTimeTestTeardown`, `New-EachTestSetup`, `New-EachTestTeardown`. Last writer wins. The `Block` type holds a single `ScriptBlock` per hook, not a list.

This bit a lot of files in the PowerShell repo migration - about -1255 failures across CI once the duplicate blocks were merged. The failure mode is silent: helper functions/variables defined in the earlier block simply aren't there at runtime, and the test fails with `CommandNotFoundException` or "argument is null" in `It`.

### Idea

Either:
1. Run all blocks of the same hook in source order (append to a list), or
2. Fail discovery with a clear error when more than one is found in the same container.

(2) is the smaller change and would have caught all the cases I just fixed. (1) matches what people coming from xUnit/NUnit-style multiple setup methods would expect, and it's what the current docs imply.

nohwnd:
Each block can only have one of each setup/teardown hook. Using multiple was never supported and could lead to unpredictable behavior since execution order wasn't guaranteed. Now throws a clear error during discovery.

- Added `` checks in `New-EachTestSetup`, `New-EachTestTeardown`, `New-OneTimeTestSetup`, `New-OneTimeTestTeardown`
- Fixed existing tests that had duplicate hooks (SetupTeardown.Tests.ps1, Output.Tests.ps1)
- Added 5 new tests for duplicate detection

Fixes #2706

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -755,6 +755,9 @@ function New-EachTestSetup {
     )
 
     if (Is-Discovery) {
+        if ($null -ne $state.CurrentBlock.EachTestSetup) {
+            throw "BeforeEach is already defined in this block. Each block can only have one BeforeEach. Combine the code into a single BeforeEach block."
+        }
         $state.CurrentBlock.EachTestSetup = $ScriptBlock
     }
 }
@@ -767,6 +770,9 @@ function New-EachTestTeardown {
     )
 
     if (Is-Discovery) {
+        if ($null -ne $state.CurrentBlock.EachTestTeardown) {
+            throw "AfterEach is already defined in this block. Each block can only have one AfterEach. Combine the code into a single AfterEach block."
+        }
         $state.CurrentBlock.EachTestTeardown = $ScriptBlock
     }
 }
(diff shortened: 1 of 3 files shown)


fflaten (approved):
Good catch 🚀

PR #2737, opened 2026-06-23
Support -Mandatory:$false in Should-HaveParameter

### LINKED ISSUE #2105: "Should -HaveParameter 'param' -Not -Mandatory" passes tests if parameter is not present

nohwnd:
The new `Should-HaveParameter` had no way to assert that a parameter exists but is *not* mandatory. `-Mandatory:$false` now covers that — the parameter must exist and must not be mandatory. `-HasArgumentCompleter:$false` works the same way.

- `Should-HaveParameter X` — must exist
- `Should-HaveParameter X -Mandatory` — must exist and be mandatory
- `Should-HaveParameter X -Mandatory:$false` — must exist and not be mandatory

`-Not` is unchanged: `Should -Not -HaveParameter X` asserts the parameter isn't there at all, so `-Not -Mandatory` does **not** mean "exists but not mandatory" — use `-Mandatory:$false` for that.

Fixes #2105

### DIFF
--- a/src/functions/assertions/HaveParameter.ps1
+++ b/src/functions/assertions/HaveParameter.ps1
@@ -262,15 +262,26 @@
     if ($buts.Count -eq 0) {
         # Parameter exists (in set if specified), assert remaining requirements
 
-        if ($Mandatory) {
+        if ($PSBoundParameters.ContainsKey('Mandatory')) {
             $testMandatory = $parameterAttributes | & $SafeCommands['Where-Object'] { $_.Mandatory }
-            $filters += "which is$(if ($Negate) {' not'}) mandatory"
 
-            if (-not $Negate -and -not $testMandatory) {
-                $buts += "it wasn't mandatory"
+            if ($Mandatory) {
+                $filters += "which is$(if ($Negate) {' not'}) mandatory"
+
+                if (-not $Negate -and -not $testMandatory) {
+                    $buts += "it wasn't mandatory"
+                }
+                elseif ($Negate -and $testMandatory) {
+                    $buts += 'it was mandatory'
+                }
             }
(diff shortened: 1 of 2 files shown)


fflaten:
👍 Include same fix for `-HasArgumentCompleter`?

Why leave the old assertion as-is? Related #2223

nohwnd:
Added `-HasArgumentCompleter:$false` support in 8028878, good catch!

As for fixing the old assertion too — I left it to @nohwnd to decide. The old `-Not` pathway has the early-return-on-missing-param bug, but changing it might break existing test suites that rely on the current behavior (even if it's arguably wrong). Happy to fix it if you both agree.

PR #2740, opened 2026-06-24
Expand only template items in test and block names (fix code injection in names)

### LINKED ISSUE #2044: Using ` (backtick) at the end of a test fixture name breaks something internally in Pester.

PowerShell's normal parsing of a test ignores `` ` `` characters in single-quoted strings. Pester seems to re-parse certain parts, which turns them into double-quoted strings, without properly escaping characters like this.

## Describe your environment

```
Pester version     : 5.2.0 /Users/joelfrancis/.local/share/powershell/Modules/Pester/5.2.0/Pester.psm1                  PowerShell version : 7.1.0-rc.2                                                                                         OS version         : Unix 11.0.0
```

## Steps to reproduce

Simple reproducible test file:

```powershell
Describe 'test' {
    It 'doesn''t break with backticks `' { 
        # this can be empty and it still breaks
    }
}
```

## Expected Behavior

The content of the parsed string should probably(?) not be fully re-parsed with PowerShell's parser, this could lead to all sorts of odd behaviour, require double escaping, accidental code injection, etc.

From testing it seems like _some_ additional escaping is done, but it's not careful enough by the looks of it, backticks are currently not escaped unlike some other characters.

## Current Behavior

```

Discovery finished in 110ms.
Running tests.
[-] test.doesn't break with `backticks` 39ms (22ms|17ms)
 ParseException: At line:1 char:1
 + "doesn't break with `backticks`"
 + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The string is missing the terminator: ".
 MethodInvocationException: Exception calling "Create" with "1" argument(s): "At line:1 char:1
 + "doesn't break with `backticks`"
 + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The string is missing the terminator: "."
```

Pester is probably doing something with the parser to expand its `<TestCase>` insertions. There are some alternatives to fully re-parsing a string, for example:

- Prefix a `$` to the string and then use `$ExecutionContext.InvokeCommand.ExpandString($str)` to expand the value just for that segment of the string, rather than the whole string.
- Opt for a regex replacement instead of invoking PowerShell's parser directly.

In my opinion, I think _fully_ re-parsing the string is hazardous at best and fragile in ways that aren't easily avoidable. Manually escaping things is a start, but as this issue shows, it's easy to miss edge cases. I suspect it's possible to bypass Pester's escaping currently by adding extra backticks, which may lead to code injection.

nohwnd:
Test and block names are expanded by turning `<template>` items into subexpressions. The old code did this by re-parsing the whole name as a double-quoted string, so a literal backtick, `$` or `"` in a name could break parsing (``It 'has `backticks`'`` threw a `ParseException`) or be used to bypass the escaping and inject code.

Now only the `<template>` items become `$(subexpressions)` and every other character is escaped so the literal text stays inert. `` `< `` and `` `> `` still escape a literal angle bracket. Applied at both the block and test expansion sites.

Per-name expansion is ~0.07 ms vs ~0.02 ms for the old whole-string parse, well under the `ExpandString` cost noted on the issue. The Runtime and RSpec suites pass, plus new tests for the backtick break and an injection attempt.

Fixes #2044

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -388,7 +388,19 @@ function Invoke-Block ($previousBlock) {
                                 # avoid using variables so we don't run into conflicts
                                 $sb = {
 
-                                    $____Pester.CurrentBlock.ExpandedName = if ($____Pester.CurrentBlock.Name -like "*<*") { & ([ScriptBlock]::Create(('"' + ($____Pester.CurrentBlock.Name -replace '\$', '`$' -replace '"', '`"' -replace '(?<!`)<([^>^`]+)>', '$$($$$1)') + '"'))) } else { $____Pester.CurrentBlock.Name }
+                                    $____Pester.CurrentBlock.ExpandedName = if ($____Pester.CurrentBlock.Name -like "*<*") {
+                                        $____PesterExpandedName = [System.Text.StringBuilder]::new($____Pester.CurrentBlock.Name.Length)
+                                        $____PesterExpandedPos = 0
+                                        foreach ($____PesterExpandedMatch in [regex]::Matches($____Pester.CurrentBlock.Name, '(?<!`)<([^>`]+)>|`([<>])|([`"$])')) {
+                                            $null = $____PesterExpandedName.Append($____Pester.CurrentBlock.Name.Substring($____PesterExpandedPos, $____PesterExpandedMatch.Index - $____PesterExpandedPos))
+                                            if ($____PesterExpandedMatch.Groups[1].Success) { $null = $____PesterExpandedName.Append('$($').Append($____PesterExpandedMatch.Groups[1].Value).Append(')') }
+                                            elseif ($____PesterExpandedMatch.Groups[2].Success) { $null = $____PesterExpandedName.Append($____PesterExpandedMatch.Groups[2].Value) }
+                                            else { $null = $____PesterExpandedName.Append('`').Append($____PesterExpandedMatch.Groups[3].Value) }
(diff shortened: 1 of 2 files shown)


src/Pester.Runtime.ps1
```diff
@@ -648,7 +660,21 @@ function Invoke-TestItem {
...
-                                & ([ScriptBlock]::Create(('"' + ($____Pester.CurrentTest.Name -replace '\$', '`$' -replace '"', '`"' -replace '(?<!`)<([^>^`]+)>', '$$($$$1)') + '"')))
+                                # Expand only the <template> items; escape every other character so literal
+                                # backticks, $ and " stay inert and cannot break parsing or inject code (#2044).
+                                # `< and `> escape a literal angle bracket. Locals are $____Pester-prefixed so a
+                                # <template> that references a user variable is not shadowed.
+                                $____PesterExpandedName = [System.Text.StringBuilder]::new($____Pester.CurrentTest.Name.Length)
+                                $____PesterExpandedPos = 0
+                                foreach ($____PesterExpandedMatch in [regex]::Matches($____Pester.CurrentTest.Name, '(?<!`)<([^>`]+)>|`([<>])|([`"$])')) {
+                                    $null = $____PesterExpandedName.Append($____Pester.CurrentTest.Name.Substring($____PesterExpandedPos, $____PesterExpandedMatch.Index - $____PesterExpandedPos))
```
nohwnd: this runs in user scope and will inherit , we don't want that, these variables need to be $private:

nohwnd: $____Pester is fine, that is already established, but not those temporary variables.

PR #2746, opened 2026-06-25
Don't crash the run on stray output during a test run

### LINKED ISSUE #2655: Pester crashes due to native command output in `BeforeAll`

nohwnd:
Fix #2655

Same problem #2656 tried to fix, and like there I can't repro the actual leak. `winrm quickconfig` and `pwsh -help` in a `BeforeAll` don't reproduce it for me, on PowerShell 7 or Windows PowerShell, file or scriptblock container.

The crash itself is easy to pin down though. When something other than a `[Pester.Container]` ends up in `Invoke-Test`'s output - e.g. a native command writing to the success stream in a setup block that wasn't redirected to `$null` - the `$run.Containers.Add($i)` loop throws `Cannot find an overload for "Add"` and the whole run dies with that opaque error.

So instead of chasing the leak this stops feeding stray output into the typed list. `Split-RSpecResult` separates the containers from anything else, the containers go into the run as before, and the rest is dropped with a warning that points at the likely cause. Worst case you lose some stray output you didn't want anyway, instead of losing the whole run.

The alternative is to find and fix the leak at the source, but neither of us could pin it down. LMK if you'd rather preserve the stray output somewhere instead of just warning.

### DIFF
--- a/tst/Pester.RSpec.ts.ps1
+++ b/tst/Pester.RSpec.ts.ps1
@@ -3250,4 +3250,47 @@ i -PassThru:$PassThru {
             $r.Containers[0].Result | Verify-Equal 'Passed'
         }
     }
+
+    b 'Stray output during the run does not crash Pester (#2655)' {
+        t 'Split-RSpecResult keeps containers and collects stray output separately' {
+            # Get a real [Pester.Container] to mix with stray output
+            $real = Invoke-Pester -Configuration ([PesterConfiguration]@{
+                    Run    = @{ ScriptBlock = { Describe 'd' { It 'i' { $true | Should -Be $true } } }; PassThru = $true }
+                    Output = @{ Verbosity = 'None' }
+                })
+            $container = $real.Containers[0]
+
+            $split = & (Get-Module Pester) {
+                param($c)
+                # Native command output that leaks into the run pipeline (e.g. an unredirected
+                # winrm/net command in BeforeAll) arrives here as plain strings next to the container.
+                Split-RSpecResult -Result (@($c, 'WinRM service is already running on this machine.', $null))
+            } $container
+
+            $split.Containers.Count | Verify-Equal 1
+            ($split.Containers[0] -is [Pester.Container]) | Verify-True
+            $split.StrayOutput.Count | Verify-Equal 1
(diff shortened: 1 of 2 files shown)


tst/Pester.RSpec.ts.ps1
```diff
@@ -3308,4 +3250,47 @@ i -PassThru:$PassThru {
...
+            $threw = $false
+            try { ([Pester.Run]::Create()).Containers.Add('stray native output') } catch { $threw = $true }
+            $threw | Verify-True
+
+            # After splitting, only the container is added and the run is built without error.
+            $run = [Pester.Run]::Create()
+            $split = & (Get-Module Pester) { param($c) Split-RSpecResult -Result (@($c, 'stray native output')) } $container
+            foreach ($i in $split.Containers) { $run.Containers.Add($i) }
+            $run.Containers.Count | Verify-Equal 1
+        }
```
fflaten, line 3294: How is this different from the first test? Neither are testing that a real run with stray output.

PR #2748, opened 2026-06-25
Show the differing index when Should -Be compares arrays

### LINKED ISSUE #1990: Show difference index / arrow for array differences
image

Ideally comparing arrays with Should would focus on where the difference is like string, but that might be a different/new issue.

_Originally posted by @fflaten in https://github.com/pester/Pester/issues/1668#issuecomment-860111171_

nohwnd:
Fix #1990

When two arrays have the same length but a different value, the failure now points at the first differing index. It prints the expected and actual element at that index, and keeps the existing length mismatch and scalar messages unchanged. Added regression tests for `Should -Be` and `Should -BeExactly`.

### DIFF
--- a/tst/functions/assertions/Be.Tests.ps1
+++ b/tst/functions/assertions/Be.Tests.ps1
@@ -23,6 +23,10 @@ InPesterModuleScope {
             $array | Should -EQ $array
         }
 
+        It 'Passes for equal arrays with the same length' {
+            @(1, 2, 3) | Should -Be @(1, 2, 3)
+        }
+
         It 'Compares arrays with correct case-insensitive behavior' {
             $string = 'I am a string'
             $array = @(1, 2, 3, 4, $string)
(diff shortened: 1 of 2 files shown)


tst/functions/assertions/Be.Tests.ps1
```diff
@@ -146,8 +150,8 @@ InPesterModuleScope {
             ShouldBeFailureMessage (, (1, 2, 3)) (1, 2, 3) -Because 'reason' | Verify-Equal "Expected a collection @(1, 2, 3) with length 3, because reason, but got a collection @(@(1, 2, 3)) with length 1."
         }
 
-        It "Keeps the plain message for collections of equal length" {
-            ShouldBeFailureMessage (1, 2, 3) (1, 2, 4) | Verify-Equal "Expected @(1, 2, 4), but got @(1, 2, 3)."
+        It "Outputs the differing index for collections of equal length" {
+            ShouldBeFailureMessage (1, 2, 3) (1, 2, 4) | Verify-Equal "Expected collections to be the same, but they were different.`nCollection lengths are both 3.`nCollections differ at index 2.`nExpected: 4`nBut was:  3"
```
nohwnd: - format with newlines so we can read it (don't forget the normalize the output so it compares on linux as well, plenty examples around)

- we no longer get to see the content of the array, not good.

nohwnd: Done in 5a1d6fe - the message now shows the full collections (like the string diff does) plus the index, so the array contents are back. Joined with 
 so it compares the same on Linux.

Issue #2452, opened 2024-05-13
TestResults.OutputEncoding option ignored


fflaten:
### Checklist

- [X] Issue has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

`TestResults.OutputEncoding` configuration option is ignored. XML delcaration is hardcoded to utf-8 and file is written in default encoding.

### Expected Behavior

Configuration option should be properly typed to valid encodings and be effective.

### Steps To Reproduce

```powershell
$c = New-PesterConfiguration
$c.Run.ScriptBlock = {
    Describe 'd' {
        It 'i' { }
    }
}
$c.TestResult.Enabled = $true
$c.TestResult.OutputFormat = 'NUnit3'
$c.TestResult.OutputEncoding = 'utf-16'
$c.TestResult.TestSuiteName = 'TestResult.OutputEncoding Demo'
$c.TestResult.OutputPath = "./demoOutputEncoding.xml"
$c.Output.Verbosity = 'Detailed'
Invoke-Pester -Configuration $c
```

demoOutputEncoding.xml is written in UTF-8 and incldues `encoding="utf-6"`.

### Describe your environment

Pester version : 5.6.0-beta1 /workspaces/Pester/bin/Pester.psm1
PowerShell version : 7.4.1
OS version : Unix 5.15.133.1

### Possible Solution?

Improve type for configuration option to only allow valid encodings. Then pass it to `Encoding`-option here:

https://github.com/pester/Pester/blob/93c661020d666c7c07577c46b8ad756b8aa74aec/src/functions/TestResults.ps1#L169-L172

Issue #2451, opened 2024-05-13
Invoke-Pester breaks on invalid CodeCoverage.OutputEncoding format


fflaten:
### Checklist

- [X] Issue has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

Providing a invalid encoding to `CodeCoverage.OutputEncoding` throws an exception at the end of `Invoke-Pester`.

### Expected Behavior

Properly typed configuration option or early validation in `Invoke-Pester` to avoid breaking the run at the end.

```console
Processing code coverage result.
System.Management.Automation.RuntimeException: Invoking step End failed:
Result 1 - Error 1:Out-File: /workspaces/Pester/src/functions/Coverage.Plugin.ps1:194:88
Line |
 194 |  … h -Encoding $PesterPreference.CodeCoverage.OutputEncoding.Value -Forc …
     |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot process argument transformation on parameter 'Encoding'. 'utf-15' is not a supported encoding name. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider
     | method. (Parameter 'name')

at <ScriptBlock>, /workspaces/Pester/src/functions/Coverage.Plugin.ps1: line 194
at Invoke-PluginStep, /workspaces/Pester/src/Pester.Runtime.ps1: line 1275
at Invoke-Pester<End>, /workspaces/Pester/src/Main.ps1: line 853
at <ScriptBlock>, <No file>: line 12

at Assert-Success, /workspaces/Pester/src/Pester.Runtime.ps1: line 1332
at Invoke-PluginStep, /workspaces/Pester/src/Pester.Runtime.ps1: line 1294
at Invoke-Pester<End>, /workspaces/Pester/src/Main.ps1: line 853
at <ScriptBlock>, <No file>: line 12
```

### Steps To Reproduce

```powershell
# Executed in pester-repo
$c = New-PesterConfiguration
$c.Run.Path = './tst/functions/assertions'
$c.Run.ExcludePath = '*/demo/*', '*/examples/*', '*/testProjects/*', '*/Pester.Tests.ps1'
$c.Output.Verbosity = 'Detailed'
[...]

PR #2749, opened 2026-06-25
Write the code coverage report in the configured encoding and match the xml declaration

### LINKED ISSUE #2450: CodeCoverage.OutputEncoding ignored in xml encoding-attribute
### Checklist

### What is the issue?

CodeCoverage XML reports always report `encoding="utf-8"` regardless of `CodeCoverage.OutputEncoding` configuration option. The file itself is written using the selected encoding.

Related comment: https://github.com/pester/Pester/pull/2298#discussion_r1107700563

### Expected Behavior

XML report should report the correct encoding in xml declaration.

### Steps To Reproduce

```powershell
# Executed in pester-repo
$c = New-PesterConfiguration
$c.Run.Path = './tst/functions/assertions'
$c.Run.ExcludePath = '*/demo/*', '*/examples/*', '*/testProjects/*', '*/Pester.Tests.ps1'
$c.Output.Verbosity = 'Detailed'
$c.CodeCoverage.Enabled = $true
$c.CodeCoverage.Path = './src/functions'
$c.CodeCoverage.UseBreakpoints = $false
$c.CodeCoverage.OutputFormat = 'JaCoCo'
$c.CodeCoverage.OutputEncoding = 'unicode' #utf-16
Invoke-Pester -Configuration $c
```

Generated coverage.xml is UTF-16 LE, but says `<?xml version="1.0" encoding="UTF-8" standalone="no"?>`

### Describe your environment

Pester version     : 5.6.0-beta1 /workspaces/Pester/bin/Pester.psm1                                                     
PowerShell version : 7.4.1
OS version         : Unix 5.15.133.1

### Possible Solution?

Replace hard-coded declaration value here:
https://github.com/pester/Pester/blob/93c661020d666c7c07577c46b8ad756b8aa74aec/src/functions/Coverage.ps1#L924

Should also improve type for configuration option or limit strings values to valid encodings. XML declaration and `Out-File` accepts different encodings.

nohwnd:
Fix #2450

The code coverage report is built by casting the JaCoCo/Cobertura string to `[xml]`, which carries a hard-coded `encoding="UTF-8"` declaration, and then written through a `StringWriter` and `Out-File -Encoding`. So the bytes on disk used `CodeCoverage.OutputEncoding` but the declaration always said utf-8 - they didn't match.

Write the report straight to the file in the configured encoding and set the declaration to match, the same way `TestResult.OutputEncoding` now works (#2452). The encoding-name to `[Text.Encoding]` mapping moved to `Get-OutputEncodingFromName` in `Pester.Utility` so both report writers share it; an invalid encoding still falls back to utf8 and warns (#2451), now up front instead of after a failed write.

### DIFF
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -456,35 +456,3 @@ function View-Flat {
         $tests
     }
 }
-
-function Get-OutputEncodingFromName {
-    # Converts a PowerShell-style encoding name (the values accepted by Out-File -Encoding, e.g. 'UTF8',
-    # 'UTF8BOM', 'Unicode', 'UTF32', 'ASCII') or a .NET web name (e.g. 'utf-16') to a [System.Text.Encoding]
-    # instance for use with XmlWriterSettings.Encoding, so the xml encoding-declaration and the bytes on disk
-    # match. Falls back to UTF-8 (with BOM, the historical default) and warns when the value is empty or not a
-    # recognized encoding. (#2452, #2450)
-    param (
-        [string] $Encoding,
-        [string] $OptionName = 'TestResult.OutputEncoding'
-    )
-
-    switch -Regex ($Encoding) {
-        '^\s*$' { return [System.Text.UTF8Encoding]::new($true) }
-        '(?i)^utf-?8(-?bom)?$' { return [System.Text.UTF8Encoding]::new($true) }
-        '(?i)^utf-?8-?nobom$' { return [System.Text.UTF8Encoding]::new($false) }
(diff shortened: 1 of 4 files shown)


src/Pester.Utility.ps1
```diff
@@ -456,3 +456,35 @@ function View-Flat {
         $tests
     }
 }
+
+function Get-OutputEncodingFromName {
```
nohwnd: looks like the same function, it just moved? put it back

nohwnd: Put it back in 5bd7268 - kept it in TestResults.ps1 and just added an optional OptionName param so the coverage caller gets the right name in the warning. Utility.ps1 is untouched now.

PR #2751, opened 2026-06-25
Don't require a verifiable mock that an inner-scope default mock shadows

### LINKED ISSUE #2672: Verifiable Mock overridden in It block causes Should -InvokeVerifiable to fail
### Checklist

### What is the issue?

I’m seeing what looks like an issue with how -Verifiable mocks behave when they are overridden in a narrower scope.

Summary

If a mock is defined as -Verifiable in an outer scope (e.g., Describe/Context) and then a different mock for the same command is defined in an inner scope (e.g., inside an It block) also marked -Verifiable, the inner mock is correctly used for the call, but Should -InvokeVerifiable fails.

If I remove -Verifiable from the outer mock, then overriding the mock in It works and Should -InvokeVerifiable passes.

This makes it difficult to have a default verifiable mock for most tests, while overriding it for specific tests.

**Workaround**

If I remove -Verifiable from the outer mock (leave it only on the inner mock), the override test passes.

Thank you.

### Expected Behavior

When a mock is overridden in a narrower scope:

The overridden (inner) verifiable mock should be the one that needs to be “verified”.

The outer verifiable mock should either be considered shadowed/ignored for verification purposes, or automatically treated as not applicable in that It run.

Therefore, Should -InvokeVerifiable should pass when the effective mock(s) for the test were invoked.

**Actual behavior**

The call goes to the overridden mock in It (confirmed by return value / debug output).

Should -InvokeVerifiable fails anyway, apparently because the outer -Verifiable mock is still tracked as “not invoked”, even though it was overridden and could never be invoked in that test.

### Steps To Reproduce

**Minimal reproduction**
Describe 'Verifiable mock overriding' {

    BeforeAll {
        function Get-Data { 'real' }
    }

    Context 'default mock in outer scope' {

        It 'uses default and passes' {
            Get-Data | Should -Be 'outer-mock'
            Should -InvokeVerifiable
        }

        It 'overrides mock inside It but InvokeVerifiable fails' {

            # Override for this specific test
            Mock Get-Data { 'inner-mock' } -Verifiable

            # Confirm the override is actually used
            Get-Data | Should -Be 'inner-mock'

            # This fails (even though the inner verifiable mock was invoked)
            Should -InvokeVerifiable
        }
    }
}

### Describe your environment

_No response_

### Possible Solution?

_No response_

nohwnd:
Fix #2672

A mock marked `-Verifiable` in an outer scope (Describe/Context) that you then override with another mock for the same command in an inner scope (It) could never pass `Should -InvokeVerifiable`: the inner mock handled the call, but `Get-VerifiableBehaviors` still collected the outer one and reported it as never invoked.

`Get-VerifiableBehaviors` now walks the scopes from the test outwards and drops a verifiable behavior when a closer scope has a default mock (one without a `-ParameterFilter`) for the same command - that default catches every call, so the outer behavior can't run and shouldn't be required. A verifiable mock that isn't shadowed like that is still required, same as before.

### DIFF
--- a/src/functions/Pester.SessionState.Mock.ps1
+++ b/src/functions/Pester.SessionState.Mock.ps1
@@ -366,51 +366,63 @@ function Get-VerifiableBehaviors {
     $currentTest = Get-CurrentTest
     $inTest = $null -ne $currentTest
 
-    $behaviors = [System.Collections.Generic.List[Object]]@()
+    # Collect the mock-behavior tables from the closest scope (the test) outwards to the root block, so we can
+    # walk them in precedence order. A default mock (one without a -ParameterFilter) in a closer scope shadows
+    # every behavior for the same command in an outer scope, because the closer default catches all calls and
+    # the outer behaviors can never be invoked. A verifiable behavior that is shadowed like that must not be
+    # required by Should -InvokeVerifiable, otherwise overriding a verifiable mock in an inner scope always
+    # fails verification. (#2672)
+    $behaviorTables = [System.Collections.Generic.List[Object]]@()
     if ($inTest) {
-        if ($PesterPreference.Debug.WriteDebugMessages.Value) {
-            Write-PesterDebugMessage -Scope Mock "We are in a test. Finding all behaviors in this test."
-        }
-        $allBehaviors = $currentTest.PluginData.Mock.Behaviors.Values
-        if ($null -ne $allBehaviors -and $allBehaviors.Count -gt 0) {
-            # all behaviors for all commands
-            foreach ($commandBehaviors in $allBehaviors) {
-                if ($null -ne $commandBehaviors -and $commandBehaviors.Count -gt 0) {
-                    # all behaviors for single command
-                    foreach ($behavior in $commandBehaviors) {
-                        if ($behavior.Verifiable) {
-                            $behaviors.Add($behavior)
-                        }
-                    }
-                }
-            }
-        }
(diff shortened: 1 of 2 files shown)


nohwnd:
Picked this back up. Merged `main` in, no conflicts, and it still does what it says.

The repro in #2672 is missing the outer `Mock` line, so I reconstructed it and checked several shapes against released 6.1.0 and against this branch:

| | 6.1.0 | this branch |
|---|---|---|
| outer verifiable in `Context` `BeforeAll`, inner verifiable in `It` | fails | passes |
| outer verifiable, inner not verifiable | fails | passes |
| outer verifiable at `Describe` level, inner default | fails | passes |

One shape I got wrong on the first try, worth writing down so nobody else chases it: an outer verifiable mock **with** a `-ParameterFilter` and an inner default mock. That looks like it should be dropped too, but the call returns `outer`, not `inner`, because a parameterized behavior wins over a default one regardless of which scope it came from. So the outer mock does run there and requiring it is correct. Not a gap.

Also worth knowing: the bug only shows when the outer verifiable mock is never invoked before the overriding test. If an earlier `It` in the same `Context` already called it, the behavior is marked invoked and the override test passes even on 6.1.0. That is why a casual repro can look fine.

`tst/Pester.Mock.RSpec.ts.ps1` 40/40 and `tst/functions/Mock.Tests.ps1` 263/263 green.

Marking ready for review.

🤖

PR #2756, opened 2026-06-27
Count dictionary entries in Should -HaveCount

### LINKED ISSUE #1000: -HaveCount assertion operator does not work when using strong typing on $null

Pester 4.2.0 introduced a new `-HasCount` assertion operator. When migrating existing tests, by replacing `$variable.Count | Should -Be $expectedCount` to `$variable | Should -HaveCount $expectedCount`, I found edge cases where this does not work. I created Pester tests to demonstrate this behaviour:

````powershell
Describe "Pesters -HaveCount operator should not fail on `$null object that has Count property" {
    It "when not using strong typing" {
        $aliases = Get-Alias | Select-Object -First 0
        $aliases.Count | Should -Be 0
        $aliases | Should -HaveCount 0
    }

    It "when LHS is casted to [System.Array]" {
        [System.Array] $aliases = Get-Alias | Select-Object -First 0
        $aliases.Count | Should -Be 0
        $aliases | Should -HaveCount 0
    }

    It "when RHS is casted to [System.Array]" {
        $aliases = [System.Array] (Get-Alias | Select-Object -First 0)
        $aliases.Count | Should -Be 0
        $aliases | Should -HaveCount 0
    }

    It "when LHS and RHS are casted to [System.Array]" {
        $aliases = [System.Array] (Get-Alias | Select-Object -First 0)
        $aliases.Count | Should -Be 0
        $aliases | Should -HaveCount 0
    }
}
````

Pester version : 4.3.1
PowerShell version : 5.1.16299.248
PowerShell version : 6.0.1
PowerShell version : 6.1.0-preview.589

## 3. Expected Behavior

All 4 tests above should pass

## 4.Current Behavior

The first test passed but not the other 3. By debugging it, the objects seem to be of the same type (`$null`), therefore I cannot explain why Pester behaves differently. PowerShell has `.Count` property on `$null` by design.

````
  Describing Pesters -HaveCount operator should not fail on $null object that has Count property
    [+] when not using strong typing 387ms
    [-] when LHS is casted to [System.Array] 132ms
      Expected an empty collection, but got collection with size 1 @().
      11:         $aliases | Should -HaveCount 0
      at <ScriptBlock>, C:\Users\cberg\git\bla.tests.ps1: line 11
    [-] when RHS is casted to [System.Array] 105ms
      Expected an empty collection, but got collection with size 1 @().
      17:         $aliases | Should -HaveCount 0
      at <ScriptBlock>, C:\Users\cberg\git\bla.tests.ps1: line 17
    [-] when LHS and RHS are casted to [System.Array] 37ms
      Expected an empty collection, but got collection with size 1 @().
      23:         $aliases | Should -HaveCount 0
      at <ScriptBlock>, C:\Users\cberg\git\bla.tests.ps1: line 23
Tests completed in 662ms
Tests Passed: 1, Failed: 3, Skipped: 0, Pending: 0, Inconclusive: 0
````

## 5. Possible Solution

Document more clearly on what types the new `HasCount` assertion operator works but to me it rather looks like a Pester internal bug that needs to be resolved.

## 6. Context

As described in paragraph: migrating to new `-HasCount` operator, i.e. not a blocking issue.

### LINKED ISSUE #1234: Problem using HaveCount on [hashtable]

nohwnd:
A hashtable or dictionary passes through the pipeline as a single object, so `Should -HaveCount` only ever saw one item and reported size 1. Now a single piped item is inspected: dictionaries report their entry count, a lone `$null` counts as empty, and any other single object stays 1.

Only real `IDictionary` types are looked into. The synthesized `.Count` that PowerShell puts on scalars and `$null` isn't available under `Set-StrictMode` (it throws), so we never read it there - scalars are just 1 and the collected array's own `.Count` covers everything else.

Trade-off, as discussed in #1234: an array holding a single dictionary now counts the dictionary's entries, so `@(@{ a = 1; b = 2 }) | Should -HaveCount 2` rather than `1`.

Closes #1234
Closes #1000

### DIFF
--- a/src/functions/assertions/HaveCount.ps1
+++ b/src/functions/assertions/HaveCount.ps1
@@ -8,12 +8,38 @@
 
     This test passes, because it expected three objects, and received three.
     This is like running `@(1,2,3).Count` in PowerShell.
+
+    .EXAMPLE
+    @{ Name = 'Jakub'; Age = 30 } | Should -HaveCount 2
+
+    This test passes. A hashtable (or any dictionary) passes through the pipeline as a
+    single object, so HaveCount looks inside it and counts its entries.
     #>
     if ($ExpectedValue -lt 0) {
         throw [ArgumentException]"Expected collection size must be greater than or equal to 0."
     }
     $count = if ($null -eq $ActualValue) {
         0
+    }
+    elseif ($ActualValue.Count -eq 1) {
+        # A single object reached us. The pipeline does not enumerate dictionaries
+        # and hashtables, so they arrive as one item - look inside and count their
+        # entries instead of reporting 1. A lone $null is treated as empty. Every
+        # other single object counts as 1.
+        #
+        # We must not read .Count off the inner item unless it is a real collection.
+        # PowerShell synthesizes .Count on scalars and $null, but that synthesized
+        # member is not available under Set-StrictMode and would throw.
+        $single = $ActualValue[0]
+        if ($null -eq $single) {
+            0
+        }
+        elseif ($single -is [System.Collections.IDictionary]) {
+            $single.Count
+        }
+        else {
+            1
+        }
     }
     else {
         $ActualValue.Count
--- a/tst/functions/assertions/HaveCount.Tests.ps1
+++ b/tst/functions/assertions/HaveCount.Tests.ps1
@@ -44,6 +44,76 @@ InPesterModuleScope {
         }
     }
 
+    Describe "Should -HaveCount with dictionaries" {
+        # The pipeline does not enumerate dictionaries and hashtables, so they arrive
+        # as a single item. HaveCount looks inside and counts their entries. (#1234, #1200)
+
+        It "counts the entries of a hashtable" {
+            @{ Property1 = '1'; Property2 = '3' } | Should -HaveCount 2
+        }
+
+        It "counts an empty hashtable as 0" {
+            @{} | Should -HaveCount 0
+        }
+
+        It "counts the entries of an ordered dictionary" {
+            ([ordered]@{ a = 1; b = 2; c = 3 }) | Should -HaveCount 3
+        }
+
+        It "counts the entries of a generic dictionary" {
+            $dictionary = [System.Collections.Generic.Dictionary[string, string]]::new()
+            $dictionary.Add('a', '1')
+            $dictionary.Add('b', '2')
+            $dictionary | Should -HaveCount 2
+        }
+
+        It "counts an empty generic dictionary as 0" {
+            [System.Collections.Generic.Dictionary[string, string]]::new() | Should -HaveCount 0
+        }
+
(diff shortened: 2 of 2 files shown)


nohwnd:
@fflaten  wdyt? worth changing the behavior on the older Should operators? Will add assertions specific to hashtables / dictionaries in the the new Should-*

nohwnd:
merged so I can pilot on some select bigger repos.

fflaten:
> @fflaten wdyt? worth changing the behavior on the older Should operators? Will add assertions specific to hashtables / dictionaries in the the new Should-*

Not a fan personally. I understand the ask, but it's easier to recommend `@{ a = 1; b = 2 }.Keys | Should -HaveCount 2` for dictionaries (or the new assertion) vs `@().Count | Should -Be ..` for all collections or *when* the items are dictionary type.

Either way, we need to document the broken case 🙂

kborowinski:
@nohwnd @fflaten I was just bitten by that with tests that were working on 5.7.1. This is breaking change, and should be mentioned in the breaking change section for `6.0.0-rc1` release _(or I just missed that?)._

PR #2772, opened 2026-06-27
Add Run.BeforeInvoke bootstrap for Invoke-Pester

nohwnd:
## Summary

Adds a `Run.BeforeInvoke` option that runs optional bootstrap code **in the caller's scope, as soon as `Invoke-Pester` starts** — before the caller's `$PesterPreference` is read and before discovery. Use it to import dependencies and to provide configuration by defining or modifying `$PesterPreference`, which Pester then picks up as the caller preference.

This mirrors the container-level `BeforeContainer` convention, but for the top-level invocation itself.

## How it works

Two sources, in priority order:

1. **`Run.BeforeInvoke`** config option (`scriptblock[]`). When set, it wins and runs as-is.
2. **Convention file** — the first `Pester.BeforeInvoke.ps1` found when walking up from each `Run.Path` toward `Run.RepoRoot` is dot-sourced (deduped, in order, never escaping the repo root).

Each scriptblock is bound to the caller's `SessionState` and **dot-sourced**, so imported modules, defined functions, and `$PesterPreference` land in the caller's scope where `Invoke-Pester` reads them next. The bootstrap runs for **top-level runs only**, so nested Pester-in-Pester does not re-run it.

## Changes

- **C#** — `BeforeInvoke` `ScriptBlockArrayOption` on `RunConfiguration` (mirrors `RepoRoot`/`ScriptBlock`).
- **`src/functions/Pester.BeforeInvoke.ps1`** — `Resolve-PesterBeforeInvoke` + `Invoke-PesterBeforeInvoke`.
- **`Main.ps1`** — resolves a preliminary preference and runs the bootstrap before the caller-preference read.
- **Help** — generated `BeforeInvoke` entry in `about_PesterConfiguration`.
- **Tests** — `tst/Pester.BeforeInvoke.ts.ps1` (9 P-tests). Full suite green.

## Open design question (reason this is a draft)

`BeforeInvoke` currently runs **once**, in the orchestrating `Invoke-Pester` call. For a **parallel run**, test containers execute in separate runspaces that would **not** have re-run the bootstrap — so dependency setup done here would be missing in the workers. Configuration (`$PesterPreference`) is inherently a once-per-invocation concern and is fine, but dependency provisioning may need to also happen per-container (e.g. via `BeforeContainer`) for parallel scenarios. Want to settle this before merging.

</body>
<parameter name="draft">true

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -488,6 +488,29 @@ function Invoke-Pester {
                 $Configuration = . Convert-PesterSimpleParameterSet -BoundParameters $PSBoundParameters
             }
 
+            # Run.BeforeInvoke: as soon as Invoke-Pester starts (top-level runs only), run optional
+            # bootstrap code in the caller's scope, before the caller's $PesterPreference is read
+            # below and before discovery. Use it to import dependencies and provide configuration by
+            # defining or modifying $PesterPreference, which is then picked up as the caller preference.
+            #
+            # A preliminary preference is resolved first so the bootstrap has somewhere to discover
+            # itself from (Run.Path/RepoRoot/BeforeInvoke) and so error reporting still works if the
+            # bootstrap throws. It is replaced with the final caller-merged preference right after.
+            if (-not $runningPesterInPester) {
+                $beforeInvokeCallerPreference = [PesterConfiguration] $PSCmdlet.SessionState.PSVariable.GetValue("PesterPreference")
+                [PesterConfiguration] $PesterPreference = if ($null -ne $beforeInvokeCallerPreference -and $null -ne $Configuration) {
+                    [PesterConfiguration]::Merge($beforeInvokeCallerPreference, $Configuration)
+                }
+                elseif ($null -ne $Configuration) {
+                    [PesterConfiguration]::Merge([PesterConfiguration]::Default, $Configuration)
+                }
+                else {
+                    [PesterConfiguration]::Default
+                }
+
+                Invoke-PesterBeforeInvoke -Configuration $PesterPreference -SessionState $PSCmdlet.SessionState
+            }
+
             # maybe -IgnorePesterPreference to avoid using $PesterPreference from the context
 
             $callerPreference = [PesterConfiguration] $PSCmdlet.SessionState.PSVariable.GetValue("PesterPreference")
--- a/src/csharp/Pester/RunConfiguration.cs
+++ b/src/csharp/Pester/RunConfiguration.cs
@@ -35,6 +35,7 @@ namespace Pester
         private StringOption _skipRemainingOnFailure;
         private BoolOption _failOnNullOrEmptyForEach;
         private StringOption _repoRoot;
+        private ScriptBlockArrayOption _beforeInvoke;
 
         public static RunConfiguration Default { get { return new RunConfiguration(); } }
         public static RunConfiguration ShallowClone(RunConfiguration configuration)
@@ -58,6 +59,7 @@ namespace Pester
                 configuration.AssignObjectIfNotNull<string>(nameof(SkipRemainingOnFailure), v => SkipRemainingOnFailure = v);
                 configuration.AssignValueIfNotNull<bool>(nameof(FailOnNullOrEmptyForEach), v => FailOnNullOrEmptyForEach = v);
                 configuration.AssignObjectIfNotNull<string>(nameof(RepoRoot), v => RepoRoot = v);
+                configuration.AssignArrayIfNotNull<ScriptBlock>(nameof(BeforeInvoke), v => BeforeInvoke = v);
             }
         }
 
@@ -75,6 +77,7 @@ namespace Pester
             SkipRemainingOnFailure = new StringOption("Skips remaining tests after failure for selected scope, options are None, Run, Container and Block.", "None");
             FailOnNullOrEmptyForEach = new BoolOption("Fails discovery when -ForEach is provided $null or @() in a block or test. Can be overridden for a specific Describe/Context/It using -AllowNullOrEmptyForEach.", true);
             RepoRoot = new StringOption("Root directory of the repository. Found by searching for the .git directory recursively. When not found, the current working directory is used.", FindRepoRoot());
+            BeforeInvoke = new ScriptBlockArrayOption("ScriptBlocks to run in the caller's scope as soon as Invoke-Pester starts, before the caller's $PesterPreference is read. Use it to import dependencies and provide configuration. In addition to these scriptblocks, the first Pester.BeforeInvoke.ps1 file found when walking up from each Run.Path to Run.RepoRoot is dot-sourced.", new ScriptBlock[0]);
         }
 
         public StringArrayOption Path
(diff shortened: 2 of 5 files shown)


johlju:
I really like this, looking forward to this merging. I will reduce a lot of duplicated code in test files by having a `Pester.BeforeInvoke.ps1` that can contain bootstrap code. We can put one with different content in folders `qa`, `unit` and `integration`.

If it would go up to the repo root and find all and run them in sequence we could even reduce even more duplication, for example

```
reporoot
  |
  | -- tests    <--- finds Pester.BeforeInvoke.ps1
           |
           | -- unit
           | -- integration     <--- finds Pester.BeforeInvoke.ps1
```

It would then invoke them in the order top to bottom:
1. ./tests/Pester.BeforeInvoke.ps1
2. ./tests/integration/Pester.BeforeInvoke.ps1

But just having the option of one Pester.BeforeInvoke.ps1 would probably reduce the duplication with ~1000 rows in a larger project (like in SqlServerDsc).

nohwnd:
@johlju -rc1 is out, it has similar feature, BeforeContainer, my main problem is that we cannot just promise that it will run once, because if you need to run tests in parallel (there is experimental parallel mode), this script needs to run in every new runspace. And I wanted to minimize the overhead of looking the file up every time, and did not want to think about how to say "this is in child folder, but don't run any of the script files you find above me" like "root=true" in .editorconfig.  

but all those are good suggestions, show me how they simplify your workflow, and we could at least add them as options, if not as defult. I just did not want to over design and over promise in the initial release.

johlju:
It says in the release notes för rc1:

> "...single Pester.BeforeContainer.ps1 in the repository root..."
> "... run before every test file is discovered and run..."

Assuming we can pass `BeforeDiscovery`, `BeforeAll` and `AfterAll` in this file it might get us a a fair bit.  Some things do not run during Discovery and some things do not run during Run. Some code are duplicate, but we do slightly different setup for different things:

- unit tests of private and public command
- Integration tests of public commands
- Unit tests for class-based DSC resource
- Integration tests for class-based DSC resource
- Unit tests for MOF-based DSC resource
- Integration tests for MOF-based DSC resource

 A single file would not handle different setup unless it can pass current test script file so the code `Pester.BeforeContainer.ps1` can determine what is about to be tested.

But I will check how far the current functionality takes us and report the GAP. If we can pass `BeforeDiscovery`, `BeforeAll` and `AfterAll` in `Pester.BeforeContainer.ps1` that will potentially reduce most duplicated code. 
If it does not support `BeforeDiscovery`, `BeforeAll` and `AfterAll` as the release notes might suggest, then we'll see what we can put there, at least some of the `BeforeDiscovery` part could be moved... 🤔

nohwnd:
> Assuming we can pass BeforeDiscovery, BeforeAll and AfterAll

I actually don't know. I did not think about that, my thinking here was that it imports modules you need, adds helper functions etc. stuff that I typically do in a "test.ps1" script. But that now needs to happen in every child runspace. 

All in all the idea of what I want to achieve here is not super clear even to me. My goal is also that I can run tests (e.g. a single `It`) from VSCode that need some setup, and not repeat the setup on top of every file.

> A single file would not handle different setup unless it can pass current test script file so the code Pester.BeforeContainer.ps1 can determine what is about to be tested.

You could probably do that through the power of powershell, e.g. by looking at your pscall stack, but I get what you mean.

johlju:
> My goal is also that I can run tests (e.g. a single It) from VSCode that need some setup, and not repeat the setup on top of every file.

That is why there are extensive bootstrap in each file in SqlServerDsc - so there our goal is aligned 🙂 also I need to use [Invoke-PesterJob](https://github.com/viscalyx/Viscalyx.Common/wiki/Invoke%E2%80%91PesterJob) in VS Code (and outside) for tests if they load classes so I do not need to kill the session each time we re-run test.

johlju:
So after a bit of testing

1. `Pester.BeforeContainer.ps1` seems to support `BeforeDiscovery`, but not a top-level `BeforeAll` and `AfterAll`.
2. I currently have in a `BeforeDiscovery`-block in each test file, I moved that successfully into a single block in `Pester.BeforeContainer.ps1`. The moved code works as `BeforeDiscovery`-block and also without such block. I don't need it to run for every container **so that block would be better to put in a `Run.BeforeInvoke.ps1` if that exist in the future.**
3. What I miss is a `Pester.BeforeContainer.ps1` that can be placed in the same folder as the tests or a parent folder and support top-level `BeforeAll` and `AfterAll`, then I would be able to move duplicated code like this (this is simplified code, but as an example):
   ```powershell
   BeforeAll {
        $script:moduleName = 'MyModule'
    
        Import-Module -Name $script:moduleName
    
        $PSDefaultParameterValues['InModuleScope:ModuleName'] = $script:moduleName
        $PSDefaultParameterValues['Mock:ModuleName'] = $script:moduleName
        $PSDefaultParameterValues['Should-Invoke:ModuleName'] = $script:moduleName
        $PSDefaultParameterValues['Should-NotInvoke:ModuleName'] = $script:moduleName
   }
   
   AfterAll {
        $PSDefaultParameterValues.Remove('InModuleScope:ModuleName')
        $PSDefaultParameterValues.Remove('Mock:ModuleName')
        $PSDefaultParameterValues.Remove('Should-Invoke:ModuleName')
        $PSDefaultParameterValues.Remove('Should-NotInvoke:ModuleName')
    
        # Unload the module being tested so that it doesn't impact any other tests.
        Get-Module -Name $script:moduleName -All | Remove-Module -Force
   }
   ```

fflaten:
**BeforeInvoke**
I'm not sold on the `BeforeInvoke` story atm. I think `BeforeContainer` can handle most use-cases while supporting parallel and without overloading users with more options. Module imports, required due to parallel. Setup global requirements like dev server etc - run it conditionally by checking if running/exists/whatever.

**Folder-scoped `BeforeContainer`**
I like this, in addition to providing the context (metadata about the current container) in the future for more fine-grained conditional logic.

> I wanted to minimize the overhead of looking the file up every time, and did not want to think about how to say "this is in child folder, but don't run any of the script files you find above me" like "root=true" in .editorconfig.

Now that `BeforeContainer` is always file-bound we could stop at first match. We discover all tests. Then `Pester.BeforeContainer.ps1` in unique relevant test folders (+ up to reporoot). Build an array/lookup desc. sorted by depth and pick first match (if any) per file-type container. Users can dot-source any shared code using `$PSScriptRoot` relative paths if needed.

**BeforeAll/AfterAll**
I can see the value of centralized setup/teardown in the folder-scoped `BeforeContainer` files (or at that point `Pester.PerContainer.ps1`?). Would require rewriting how and when it's invoked + make sure you can't include `Describe/Context` blocks in it.

Something to explore, but maybe not critical? While it wouldn't remove the boilerplate, you could deduplicate 99% by dot-sourcing, couldn't you? `BeforeAll { . "$PSScriptRoot\beforeall.ps1" }; AfterAll { . "$PSScriptRoot\afterall.ps1" };`

johlju:
> Something to explore, but maybe not critical? While it wouldn't remove the boilerplate, you could deduplicate 99% by dot-sourcing, couldn't you? BeforeAll { . "$PSScriptRoot\beforeall.ps1" }; AfterAll { . "$PSScriptRoot\afterall.ps1" };

Yes, but having a standard it would be easier for contributors to understand where to look and what its for, regardless of repo. Dot-sourcing would probably work, but naming and location would most likely be different for each repo a contributor looks at. 🤔 

> I think BeforeContainer can handle most use-cases

I think so too. BeforeInvoke would be good for things that only need to run once for each `Invoke-Pester` call, not needing to run per container. But I would think it would not be used at all if several BeforeContainter can me concatenated from the parent directory tree.

nohwnd:
Superseded by #2993 thanks for the input @johlju .

PR #2802, opened 2026-06-29
Fix parallel run total durations

### LINKED ISSUE #2794: [v6 Parallel] Total durations are wrong
### Checklist

### What is the issue?

When running test files in parallel, the "Tests completed in ..." output and Duration-values in the Pester.Run output object are way too high.

Container-durations are summed, which doesn't work with parallel execution.

https://github.com/pester/Pester/blob/b03b644962da843b6c0439002a694b84bde281bc/src/Pester.RSpec.ps1#L279-L282

### Expected Behavior

Correct duration in output

### Steps To Reproduce

_No response_

### Describe your environment

_No response_

### Possible Solution?

`$run.Duration` is the actual duration of the run. Endtime - Starttime from parent / orchestrator

Leave Discovery-/Framework-/UserDuration values blank when in parallel mode. Precise totals requires excluding overlapping time between container using timestamps. Not worth it.

nohwnd:
Container durations were summed, which overstates a parallel run since the files overlap in wall-clock time. Measure the actual elapsed time for `Run.Duration` in parallel instead.

The per-phase totals (user, framework, discovery) are still summed rather than blanked - they are cumulative work measurements, so they stay meaningful across the overlapping workers and add up to the total container work.

Added a parallel duration test.

Fix #2794

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -663,7 +663,8 @@ function Invoke-Pester {
             # If every file opted out with #pester:no-parallel, the run is effectively sequential,
             # so fall through to the sequential path - which fires the framework's own global
             # plugin steps at the correct interleaved points.
-            if ($useParallel -and $parallelSupported -and $allFileContainers -and -not $coverageEnabled -and -not $skipRemainingRunScope -and 0 -lt $parallelContainers.Count) {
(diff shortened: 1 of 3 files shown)


nohwnd:
Superseded by #2807. That one measures the run wall-clock for `Run.Duration` instead of summing the container durations, and blanks the per-phase run totals (user/framework/discovery) rather than keeping the sums - the summed per-phase values here were the misleading part, since you can't add durations of work that ran in parallel and get an elapsed time. Closing in favour of #2807.

PR #2819, opened 2026-06-30
Import Pester via its manifest in parallel workers

### LINKED ISSUE #2816: Parallell bug
> @nohwnd I have an edge case that fails when running in parallel. One of my modules that I use for infrastructure testing `KBInfrastructureTesting` requires specific version of Pester in the manifest:
> ```pwsh
> RequiredModules = @(
>     @{ModuleName = 'KBActiveDirectory'; ModuleVersion = '7.0.0'},
>     @{ModuleName = 'KBSystemFunctions'; ModuleVersion = '7.0.0'},
>     @{ModuleName = 'KBSystemManagement'; ModuleVersion = '7.0.0'},
>     @{ModuleName = 'KBServiceManagement'; ModuleVersion = '7.0.0'},
>     @{ModuleName = 'KBDomainComputerManagement'; ModuleVersion = '7.0.0'},
>     @{ModuleName = 'Pester'; ModuleVersion = '5.7.1'},
>     @{ModuleName = 'PowerShellHumanizer'; ModuleVersion = '13.2.1'},
>     'SQLServer',
>     'Microsoft.PowerShell.ThreadJob'
> )
> ```
> In one of my tests I'm testing module import, and this fails now:
> ```pwsh
> Context 'Module Import' {
>     It '[<_.Name>] Module imports successfully' {
>         {Import-Module -ModuleInfo $_ -Force -Verbose:$false -ErrorAction Stop} |
>             Should -Not -Throw
>     }
> }
> ```
> The error:
> ```pwsh
> [Bukem@VOYAGER]≥# gmo KBInfrastructureTesting -ListAvailable | Test-KBModule -Tag Manifest
> Running tests from 1 files in parallel.
> [-] Manifest Specification In Module: "KBInfrastructureTesting".Module Integrity.Manifest Property.Module Import.[KBInfrastructureTesting] Module imports successfully 2.87s
>  Expected no exception to be thrown, but an exception "The required module 'Pester' was not loaded because no valid module file was found in any module directory." was thrown from D:\Users\Bukem\Documents\PowerShell\.projects\Modules\KBDevelopmentTools\Integration\01-Manifest.Tests.ps1:236 char:22
>      + …            {Import-Module -ModuleInfo $_ -Force -Verbose:$false -Erro …
>      +               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
> Tests completed in 4.42s
> Tests Passed: 21, Failed: 1, Skipped: 0, Inconclusive: 0, NotRun: 0
> Exception: Pester run failed, because 1 test failed
> ```
> The fix is easy. I could move the `Pester` module specification from `RequiredModules` to `ExternalModuleDependecies` and then check for the required version in the root module. But I just wanted to let you know anyways.
>  

 _Originally posted by @kborowinski in [#2795](https://github.com/pester/Pester/issues/2795#issuecomment-4840679498)_

nohwnd:
## Problem

Under `Run.Parallel`, a test that imports a module whose manifest lists Pester in `RequiredModules` fails, e.g.:

```powershell
RequiredModules = @( @{ ModuleName = 'Pester'; ModuleVersion = '5.7.1' } )
```

```
Expected no exception to be thrown, but an exception "The required module 'Pester' was not
loaded because no valid module file was found in any module directory." was thrown ...
```

The same test passes in a sequential run.

## Root cause

`Invoke-TestInParallel` handed each worker `$ExecutionContext.SessionState.Module.Path`, which is the **root module** (`Pester.psm1`), not the **manifest** (`Pester.psd1`). Each worker does `Import-Module <that path>`, and importing the bare `.psm1` loads Pester **without its manifest metadata**, so the worker's Pester reports `ModuleVersion` `0.0.0.0`.

When a test then imports a third‑party module that requires `Pester >= 5.7.1`, the loaded `0.0.0.0` Pester does not satisfy the requirement, so PowerShell tries to load Pester from disk — throwing either *"no valid module file was found in any module directory"* (reporter's environment) or *"Could not load file or assembly 'Pester … already loaded"* (when a real Pester is installed). Sequential runs are unaffected because the parent imported Pester through its manifest, so the version is correct.

## Fix

Resolve the manifest (`<ModuleBase>/<Name>.psd1`) and import that in the worker, falling back to the root‑module path when no manifest is present. Workers now load Pester with its real version, so `RequiredModules = Pester` resolves against the already‑loaded module.

## Test

Adds a regression test in `tst/Pester.RSpec.Parallel.ts.ps1` that runs a module declaring `RequiredModules = @{ ModuleName = 'Pester'; … }` under `Run.Parallel`. It fails before the fix (worker Pester is `0.0.0.0`) and passes after. Verified the full parallel suite stays green and the custom build analyzer is clean.

Fix #2816

### DIFF
--- a/src/functions/Pester.Parallel.ps1
+++ b/src/functions/Pester.Parallel.ps1
@@ -175,7 +175,17 @@ function Invoke-TestInParallel {
         })
 
     # Path to the currently loaded Pester so each worker imports the exact same build.
-    $modulePath = $ExecutionContext.SessionState.Module.Path
+    # Use the module manifest (.psd1), not the root module (.psm1) that Module.Path points to:
+    # importing the bare .psm1 loads Pester without its manifest metadata, so its ModuleVersion
+    # becomes 0.0.0.0. That then fails to satisfy any module a test imports whose manifest lists
+    # Pester in RequiredModules (e.g. @{ ModuleName = 'Pester'; ModuleVersion = '5.7.1' }),
+    # because the loaded 0.0.0.0 is below the required version (#2816).
+    $pesterModuleInfo = $ExecutionContext.SessionState.Module
+    $modulePath = $pesterModuleInfo.Path
+    $manifestPath = & $SafeCommands['Join-Path'] $pesterModuleInfo.ModuleBase "$($pesterModuleInfo.Name).psd1"
+    if (& $SafeCommands['Test-Path'] -LiteralPath $manifestPath -PathType Leaf) {
+        $modulePath = $manifestPath
+    }
 
     # Cap concurrency at Run.ParallelThrottleLimit when set (> 0); otherwise use all processors.
     $requestedThrottle = [int]$Configuration.Run.ParallelThrottleLimit.Value
--- a/tst/Pester.RSpec.Parallel.ts.ps1
+++ b/tst/Pester.RSpec.Parallel.ts.ps1
@@ -193,6 +193,52 @@ Describe 'D' { It 'sees data' { $Module | Should -Be 'hello'; $Data.k | Should -
         }
     }
 
+    b "Run.Parallel module loading" {
+        t "imports a module that lists Pester in RequiredModules (#2816)" {
+            # Each parallel worker imports Pester so test bodies can use it. The worker must import
+            # Pester *via its manifest* so the loaded module keeps its real ModuleVersion. Importing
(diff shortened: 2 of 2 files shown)


fflaten:
Of course! That's why the formatview tests failed when I tested parallel in this repo. 💡

PR #2821, opened 2026-06-30
Anchor module-less Mock to the test scope instead of the caller's module

### LINKED ISSUE #2025: Pester creating mock in module scope when -ModuleName is not provided

Pester is placing mock inside of module scope even when ```-ModuleName``` isn't specified. Diagnostic output incorrectly assets that ModuleName _was_ specified. 

## Describe your environment

Pester version     : 5.2.2 /User/powershell/Modules/Pester/5.2.2/Pester.psm1                                         
PowerShell version : 7.1.3                                                                                                                  
OS version         : Unix 11.0.0 

## Step to reproduce

ExampleModule.psm1 - the function I am attempting to mock

```powershell
function Invoke-MyExampleFunction {
    [Cmdletbinding()]
    param()
    Write-Error "Mock failed. Original function was called."
}
```

MockInjection.psm1 - I call the function in this module to insert the mock into my test file. This allows me to easily reuse my common mocks

```powershell
function Invoke-MockInjection {
    Mock Invoke-MyExampleFunction {
        Write-Host 'Mocked from MockInjection module'
    }
}
```

MyScript.ps1

```powershell
Import-Module ExampleModule

Invoke-MyExampleFunction
```

MyScript.Tests.ps1

```powershell

$PesterPreference = New-PesterConfiguration @{
    Output = @{
        Verbosity = 'Diagnostic'
    }
}

Describe "Example Tests" {

    BeforeAll {

        Import-Module ExampleModule
        Import-Module MockInjection

        # The mock works as expected if defined inside of the BeforeAll block in the test script
        # Mock Invoke-MyExampleFunction {
        #     Write-Host "Mocked from inside BeforeAll block. Works as expected."
        # }

        # This does not work. See below diagnostic output. 
        Invoke-MockInjection

        $scriptPath = $PSCommandPath -replace '.Tests.ps1', '.ps1'
    }
    
    It "Mocks correctly" {
        { . $scriptPath } | Should -Not -Throw
    }
}

```

The mock works as expected when it is created inside of the test script itself. If it is injected from another module for the purposes of reusing some mock behavior, it fails. Diagnostic output makes it appear as though moduleName is being specified when it is not. 

diagnostic output. Note line 4 and 5 that ModuleName was specified when it was not. 
```
Mock: Setting up default mock for MockInjection - Invoke-MyExampleFunction. 
Mock: We are in a block, one time setup or similar. Returning mock table from test block. 
Mock: Resolving command Invoke-MyExampleFunction. 
Mock: ModuleName was specified searching for the command in module MockInjection. 
Mock: We are already running in MockInjection. Using that. 
Mock: Found the command Invoke-MyExampleFunction in a different module. 
Mock: Mock does not have a hook yet, creating a new one. 
Mock: Defined new hook with bootstrap function PesterMock_MockInjection_Invoke-MyExampleFunction_8683fddf-5380-43a6-a489-ab324f65a01b and aliases Invoke-MyExampleFunction, ExampleModule\Invoke-MyExampleFunction. 
Mock: Adding a new default behavior to MockInjection - Invoke-MyExampleFunction. 
```

nohwnd:
Fix #2025

## Problem

When `Mock` (or `Should -Invoke`) is called **without** `-ModuleName` from a helper function that lives in a module — for example a reusable mock-injection helper — the mock was silently scoped to that helper module instead of the test scope. As a result the mock never applied to the command under test, and the real command ran instead.

The root cause is that the target was inferred purely from the caller's session state module: any call originating inside a module inherited that module, with no way to tell an *intentional* in-module mock (`InModuleScope M { Mock ... }`) from one that merely passes through a module-defined helper.

## Fix

Resolve the mock target explicitly via a shared `Resolve-MockCallerScope` helper used by **both** `Mock` and `Should -Invoke`, so they always agree on where the mock lives:

- **Intentional in-module call** → inherit the module (unchanged behavior). This is detected when either:
  - we are directly inside `InModuleScope <module> { ... }` (InModuleScope now pushes/pops a marker for the module it runs in), or
  - the current block/test body was itself defined in that module (e.g. `InModuleScope` wrapping a `Describe`).
- **Helper-in-a-module call** (the #2025 case) → anchor the mock to the **test/script** session state, exactly as a direct `Mock` in the test would.
- **No running test/block** → historical fallback.

## Tests

- New regression `Describe` in `tst/functions/Mock.Tests.ps1` covering the helper-injected mock applying to the test, `Should -Invoke` consistency from both the helper module and the test scope, and a boundary check that `InModuleScope` without `-ModuleName` still targets the module. These were verified to **fail on the unpatched build** and pass with the fix.
- No existing tests needed rewriting (the in-module/`InPesterModuleScope` cases take the block-match path and are preserved).

## Validation

Full suite green — P-tests pass plus 2496 Pester tests, 0 failed. Mock.Tests.ps1 (254) and InModuleScope.Tests.ps1 (39) all pass. Build analyzer reports no new violations in the edited source files.


nohwnd:
needs careful review

PR #2850, opened 2026-07-09
Add experimental global mocks

### LINKED ISSUE #2705: Idea: can we inject mock into the original module, even if binary?
can we make global mocks by injecting into the original module, even when binary, to make sure we never ever call a command like invoke-webrequest?

nohwnd:
A normal mock only applies to calls from the scope where it is defined, or from the module you pass with `-ModuleName`. Code in another module looks the command up in its own session state and never sees the mock. So to be sure a command like `Invoke-WebRequest` is never called from any code under test, you have to know every module that might call it and mock it in each one.

Global mocks apply to calls of the mocked command from any module or script in the runspace, so you don't have to. This is experimental.

## `Mock.Global` configuration option

Turn global mocks on for the whole run with the experimental option:

```powershell
$config.Mock.Global = $true
```

With it on you write your mocks exactly as you do today and they simply reach the command wherever it is called. One mock covers every module, so you no longer have to name each calling module:

```powershell
Mock Invoke-WebRequest { '<html />' }
Get-Data   # a function in another module that calls Invoke-WebRequest
Should -Invoke Invoke-WebRequest -Times 1
```

To make sure a command never runs, mock it to throw. Combine that with `-ParameterFilter` to block only the calls you care about and let the rest fall through to the real command:

```powershell
# block deleting anything outside TestDrive, from any code under test
Mock Remove-Item { throw 'blocked' } -ParameterFilter { $Path -notlike "$TestDrive*" }
```

The mock is removed when the test or block that defined it ends, like any other mock.

With the option on, `-ModuleName` no longer scopes a mock. It is used only as a hint to resolve the command, so a module-private command can still be found and mocked. You can point it to any place where the command is resolvable - that is to the module where it is defined, or to the module where the function is exported to. If your current mocks use `-ModuleName`, you should not need to change them to use  the new .Global option.

## How it works

A global mock is a normal Pester mock plus a registration in the runspace-wide `InvokeCommand.PreCommandLookupAction` hook. That hook fires on every parser-driven command lookup, in any module, so the lookup is redirected to the mock's bootstrap function no matter where the call comes from. The hook is installed while at least one global mock exists and removed once none remain.

Pester's own internals are not affected. Pester calls its internal commands through `$SafeCommands`, a captured `CommandInfo` invoked with the call operator, which does not go through command lookup. So even mocking a command Pester uses internally (for example `Get-Command`) does not break Pester itself.

### Isolation between runs

Because the hook and its bootstrap alias are runspace-wide, a global mock is tied to the run that created it so it can't leak elsewhere:

- **Each run has its own identity.** Every `Invoke-Pester` run gets a unique id that is baked into each mock it creates. A mock only applies while its own run is executing.
- **Nested Pester-in-Pester runs are fully isolated.** If an outer run's global mock leaks into a nested run through the outer script's scope, the nested run sees the id mismatch and calls the real command instead of applying the leaked mock. The nested run can mock the same command independently, and the outer run's global mocks are snapshotted and restored around the nested run so they are never clobbered.
- **Interrupted runs recover.** The hook is runspace state that outlives a single test, so a `Ctrl+C` during a global mock could leave it armed. A fresh top-level run resets the hook before it starts, so it always begins with a clean slate.

The configuration option is experimental and may change based on feedback.

## What `-ModuleName` does

Without the option, a mock is module-scoped: `Mock -ModuleName A Cmd` affects only calls made from module `A`. This is the default, and it gives you three things a global mock does not:

- **Isolation to a single module.** Only calls from `A` are affected; the same-named command in module `B`, or in the test script, is untouched. A global mock reaches all of them.
- **Disambiguating same-named commands.** Two modules — or two nested copies — that each define `Get-Data` can be mocked independently with `-ModuleName`. One global `Get-Data` mock catches every copy.
- **Module-scoped call history and messages.** `Should -Invoke -ModuleName A` counts only `A`'s calls, and the "verifiable not called" message names the module. A global mock records once with no module, so both the per-module count and that message qualifier go away.

Fix #2705

### DIFF
--- a/src/Main.ps1
+++ b/src/Main.ps1
@@ -460,14 +460,34 @@ function Invoke-Pester {
         # this will inherit to child scopes and allow Describe / Context to run directly from a file or command line
         $invokedViaInvokePester = $true
 
+        # global mock hook state carried from begin to the finally in the end block (nested runs only)
+        $runningPesterInPester = $false
+        $savedGlobalMockState = $null
+
+        # Give this run a unique identity used to isolate global mocks between (possibly nested) runs. A
+        # mock's bootstrap records the run that created it; a leaked bootstrap whose id does not match the
+        # currently executing run defers to the original command instead of applying (see Invoke-Mock).
+        # The previous id is restored when this run ends so nested runs each get their own identity.
+        $pesterRunId = [Guid]::NewGuid().Guid
+        $previousPesterRunId = [Pester.GlobalMockHook]::SetCurrentRun($pesterRunId)
+
         if ($null -eq $state) {
             # Cleanup any leftover mocks from previous runs, but only if we are not running in a nested Pester-run
             # todo: move mock cleanup to BeforeAllBlockContainer when there is any?
             Remove-MockFunctionsAndAliases -SessionState $PSCmdlet.SessionState
+            # The global mock hook is runspace-wide state that a normal mock cleanup does not touch, so an
+            # interrupted previous run (e.g. Ctrl+C during a global mock) can leave it armed. Reset it here
+            # so a fresh top-level run always starts with no global mocks and no lookup handler installed.
+            Reset-GlobalMockHook
         }
         else {
             # this will inherit to child scopes and affect behavior of ex. TestDrive/TestRegistry
             $runningPesterInPester = $true
+            # This is a nested run. The global mock hook is shared across the whole runspace, so give this
+            # run its own clean slate and protect the outer run: snapshot the outer run's global mocks, then
+            # clear the shared state. It is restored once this run ends (see the finally in the end block).
+            $savedGlobalMockState = Get-GlobalMockHookState
+            Reset-GlobalMockHook
         }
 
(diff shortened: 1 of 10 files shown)


nohwnd:
wrote to wrong pr.

fflaten:
Wrong PR?

nohwnd:
yup

nohwnd:
@fflaten tested on pester itself found few issues, like us totally ignoring the moduleName, and because of that not being able to resolve the function when it is module private.

nohwnd:
so don't waste your time reviewing yet. I will anyway publish it for people to try under experimental a flag. Makes it easiest to find edge cases.

nohwnd:
So with making `-ModuleName` a hint where the function can be found (e.g. where it is exported to or where it comes from), there are no failures in pester own test suite - when we run it with the new mock.global = true option. Apart from 5 that test that the mock isolation works  - the exact opposite of what we are trying to achieve here.

PR #2858, opened 2026-07-10
[DIAGNOSTIC - do not merge] Should-BeFasterThan flake hammer

nohwnd:
Throwaway diagnostic to settle the root cause of the `Should-BeFasterThan` timing flake.

Adds `diag-sleep.ps1` and a **SLEEP DIAGNOSTIC** CI step that runs on every matrix agent (PS7 + PS5.1, Win/Linux/mac). On each agent it:

1. **MECH** — sanity: confirms a `Select -First`-absorbed `Start-Sleep` interrupt is silent and <1ms.
2. **HAMMER** — replicates the exact production measurement (`$sw=StartNew; & $Actual; $sw.Stop`) for ~120s, bracketed with an **independent non-QPC clock** (winmm `timeGetTime`) and a **post-sleep marker**. On any sub-1ms reading or exception it emits a `SLEEPDIAG:: ANOM` line so we can tell an **interrupt** (marker missing / exception) from a **QPC short-read** (Stopwatch <1ms but winmm ~10ms).
3. **REALASSERT** — hammers the real exported `Should-BeFasterThan -Expected 1ms` and counts self-misfires (expect 0).

Harvest `SLEEPDIAG::` lines from the build logs. Not for merge; will be deleted.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -117,6 +117,14 @@ stages:
               script: |
                 & ./test.ps1 -CI -CC -PassThru -NoBuild
               workingDirectory: '$(Build.SourcesDirectory)'
+          - task: PowerShell@2
+            displayName: 'SLEEP DIAGNOSTIC'
+            condition: succeededOrFailed()
+            inputs:
+              targetType: filePath
+              filePath: '$(Build.SourcesDirectory)/diag-sleep.ps1'
+              pwsh: $(pwsh)
+              workingDirectory: '$(Build.SourcesDirectory)'
           - task: PublishCodeCoverageResults@2
             inputs:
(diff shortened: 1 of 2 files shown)


nohwnd:
Diagnostic complete. Across ~62k hammer samples on all 8 CI agents (incl. WinPS 5.1 + PS7 on Windows Server 2022/2025 with an independent winmm witness): under1ms=0, exceptions=0, qpcMin always >=9.15ms, and the real assertion self-misfired 0/13848 times. No interrupt, no QPC short-read. The Should-BeFasterThan measurement is sound; the historical flake does not reproduce. Closing and deleting branch.

PR #2863, opened 2026-07-10
Allow empty string for `Should-BeString -Expected`

### LINKED ISSUE #2857: `Should-BeString` should allow empty strings for the `-Expected` parameter
### Checklist

- [x] Feature request has a meaningful title

I've this kind of v5 test:

```powershell
It "should do something" -ForEach @(
	@{ Value = ""; Length = 0; Expected = "" }
	@{ Value = "foo bar"; Length = 3; Expected = "foo" }
) {
	Format-String $value -Truncate -Length $length | Should -BeExactly $expected
}
```

When porting it to v6, I now must use:

```powershell
It "should do something" -ForEach @(
	@{ Value = ""; Length = 0; Expected = "" }
	@{ Value = "foo bar"; Length = 3; Expected = "foo" }
) {
	$actual = Format-String $value -Truncate -Length $length
	if ($expected) { Should-BeString $expected $actual -CaseSensitive }
	else { Should-BeEmptyString $actual }
}
```

I would have preferred not to have to use two different assertions depending on whether the expected value is an empty string or not.

nohwnd:
Fix #2857

## Problem

`Should-BeString` rejects an empty string for `-Expected`:

```powershell
"" | Should-BeString ""
# Cannot bind argument to parameter 'Expected' because it is an empty string.
```

This forces users porting v5 `-ForEach` tests to branch between `Should-BeString` and `Should-BeEmptyString` depending on whether the expected value is empty.

## Root cause

The rejection is not a deliberate assertion rule — there's no validation in the function body checking for empty. It's an implicit side effect of declaring the parameter as `[Parameter(Mandatory)][String]$Expected`: a mandatory, typed `[String]` parameter automatically refuses `""` during parameter binding.

Everything downstream already handles empty strings:
- `Test-StringEqual` compares empty strings correctly.
- `Get-StringDifferenceMessage` already declares `[AllowEmptyString()]` on `$Expected`/`$Actual`.
- The sibling `Should-NotBeString` doesn't mark `Expected` as `Mandatory`, so it already accepts empty strings — `Should-BeString` was the inconsistent one.

## Change

- Add `[AllowEmptyString()]` to `$Expected` (keeping `Mandatory`, so the argument is still required — it may just be empty).
- Add a comment-based-help example.
- Add tests covering: both empty (named, positional, pipeline) and the failure/diff message when expected is `""` but actual is non-empty.

All 42 tests in `Should-BeString.Tests.ps1` pass.

### DIFF
--- a/src/functions/assert/String/Should-BeString.ps1
+++ b/src/functions/assert/String/Should-BeString.ps1
@@ -69,6 +69,13 @@ function Should-BeString {
 
     This assertion will fail, because the actual value is not equal to the expected value.
 
+    .EXAMPLE
+    ```powershell
+    "" | Should-BeString ""
+    ```
+
+    This assertion will pass, because an empty string is allowed as the expected value.
+
     .NOTES
     The `Should-BeString` assertion is the opposite of the `Should-NotBeString` assertion.
 
@@ -88,6 +95,7 @@ function Should-BeString {
         [Parameter(Position = 1, ValueFromPipeline = $true)]
         $Actual,
         [Parameter(Position = 0, Mandatory)]
+        [AllowEmptyString()]
         [String]$Expected,
         [String]$Because,
         [switch]$CaseSensitive,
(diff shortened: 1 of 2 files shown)


johlju (approved):
Looks good to me

PR #2866, opened 2026-07-13
Optimize formatting for template strings and numbers

### LINKED ISSUE #2865: Name expansion hangs on complex objects like CommandInfo
### Checklist

### What is the issue?

Name expansion in Pester v6 "hangs" when referencing complex objects with many properties like `CommandInfo`.

Cause: Getting stuck in recursion with `Format-Nicely2`-> `Format-Object2` -> `Format-Nicely2` .. looping properties.

### Expected Behavior

Not hang.

### Steps To Reproduce

```powershell
Describe 'Infinite format loop?' {
    BeforeAll {
        $func = Get-Command -Name 'Invoke-Pester'
    }

    it '<func.name> is cool' {
        $func.Name | Should -Be 'Invoke-Pester'
    }

    it '<func> will hang' {
        $func.Name | Should -Be 'Invoke-Pester'
    }
}
```

Or just `& (Get-Module Pester) { Format-Nicely2 (Get-Command Invoke-Pester) }`

### Describe your environment

```
# Began in Pester 6.0.0

Pester version     : 6.1.0-alpha2 /workspaces/Pester/bin/Pester.psm1                                                    
PowerShell version : 7.4.16
OS version         : Unix 6.18.35.2
```

### Possible Solution?

Guidance: Test authors should reference the string-property, like `<mycommandinfo.name>`

Technical: We should have some guardrails to avoid errors. Restrict `Format-NicelyForTemplate` to 0 or 1 level of recursion. Anything else gets messy.

We could also special case `CommandInfo` to `[string]` cast like v5, but do we want to? Without deep format, v6 will return `[Management.Automation.FunctionInfo]` in this scenario. That would help authors identify a bad practice and fix it.

fflaten:
## PR Summary

Restricts template strings to one level of object recursion in formatting to avoid hangs on large complex objects.

To address the issue above, this PR also:
- Refactors Format2 to use a decreasing and internally configurable `-MaxDepth` instead of increasing `-Depth` with hard limit. Easier to restrict something a max 1 level than start at depth 9.
- Uses `Format-Numbers*` for integers and not just decimals. Integers was formatted as ValueType, requring `-MaxDepth 2` for formatting simple arrays.
  - Same applies to `[char]` vs `[string]` etc, but not fixed for now.

Fix #2865

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Format2.ps1
+++ b/src/Format2.ps1
@@ -1,7 +1,11 @@
-﻿function Format-Collection2 ($Value, [switch]$Pretty, [int]$Depth = 0) {
+﻿# Default recursion limit for formatting. A depth of 10 (default) is never useful in an assertion message
+# and is well below PowerShell's own call-depth limit, so it is fixed here rather than exposed as a configuration.
+$maximumFormatDepth = 10
+
+function Format-Collection2 ($Value, [switch]$Pretty, [int]$MaxDepth = $maximumFormatDepth) {
     $length = 0
     $o = foreach ($v in $Value) {
-        $formatted = Format-Nicely2 -Value $v -Pretty:$Pretty -Depth ($Depth + 1)
+        $formatted = Format-Nicely2 -Value $v -Pretty:$Pretty -MaxDepth ($MaxDepth - 1)
         $length += $formatted.Length + 1 # 1 is for the separator
         $formatted
     }
@@ -19,7 +23,7 @@
     }
 }
 
-function Format-Object2 ($Value, $Property, [switch]$Pretty, [int]$Depth = 0) {
+function Format-Object2 ($Value, $Property, [switch]$Pretty, [int]$MaxDepth = $maximumFormatDepth) {
     if ($null -eq $Property) {
         $Property = foreach ($p in $Value.PSObject.Properties) { $p.Name }
     }
@@ -31,7 +35,7 @@ function Format-Object2 ($Value, $Property, [switch]$Pretty, [int]$Depth = 0) {
     $valueType = Get-ShortType $Value
     $items = foreach ($p in $orderedProperty) {
         $v = ([PSObject]$Value.$p)
-        $f = Format-Nicely2 -Value $v -Pretty:$Pretty -Depth ($Depth + 1)
+        $f = Format-Nicely2 -Value $v -Pretty:$Pretty -MaxDepth ($MaxDepth - 1)
         "$p=$f"
     }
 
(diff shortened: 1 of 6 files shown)


src/Format2.ps1
```diff
@@ -175,7 +177,8 @@ function Format-NicelyForTemplate ($Value) {
         return $Value
     }
 
-    Format-Nicely2 -Value $Value
+    # Limit recursion to simple arrays and hashtable to avoid hangs and complex test names/paths.
+    Format-Nicely2 -Value $Value -MaxDepth 1
```
fflaten: This will still print a massive multi-line name for something like `(Get-Command Invoke-Pester)`, but at least it doesn't hang. Questions:

- Should we special case `CommandInfo` to just return the name like v5? Or let it be so the author will identify a bad practice (not referencing the name property) and fix it.
- Should tests names expand arrays/dictionaries etc in general?
- Should tests names allow multi-line at all? Messy console output. Trim or run through `[Pester.Formatter]::EscapeControlChars()`?

src/TypeClass.ps1
```diff
@@ -16,8 +16,13 @@ function Is-ScriptBlock ($Value) {
     $Value -is [ScriptBlock]
 }
 
+function Is-IntegralNumber ($Value) {
+    # Note: Using int16/uint16 as short/ushort names doesn't exist in PS 5.1
+    $Value -is [int] -or $Value -is [long] -or $Value -is [int16] -or $Value -is [sbyte] -or $Value -is [uint] -or $Value -is [ulong] -or $Value -is [uint16] -or $Value -is [byte]
```
fflaten: Use .NET names. Some C# names are missing in 5.1. https://github.com/PowerShell/PowerShell/commit/8f4b66af295e96c86eb843f432ade9fbda60e6e8

src/TypeClass.ps1
```diff
@@ -16,8 +16,14 @@ function Is-ScriptBlock ($Value) {
...
 
+function Is-IntegralNumber ($Value) {
+    # Note: Using .NET type names for consistency because PowerShell 5.1 doesn't support short/ushort/uint/ulong
+    $Value -is [Int32] -or $Value -is [Int64] -or $Value -is [Int16] -or $Value -is [SByte]
+    $Value -is [UInt32] -or $Value -is [UInt64] -or $Value -is [UInt16] -or $Value -is [Byte]
+}
+
 function Is-DecimalNumber ($Value) {
-    $Value -is [float] -or $Value -is [single] -or $Value -is [double] -or $Value -is [decimal]
+    $Value -is [Single] -or $Value -is [Double] -or $Value -is [Decimal]
```
fflaten, line 26: Just cosmetic to get consistent naming in code and tests `<Type.Name>` for both PS 5.1 and 7+

nohwnd:
Hey @fflaten :wave: I pushed a commit onto this branch to iterate on it — hope that's ok, happy to revert if you'd rather take it a different way.

Your `-MaxDepth` refactor + integer fix are exactly the right foundation, so I built on top rather than starting over. The one thing left over was that a complex object referenced whole in a name (e.g. `<cmd>` instead of `<cmd.Name>`, where `cmd` is a `CommandInfo`) still expanded one full level — for a `CommandInfo` that's the entire `Definition` and ends up ~80k chars, so the name is huge even though it no longer hangs.

The commit finishes the type-map idea that was already sketched in the `Get-DisplayProperty2` comment: match by **base type** (one `CommandInfo` entry covers `FunctionInfo`/`CmdletInfo`/`AliasInfo`/`ExternalScriptInfo`/…) and register `CommandInfo -> Name`. `Format-Nicely2` consults the registry before expanding an object, so a registered type renders a compact summary in both assertion messages and test-name templates:

```
<cmd>  ->  Management.Automation.FunctionInfo{Name='Invoke-Pester'}
```

Anything not in the registry keeps today's behaviour (full expansion in assertion messages), and in a name it just collapses to `[SomeType]` so an unknown object can never blow up a test name. Taming a new noisy type is then a one-line registry entry — that's the "least maintenance" part.

Added tests for the base-type match, the compact summary, and an RSpec acceptance test for the #2865 repro. Clean build is green; only the unrelated pre-existing `SkipRemainingOnFailureCount` failures remain. Let me know what you think!

nohwnd:
Superseded by #2926, which took the same approach (thanks @fflaten). Merged now, so closing this one.

PR #2869, opened 2026-07-15
Use v6 Should-* syntax in the Planets example

nohwnd:
The Planets example under `docs/Examples` still teaches the classic `Should -Be` syntax. v6 recommends the new `Should-*` assertions, so the example should use them.

The name filter test used one `Should -Be` for both the single-match filters ('Earth', 'ne*', 'ur*') and the multi-match 'm*' filter. v6 keeps single-value and collection assertions separate, so I split it: single matches use `Should-Be`, and 'm*' uses `Should-BeCollection`. The no-match case becomes `Should-BeNull`.

I also updated the "You should see this output" comment, in v6 a collection renders as `@('Mercury', 'Mars')` rather than `Mercury Mars`.

Verified green against Pester 6.

### DIFF
--- a/docs/Examples/Planets/Get-Planet.Tests.ps1
+++ b/docs/Examples/Planets/Get-Planet.Tests.ps1
@@ -54,8 +54,8 @@ Describe 'Get-Planet' {
         $allPlanets = Get-Planet
 
         # We count how many planets we got. And validate it by using
-        # the Should-Be assertion.
-        $allPlanets.Count | Should-Be 8
+        # the Should-BeCollection assertion.
+        $allPlanets |  Should-BeCollection -Count 8
 
         # The assertion will do nothing if the count is 8,
         # and throw an exception if the count is something else.
@@ -117,7 +117,6 @@ Describe 'Get-Planet' {
         It "Given valid -Name '<Filter>', it returns '<Expected>'" -TestCases @(
             @{ Filter = 'm*'   ; Expected = 'Mercury', 'Mars' }
         ) {
-            param ($Filter, $Expected)
             $planets = Get-Planet -Name $Filter
             $planets.Name | Should-BeCollection $Expected
         }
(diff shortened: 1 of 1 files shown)


docs/Examples/Planets/Get-Planet.Tests.ps1
```diff
@@ -54,8 +54,8 @@ Describe 'Get-Planet' {
         $allPlanets = Get-Planet
 
         # We count how many planets we got. And validate it by using
-        # the Should -Be assertion.
-        $allPlanets.Count | Should -Be 8
+        # the Should-Be assertion.
+        $allPlanets.Count | Should-Be 8
```
johlju: Maybe change to:

```suggestion
        $allPlanets |  Should-BeCollection -Count 8
```

docs/Examples/Planets/Get-Planet.Tests.ps1
```diff
@@ -106,18 +108,29 @@ Describe 'Get-Planet' {
...
+            $planets.Name | Should-Be $Expected
+        }
+
+        # When the filter matches more than one planet, Get-Planet returns a
+        # collection of names. In v6 we assert collections with Should-BeCollection,
+        # which compares the items one by one. Should-Be is only for single values.
+        It "Given valid -Name '<Filter>', it returns '<Expected>'" -TestCases @(
+            @{ Filter = 'm*'   ; Expected = 'Mercury', 'Mars' }
+        ) {
+            param ($Filter, $Expected)
```
johlju: Do we really need param now, isn't that automatically done?

```suggestion
```

docs/Examples/Planets/Get-Planet.Tests.ps1
```diff
@@ -54,8 +54,8 @@ Describe 'Get-Planet' {
         $allPlanets = Get-Planet
 
         # We count how many planets we got. And validate it by using
-        # the Should -Be assertion.
-        $allPlanets.Count | Should -Be 8
+        # the Should-Be assertion.
```
johlju: If changing the one below this:

```suggestion
        # the Should-BeCollection assertion.
```

johlju (approved):
Minor. Works as-is but maybe can showcase new things.

PR #2870, opened 2026-07-15
Surface verbose/debug output when a CI debug flag is enabled

### LINKED ISSUE #1694: Write-Verbose cannot be invoked by System.Debug from Azure DevOps

Enabling `System.Debug` in Azure DevOps allows Verbose logging within PowerShell, with the exception of the Pester Module.

This is executed via a default `windows-2019` Hosted agent.

## 3. Expected Behavior

Any output from `Write-Verbose` with Pester would be visible if the `System.Debug` flag is set in Azure DevOps.

## 4.Current Behavior

Unless `$VerbosePreference` is explicitly set, then Pester will not log anything for the Verbose log.
Taken from the following [sample](https://github.com/almmechanics/pester5).

An example of this is shown by the following invocation script and test suite:

### Wrapper.ps1

This file allows the current runtime variables to be seen from powershell before the invocation of Pester:

```
write-host ('Number of Environment Variables:{0}' -f (Get-ChildItem env:).count)
write-host ('VerbosePreference set to: {0}' -f $VerbosePreference)
write-host ('System Debug set to :{0}' -f $env:system_debug)

Import-Module pester -Verbose

write-host ('Pester Version: {0}' -f (Get-Module 'Pester').Version.ToString())

# cast from empty hashtable to get default
$configuration = [PesterConfiguration]@{}
$configuration.Run.Path = 'tests'
$configuration.TestResult.OutputPath = 'TEST-Environment.nunit.xml'
$configuration.TestResult.OutputFormat = 'NUnitXml'
$configuration.TestResult.Enabled = $False
$configuration.Run.PassThru = $True
$configuration.Run.Exit = $True

Invoke-Pester -Configuration $configuration
```

### Env.Tests.ps1

The following set of Pester tests show the Azure DevOps variables are carried across, however `Write-Verbose` logging is not available. 

```
BeforeAll {
    (Get-ChildItem env:).Name | foreach-object {Write-Verbose $_}
}

Describe "Get Environment Configuration" {
    It "All Environment variables carried across" {
        (Get-ChildItem env:).count | Should -Be 193
    }

    It "System Debug enabled" {
        $env:system_debug | should -be $True
    }

    It "Verbose Preference is set to Continue" {
        $VerbosePreference | Should -Be 'Continue'
    }
}
```

## 5. Possible Solution

Allow the value of `System.Debug` to override the `$VerbosePreference` variable within Pester

## 6. Context

A separate project to analyse AzSK output is proving difficult to debug without this capability being available.

nohwnd:
Fix #1694

## Problem

In Azure DevOps, turning on the `System.Debug` pipeline flag makes PowerShell verbose **everywhere except inside Pester tests**: a `Write-Verbose` in a test stays hidden unless the user manually sets `$VerbosePreference = 'Continue'`. Pester deliberately never mutates the user's ambient preference variables, and the `-Verbose` common parameter does not propagate into the test script blocks (they run in a separate session state without `CmdletBinding`), so the pipeline debug switch has no effect on test output. The same gap exists in GitHub Actions.

## What this does

When a known CI system has its debug switch enabled — and the user has not opted out — Pester makes the run more verbose automatically:

- **Raises `Output.Verbosity` to `Diagnostic`**, but only when the user did not set a verbosity explicitly. This mirrors how `System.Debug` makes the rest of the pipeline verbose, and also turns on Pester's own debug messages.
- **Surfaces the user's own `Write-Verbose` / `Write-Debug`** from tests and setup blocks, by setting `$VerbosePreference` / `$DebugPreference` to `Continue` in the test-container root scope. This is scoped to the run and does **not** leak into the caller's session afterwards.

The two behaviours are independent: an explicit `Output.Verbosity = 'Detailed'` is respected (not overridden), yet the test's `Write-Verbose` is still surfaced.

## Detected CI debug flags

| CI | Env var | Enabled value |
|----|---------|---------------|
| Azure DevOps (`System.Debug`) | `SYSTEM_DEBUG` | `true` (compared case-insensitively) |
| GitHub Actions (step debug logging) | `RUNNER_DEBUG` | `1` (the same signal `actions/toolkit` `core.isDebug()` uses) |

Detection lives in a small, CI-agnostic helper pair (`Get-CIDebugFlag` / `Test-CIDebugOutputEnabled` in `Pester.Utility`), mirroring the existing `TF_BUILD` / `GITHUB_ACTIONS` detection used for `Output.CIFormat`, so more CI systems can be added in one place later. It lives in `Pester.Utility` so both `Resolve-OutputConfiguration` and the runtime (`Invoke-ContainerRun`) can reuse the same detection.

## Example

A test that logs progress with `Write-Verbose`:

```powershell
# Sample.Tests.ps1
Describe 'Deploy' {
    It 'creates the resource' {
        Write-Verbose "Requesting resource from https://example.test"
        Write-Debug   "raw response: { id = 42 }"
        $result = 42
        $result | Should -Be 42
    }
}
```

**Before this change** — even with the CI debug flag on, the verbose/debug lines are hidden:

```
Describing Deploy
  [+] creates the resource 8ms
Tests Passed: 1, Failed: 0
```

**After this change** — with the CI debug flag on, the same run shows the messages (verbosity is raised to `Diagnostic`):

```
Describing Deploy
VERBOSE: Requesting resource from https://example.test
DEBUG: raw response: { id = 42 }
  [+] creates the resource 9ms
Tests Passed: 1, Failed: 0
```

### Azure DevOps

Turn on `System.Debug` — either queue the pipeline with the **Debug** system diagnostics checkbox, or set the variable in YAML:

```yaml
variables:
  system.debug: true          # sets SYSTEM_DEBUG=true for every step

steps:
  - pwsh: Invoke-Pester -Path ./tests
    displayName: Run Pester
```

### GitHub Actions

Turn on step debug logging — either re-run the job with **"Enable debug logging"**, or set the repository secret/variable `ACTIONS_STEP_DEBUG=true`. The runner then exposes `RUNNER_DEBUG=1` inside each step:

```yaml
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - shell: pwsh
        run: Invoke-Pester -Path ./tests
```

No change to the test code or the Pester invocation is required in either case.

## Opt-out

A new `Output.CIDebugOutput` option (`None`, `Auto`; default `Auto`) controls the behaviour. Set it to `None` to disable both the verbosity raise and the verbose/debug surfacing even when a CI debug flag is present:

```powershell
$config.Output.CIDebugOutput = 'None'
$config.Run.Path = './tests'
```

Choosing a verbosity explicitly keeps that choice — the run stays at `Detailed`, but the test's `Write-Verbose` is still surfaced:

```powershell
$config.Output.Verbosity = 'Detailed'   # not overridden to Diagnostic
$config.Run.Path = './tests'
```

## Tests

- `tst/functions/Output.Tests.ps1` — unit tests for `Get-CIDebugFlag`, `Test-CIDebugOutputEnabled`, and the `Resolve-OutputConfiguration` behaviour (Azure DevOps flag, GitHub Actions flag, no-CI default, explicit-verbosity not overridden, opt-out, and invalid value throws).
- `tst/Pester.RSpec.Configuration.ts.ps1` — default value, set/update, invalid-throws, and end-to-end integration tests that run a real `Invoke-Pester` and assert both the verbosity raise and the surfaced `Write-Verbose` / `Write-Debug` (including the opt-out case).

All P-tests and the affected Pester tests pass, and `Invoke-ScriptAnalyzer` on `./src` reports no new findings versus the baseline.

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -1073,8 +1073,20 @@ function Invoke-ContainerRun {
         # add OneTimeTestSetup to set variables, by having $setVariables script that will invoke in the user scope
         # and $setVariablesWithContext that carries the data as is closure, this way we avoid having to provide parameters to
         # before all script, but it might be better to make this a plugin, because there we can pass data.
+
+        # When a CI system has its debug switch enabled (e.g. Azure DevOps System.Debug or GitHub Actions
+        # runner debug logging) and the user did not opt out, surface Write-Verbose / Write-Debug from
+        # tests and setup blocks by raising the preferences in the user's root scope below. This is scoped
+        # to the run and does not leak into the caller's session afterwards. See issue #1694.
+        $surfaceCIDebugOutput = Test-CIDebugOutputEnabled -PesterPreference $PesterPreference
+
         $setVariables = {
             param($private:____parameters)
+
+            if ($____parameters.SurfaceCIDebugOutput) {
+                $VerbosePreference = 'Continue'
+                $DebugPreference = 'Continue'
+            }
 
             if ($null -eq $____parameters.Data) {
                 return
@@ -1092,8 +1104,9 @@ function Invoke-ContainerRun {
             $action = $setVariables
             $setup = $RootBlock.OneTimeTestSetup
             $parameters = @{
-                Data         = $RootBlock.BlockContainer.Data
-                Set_Variable = $SafeCommands["Set-Variable"]
+                Data                 = $RootBlock.BlockContainer.Data
+                Set_Variable         = $SafeCommands["Set-Variable"]
+                SurfaceCIDebugOutput = $surfaceCIDebugOutput
             }
 
             {
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -122,6 +122,25 @@ function notDefined {
 }
 
 
+function Get-CIDebugFlag {
+    # Returns $true when a known CI system has its debug/verbose logging switch enabled.
+    # Mirrors the CI detection used for Output.CIFormat so more CI systems can be added here later.
+    # - Azure DevOps: enabling the 'System.Debug' pipeline variable sets $env:SYSTEM_DEBUG to the
+    #   literal string 'true'.
+    # - GitHub Actions: enabling step debug logging (secret/variable ACTIONS_STEP_DEBUG=true) sets
+    #   $env:RUNNER_DEBUG to '1' inside a step. This is the same signal actions/toolkit core.isDebug() uses.
+    # -eq is case-insensitive in PowerShell, so 'True'/'TRUE'/'true' all match.
+    ('true' -eq $env:SYSTEM_DEBUG) -or ('1' -eq $env:RUNNER_DEBUG)
+}
+
+function Test-CIDebugOutputEnabled ([PesterConfiguration]$PesterPreference) {
+    # CI debug-output surfacing is enabled when the user has not opted out (Output.CIDebugOutput is
+    # not 'None') and a known CI system has its debug switch enabled. Lives here in Pester.Utility so it
+    # is available both to Resolve-OutputConfiguration and to the runtime (Invoke-ContainerRun).
+    ('None' -ne $PesterPreference.Output.CIDebugOutput.Value) -and (Get-CIDebugFlag)
+}
+
+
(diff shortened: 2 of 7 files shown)


nohwnd:
also lol on the pr description,  I should have not scolded that agent for keeping it empty :D more is not always better

Issue #2874, opened 2026-07-16
Allow different order in collections


fflaten:
### Checklist

- [x] Feature request has a meaningful title
- [x] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [x] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### Summary of the feature request

Should assertions like `Should-ContainCollection` have an option to ignore item order? Sometimes the returned order can not be guaranteed, especially in integration tests.

This is especially visible when migrating from `Should -BeIn` which checked individual items in the expected collection.

```powershell
'a',2,$null | Should -BeIn 'b',3,2,'a',$null
# pass

'b',3,2,'a',$null | Should-ContainCollection 'a',2,$null
Expected [Object[]] @('a', 2, $null) to be present in [Object[]] @('b', 3, 2, 'a', $null), but it was not there.

'b',3,2,'a',$null | Sort-Object | Should-ContainCollection ('a',2,$null | Sort-Object)
# pass but ugly
```

### How should it work?

`-IgnoreOrder` parameter?
Or close request and let this issue document the required use of `Sort-Object`.

nohwnd:
There is definitely room for adding parameters to the assertions, that is one of the reasons we have them separate. I was not sure if the default  checking in order was correct, and probably we should choose a direction, to decide if it should go strict (default) -> less strict, or the otherway around, and what is the convenience price for that.

Probalby worth asking copilot to look at all our new assertions and figure out the pattern.

Issue #2873, opened 2026-07-16
Should-Throw returns wrong exeption type


fflaten:
### Checklist

- [x] Issue has a meaningful title
- [x] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [x] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

`Should-Throw` returns exception type `ParentContainsErrorRecordException` and not the real exception in certain scenarios.

### Expected Behavior

Same behavior as `Should -Throw`

### Steps To Reproduce

```powershell
# Should-Throw fails. Wrong exception type identified
{ Get-ShouldOperator 'Demo' } | Should-Throw -ExceptionType ([System.Management.Automation.ParameterBindingException])
Expected an exception, of type [Management.Automation.ParameterBindingException] to be thrown, but the exception type was [Management.Automation.ParentContainsErrorRecordException].

# Should -Throw works. Matches exception basetype
{ Get-ShouldOperator 'Demo' } | Should -Throw -ExceptionType ([System.Management.Automation.ParameterBindingException])
```

### Describe your environment

Pester version     : 6.0.0 C:\Users\frode\Documents\PowerShell\Modules\pester\6.0.0\Pester.psm1
PowerShell version : 7.6.3
OS version         : Microsoft Windows NT 10.0.26200.0

### Possible Solution?

Change where `Get-ErrorObject` finds the exception. The exceptions are different because `Should-Throw` replaced `& {}` with `{}.InvokeWithContext(..`

```powershell
# Should -Throw, current (works)
try { & { Get-ShouldOperator 'Demo' } } catch { Write-Host $_.Exception.GetType().Name }

ParameterBindingValidationException

# Should-Throw, current (wrong)
try { $null = { Get-ShouldOperator 'Demo' }.InvokeWithContext($null, $eap, $null) 2>&1 } catch { Write-Host $_.Exception.InnerException.ErrorRecord.Exception.GetType().Name }

ParentContainsErrorRecordException

# Should-Throw, innerexception directly (works)
try { $null = { Get-ShouldOperator 'Demo' }.InvokeWithContext($null, $eap, $null) 2>&1 } catch { Write-Host $_.Exception.InnerException.GetType().Name }

ParameterBindingValidationException
```

However it needs to be conditional, because we need the current way for ex. `Write-Errror -ErrorAction Stop`

```powershell
[...]

Issue #2872, opened 2026-07-16
String difference error broken when ErrorView is ConciseView


fflaten:
### Checklist

- [x] Issue has a meaningful title
- [x] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [x] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

In PowerShell 7.1+ the string difference error from `Should-BeString`, `Should -Be/BeExactly` is broken using the default ErrorView (`ConciseView`)

### Expected Behavior

A clean and readable error message

### Steps To Reproduce

```powershell
# Same error with 'abc' | Should -Be 'Two' etc.
PS /workspaces/Pester> 'abc' | Should-BeString 'ab'
InvalidResult: 
Line |
   1 | 'abc' | Should-BeString 'ab'
     | Expected strings to be the same, but they were different. Expected length: 2 Actual length:   3 Strings differ at index 2. Expected: 'ab' But was:  'abc'       
     | --^

# Expected
PS /workspaces/Pester> $ErrorView = 'NormalView'                                         
PS /workspaces/Pester> 'abc' | Should-BeString 'ab'
Expected strings to be the same, but they were different.
Expected length: 2
Actual length:   3
Strings differ at index 2.
Expected: 'ab'
But was:  'abc'
          --^
At /workspaces/Pester/src/functions/assert/Common/Invoke-AssertionFailed.ps1:58 char:9
+         throw $errorRecord
+         ~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidResult: (System.Collections.…ring,System.Object]:Dictionary`2) [], Exception
+ FullyQualifiedErrorId : PesterAssertionFailed
```

### Describe your environment

Pester version     : 6.1.0-alpha2 /workspaces/Pester/bin/Pester.psm1                                                    
PowerShell version : 7.6.2
OS version         : Unix 6.18.35.2

### Possible Solution?

ConciseView trims newlines, so maybe fallback to simple `Should-Be` error like `Expected [string] 'ab', but got [string] 'abc'.`? 

Override `$ErrorView` to `NormalView` if `ConciseView`?

fflaten:
Side note: Should the caret point at the last matching character or the first difference? Classic vs new assertions behave different

```powershell
# 'abc' | Should-BeString 'ab'
Expected: 'ab'
But was:  'abc'
          --^
[...]

PR #2875, opened 2026-07-16
Migrate tests to v6 assertions

fflaten:
## PR Summary
Start migrating tests in the repo to Should-* assertions.

Not covered: 
- Tests in assert and assertions folders which test themselves and should use `Verify-*`.
- `Should -Not -Throw`. I like the errors with *-Because* 🙂 
- `Should -Exist`. No equivalent and I like the path-specific error message.
- `Should -Invoke`. `Should-Invoke` syntax has own tests, but not `Should -Invoke` which relies on `Mock.Tests.ps1`

## PR Checklist

- [x] PR has meaningful title


fflaten:
Identified a few issues during this dogfooding: #2872, #2873 and #2874 + notes in not covered section above

Pester.Tests.ps1 should also be cleaned up. PSv3 guards and use of script variables (= test order dependency) #2877

PR #2879, opened 2026-07-16
WIP - Debug PS7_Windows_Server2022 CI

fflaten:
## PR Summary

Figuring out why "captures debug output and replays it interleaved with each file's tests" test did not output anything. Did it skip it like only 5.1 should?


fflaten:
Brainfart. Output in PS7 Win2025 was only because it failed (fixed by #2880). Working as expected

PR #2880, opened 2026-07-16
Fix flaky test on Windows CI

fflaten:
## PR Summary
Fixing a flaky output test because Windows might use seconds and not milliseconds.

Includes minor cleanup in output to remove whitespace between time and time suffix. Future PR should consider extracting `Get-HumanTime` from `Output.ps1` to `Pester.Utility.ps1` to get consistency.

## PR Checklist

- [x] PR has meaningful title

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -999,7 +999,7 @@ function Invoke-ContainerDiscovery {
 
     $root.DiscoveryDuration = $perContainerDiscoveryDuration.Elapsed
     if ($PesterPreference.Debug.WriteDebugMessages.Value) {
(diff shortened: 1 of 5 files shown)


fflaten:
> Future PR should consider extracting `Get-HumanTime` from `Output.ps1` to `Pester.Utility.ps1` to get consistency.

@nohwnd Thoughts? Any reason to keep some output hardcoded to milliseconds, or should I implement?

nohwnd:
We are probably already outputting human time in all the normal output, in  the diag logs I am okay if it stays using just ms. Less overhead (no function call) and less complicated to write the message.

PR #2887, opened 2026-07-17
Fix mock bootstrap proxy for cmdlets with no DefaultParameterSetName

### LINKED ISSUE #1531: Cannot execute mock for Get-PackageSource cmdlet
In order to fix  `Get-PackageSource` cmdlet, another level of indirection was introduced which broke other things. 

Original issue and PR:
#1308
#1317 

Issue it caused and revert PR:
#1519 
#1528

nohwnd:
Fix #1531

Mocking a cmdlet with multiple non-mandatory parameter sets and no `DefaultParameterSetName` (e.g. `Get-PackageSource` with its `NuGet` and `PowerShellGet` dynamic sets) failed at call time with:

```
ParameterBindingException: Parameter set cannot be resolved using the specified named parameters.
One or more parameters issued cannot be used together or an insufficient number of parameters were provided.
```

When `Create-MockHook` builds the bootstrap proxy it always injects `PositionalBinding=$false` into the generated `[CmdletBinding()]`. For a cmdlet whose dynamic parameters add several named sets with no default, PowerShell has no set to fall back on when the proxy is called with no arguments, hence the ambiguity. The real cmdlet handles this in C# and is fine, the PS-level proxy is not.

Fix: when `$metadata.DefaultParameterSetName` is null or empty, inject `DefaultParameterSetName='__AllParameterSets'` into the proxy's `[CmdletBinding()]`. Cmdlets that already have a default set are untouched, the existing value is preserved and the injection is skipped.

Added a regression test in `Mock.Tests.ps1` that mocks `Get-PackageSource` (ships with PS7) and calls it with no arguments, skipped when the cmdlet is not available.

### DIFF
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -88,6 +88,7 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
             # default makes the no-argument call succeed, matching the real cmdlet's behaviour. (#1531)
             if ([string]::IsNullOrEmpty($metadata.DefaultParameterSetName)) {
                 $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, "DefaultParameterSetName='__AllParameterSets'")
+                $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, ',')
             }
             $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, 'PositionalBinding=$false')
         }
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -79,18 +79,15 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
         }
         $cmdletBinding = [Management.Automation.ProxyCommand]::GetCmdletBindingAttribute($metadata)
         if ($contextInfo.Command.CommandType -eq 'Cmdlet') {
+            if ($cmdletBinding -ne '[CmdletBinding()]') {
+                $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, ',')
+            }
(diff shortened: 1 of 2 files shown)


src/functions/Mock.ps1
```diff
@@ -79,6 +79,16 @@ function Create-MockHook ($contextInfo, $InvokeMockCallback) {
...
+            # default makes the no-argument call succeed, matching the real cmdlet's behaviour. (#1531)
+            if ([string]::IsNullOrEmpty($metadata.DefaultParameterSetName)) {
+                if ($cmdletBinding -ne '[CmdletBinding()]') {
+                    $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, ',')
+                }
+                $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, "DefaultParameterSetName='__AllParameterSets'")
+            }
             if ($cmdletBinding -ne '[CmdletBinding()]') {
                 $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, ',')
             }
```
fflaten:
```suggestion
            if ($cmdletBinding -ne '[CmdletBinding()]') {
                $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, ',')
            }
            # When the cmdlet has no explicit DefaultParameterSetName and its dynamic parameters
            # introduce multiple named parameter sets, PowerShell cannot resolve which set applies
            # when the mock is called without arguments. Injecting '__AllParameterSets' as the
            # default makes the no-argument call succeed, matching the real cmdlet's behaviour. (#1531)
            if ([string]::IsNullOrEmpty($metadata.DefaultParameterSetName)) {
                $cmdletBinding = $cmdletBinding.Insert($cmdletBinding.Length - 2, "DefaultParameterSetName='__AllParameterSets'")
            }
```

? 🙂

PR #2890, opened 2026-07-17
Clarify Should -Not -Invoke failure message

### LINKED ISSUE #1832: Should -Not -Invoke message is a but unclear
A different message is displayed for `Should -Not -Invoke` and `Should -Invoke -Time 0 -Exactly`. The message `Expected Test-Me not to be called exactly 1 times` is a but unclear to me.

PowerShell version : 5.1.14393.3866

## Steps to reproduce
```powershell
Describe 'test the Pester error message' {
    BeforeAll {
        Function Test-Me { 1 }
        Mock Test-Me
        Test-Me
    }
    It 'should ' {
        Should -Not -Invoke test-Me -Scope Describe
        # Expected Test-Me not to be called exactly 1 times
    }
    It 'should ' {
        Should -Times 0 -Exactly -Invoke test-Me -Scope Describe
        # Expected Test-Me to be called 0 times exactly but was called 1 times
    }
}
```

## Expected Behavior
For ` Should -Not -Invoke test-Me -Scope Describe` it might be better to say:
> Expected Test-Me not to be called but it was called 1 time

## Current Behavior
See comments above

nohwnd:
Fix #1832

A plain `Should -Not -Invoke` (no `-Times`/`-Exactly`) failed with:

> Expected Test-Me not to be called exactly 1 times, but it was ...

That is confusing for a negative assertion, the user never asked for "exactly 1", `1` is just the internal default for `-Times`. `Should -Not -Invoke Foo` just means "Foo should not have been called at all".

The negative branch of `Should-InvokeInternal` now detects the plain case (`-Times` not bound, no `-Exactly`) and words it naturally, with correct singular/plural:

> Expected Test-Me not to be called, but it was called 1 time

> Expected Test-Me not to be called, but it was called 2 times

The `-Times` and `-Exactly` wordings are left as they were, so explicit-count assertions and the positive-assertion messages are unaffected. Updated `Mock.Tests.ps1` for the new message and added coverage for the plural case and `-Because`. 252 passed, no new analyzer findings.

### DIFF
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -483,6 +483,11 @@ function Should-InvokeVerifiableInternal {
     }
 }
 
+function Format-TimesCalled ([int] $Count) {
+    # Format a call count with the grammatically correct noun, e.g. "1 time" or "2 times".
+    if ($Count -eq 1) { "$Count time" } else { "$Count times" }
+}
+
 function Should-InvokeInternal {
     [CmdletBinding(DefaultParameterSetName = 'ParameterFilter')]
     [OutputType([Pester.ShouldResult])]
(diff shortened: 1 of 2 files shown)


src/functions/Mock.ps1
fflaten, line 1: Can we dedupe this? Should also use time vs times in the remaining messages.

nohwnd: Done in 14731a2. Extracted the repeated ShouldResult construction into a single builder (each branch now just sets Expected/Actual/But), and added a Format-TimesCalled helper so every Invoke failure message uses "time" vs "times" correctly (e.g. "1 time" vs "2 times"). Updated the affected tests; ./test.ps1 -File ./tst/functions/Mock.Tests.ps1 -SkipPTests is green (252 passed).

PR #2898, opened 2026-07-17
Add -DefaultValueType to Should -HaveParameter

### LINKED ISSUE #1888: Should -HaveParameter should check DefaultParameterType

nohwnd:
Fix #1888

Adds `-DefaultValueType` to both `Should -HaveParameter` and the assert-style `Should-HaveParameter`.

The issue wants to tell an expression default apart from a string literal that has the same `-DefaultValue` text. The catch is that the two are the *same type*: `(Get-DefaultPath)` returns a path string and `'(Get-DefaultPath)'` is a string, so a type alone cannot separate them. The real difference is that one is a **computed expression** and the other is a **literal**.

So `-DefaultValueType` takes a real .NET type, given as a `[type]` or a type name the same way `-Type` is given, matched against the type of a literal default. For a computed default whose type is not known until it runs, you pass the special value `Expression`:

```powershell
function Connect-Thing {
    param(
        [string] $Url      = (Get-DefaultUrl),    # expression, evaluated at call time
        [string] $Fallback = '(Get-DefaultUrl)',  # literal string that looks identical
        [int]    $Retries  = 3,
        [bool]   $Force    = $false
    )
}

$cmd = Get-Command Connect-Thing
$cmd | Should -HaveParameter Retries -DefaultValueType ([int])       # or -DefaultValueType int
$cmd | Should -HaveParameter Force   -DefaultValueType ([bool])
$cmd | Should -HaveParameter Fallback -DefaultValueType String       # literal string

# The #1888 distinction, same -DefaultValue string, different kind of default:
$cmd | Should -HaveParameter Url      -DefaultValue '(Get-DefaultUrl)' -DefaultValueType Expression
$cmd | Should -HaveParameter Fallback -DefaultValue '(Get-DefaultUrl)' -DefaultValueType String
```

Notes:

- The type can be a `[type]` (`([bool])`) or a type name string (`bool`, `int`, `string`, `datetime`), same as `-Type`. An unknown type name throws, and points you at `-DefaultValueType Expression`.
- `Expression` covers any computed default whose type is not statically known, e.g. `(Get-Date)`, `[datetime]::Now`, `$someVariable`. You cannot assert a concrete type for those, because the type is only known once the code runs.
- `$true`/`$false` are `[bool]`, a number is `[int]`/`[double]`, `{ ... }` is `[scriptblock]`, `@{ }` is `[hashtable]`. A `$null` default has no type. An interpolated string like `"$x bar"` is `[string]`.
- Failure messages read as, for example, `the default value to be of type [System.Int32], but the default value was of type [System.String]`, or `the default value to be an expression, but the default value was of type [System.String]`.

`HaveParameter.ps1` and `Should-HaveParameter.ps1` plus tests for both. All existing `-HaveParameter` behavior is unchanged. `test.ps1` for the affected files: 202 passed, no new analyzer findings.

### DIFF
--- a/src/functions/assert/Parameter/Should-HaveParameter.ps1
+++ b/src/functions/assert/Parameter/Should-HaveParameter.ps1
@@ -16,13 +16,14 @@
     The default value of the parameter to check. E.g. "https://example.com"
 
     .PARAMETER DefaultValueType
-    The kind of the parameter's default value. This distinguishes how the default is written, which
-    -DefaultValue (a string comparison) cannot: for example `$Path = (Get-Date)` is an `Expression`
-    while `$Path = '(Get-Date)'` is a `String`. Valid values:
-    `String` (literal string, e.g. 'abc' or "abc"), `InterpolatedString` (e.g. "$x/bar"),
-    `Number` (numeric literal, e.g. 1), `Variable` (e.g. $true, $false, $null, $var),
-    `Expression` (e.g. (Get-Date)), `ScriptBlock` (e.g. { Get-Date }), `Array` (e.g. @(1, 2) or 1, 2),
-    `Hashtable` (e.g. @{}). Matching is case-insensitive.
+    The value type of the parameter's default. Valid values:
+    `Boolean` (`$true`, `$false`), `Null` (`$null`), `Number` (e.g. 1, 1.5), `String` (a literal or
+    interpolated string, e.g. 'abc' or "$x bar"), `ScriptBlock` (e.g. { Get-Date }), `Array` (e.g. @(1, 2)),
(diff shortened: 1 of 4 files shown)


nohwnd:
doubt that people want to write StringConstantExpressionAst as the expected type of the parameter. need to iterate

PR #2905, opened 2026-07-18
Move CI to GitHub Actions, remove the Azure pipeline

nohwnd:
Moves CI to GitHub Actions and deletes `azure-pipelines.yml`. Releasing stays on Azure DevOps (`azure-pipelines-publish.yml`), it needs the signing service connection and the publishing variable groups.

Why now: Azure DevOps public projects are being retired, existing ones convert to private in 2027 and the free grant drops from 10 parallel jobs to 1. GitHub Actions is free for public repos with 20.

The workflow builds once on `windows-latest`, then runs the same 8 leg matrix Azure ran: PowerShell 7 on ubuntu/macOS/windows (latest plus one previous GA), Windows PowerShell 5.1 on windows (latest plus previous). Coverage shows in each run's job summary and in a consolidated table, and uploads to Codecov with a flag per leg. One `Done` job gates the matrix, same idea as the Azure `Done` stage, so branch protection requires one check instead of listing all 8 legs.

@fflaten the test report is fixed. It moved out of the matrix into `test-report.yml`, which is the setup dorny/test-reporter recommends for public repos, and also the reason nothing was showing: a run triggered by a pull request from a fork gets a read only token and cannot create a check run. A `workflow_run` workflow runs in the base repo and can. It lists only failed suites and tests now. One catch, `workflow_run` only fires for the copy of the workflow on the default branch, so the report cannot be seen on this PR, it starts working once this is merged.

Before merging, main still requires `Pester PR (Build build)` and `Pester PR (Done Done)`. Neither can report once the pipeline file is gone, so the required check has to change to `Done` at the same time, and the Pester PR pipeline in Azure DevOps should stop triggering on pull requests.

Next: the same workflow on `rel/5.x.x`, and a `Done` job in the v4 workflow on `rel/4.x.x`, then one branch policy requiring `Done` covers main and both release branches.

Images are GA only, floating `*-latest` plus one pinned previous GA. `macos-14` is deprecating (removed Nov 2026) and will need swapping before then.

🤖


fflaten:
Add https://github.com/dorny/test-reporter to the test matrix jobs?
Could also add dependabot for github-actions (at least major versions) to keep up with the Node.js deprecations.

> Why now: Azure DevOps public projects are being retired (existing ones convert to private in 2027, dropping the free grant from 10 parallel jobs to 1)

Didn't know. 🙁  Good thing we don't use exclusive features here.

fflaten:
Test result summary not showing, not sure why.

Looks like we need to extract the test-report to a follow-up workflow into separate workflow to handle forks (https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories).

And maybe limit the summary to failed only using `list-suites`  and `list-tests`?

PR #2907, opened 2026-07-18
Remove unused internal functions

nohwnd:
Draft, and stacked on #2906. It branches off that PR, so until #2906 merges the diff here shows both changes, review this one after that lands and it rebases clean.

While generating a full coverage report I noticed a chunk of the missed lines are not gaps, they are internal functions that nothing calls anymore. This removes 51 of them. Each one has no callers in `src` or `tst` (checked case insensitively, ignoring comments and its own definition), and none are exported.

By group:

- **Old v4 assertion messages.** The orphaned `Should<Name>FailureMessage` / `NotShould<Name>FailureMessage` pairs for BeGreaterThan, BeIn, BeLessThan, BeLike, BeLikeExactly, BeOfType, BeTrueOrFalse, Contain and HaveCount, plus `Get-FailureMessage`. The operators build the message inline now. Be, Match and FileContentMatch still call theirs so they stay. Exist and Throw also stay, because `Should.Tests.ps1` asserts their message functions exist (it builds the names as strings, so a plain grep does not see that use).
- **Unfinished pretty printer in Format.ps1.** `Format-Object`, `Format-Hashtable`, `Format-Dictionary`, `Get-DisplayProperty`. `Format-Nicely` returns the value as a string and the advanced path is commented out.
- **Orphaned helpers** across Mock, RSpec, Runtime, Coverage, Debugging and Utility: `Run-Test`, `Merge-CommandCoverage`, `Show-ParentList`, `Add-MockBehavior`, `FindMock`, `LastThat`, `Test-IsClosure`, `Reset-ConflictingParameters`, `New-BlockWithoutParameterAliases`, `New-OneTimeBlockSetup`, `Set-CurrentBlock`, `Reset-TestSuiteTimer`, `New-DiscoveredBlockContainerObject`, `Get-SessionStateHint`, `Assert-RunInProgress`, and the `Pester.Utility` dictionary helpers.

## Verification

Full `test.ps1` run passes (P tests and Pester tests, 2606 passed, 0 failed). Every touched file parses.

## Notes for review

- I removed by AST extent, so the diff is whole functions only.
- `Sort-Property` in Format.ps1 is now only reachable from the removed `Get-DisplayProperty`, so it is dead too, I left it in this pass, easy to drop as a follow up.
- A couple of things surfaced that are worth a look but out of scope here: `PostProcess-RspecTestRun` and `Remove-RSpecNonPublicProperties` are defined with lowercase `Rspec` but called with `RSpec`/`RSPec`, they only work because PowerShell is case insensitive. `ConvertTo-PesterResult` has no production caller, it is kept because a test uses it as the canary for "internals do not leak to the SUT".

### DIFF
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -345,7 +345,6 @@ function Contain-AnyStringLike ($Filter, $Collection) {
     return $false
 }
 
-# TODO: Remove?
 function View-Flat {
     [CmdletBinding()]
     param (
--- a/src/Pester.Utility.ps1
+++ b/src/Pester.Utility.ps1
@@ -1,4 +1,4 @@
-﻿function or_ {
+function or_ {
     [CmdletBinding()]
     param (
         [Parameter(Mandatory = $true, Position = 0)]
@@ -39,25 +39,6 @@ function tryGetProperty_ {
     # }
 }
 
-function trySetProperty_ {
-    [CmdletBinding()]
-    param (
-        [Parameter(Position = 0)]
-        $InputObject,
-        [Parameter(Mandatory = $true, Position = 1)]
-        $PropertyName,
-        [Parameter(Mandatory = $true, Position = 2)]
-        $Value
-    )
-
-    if ($null -eq $InputObject) {
-        return
-    }
-
-    $InputObject.$PropertyName = $Value
-}
-
-
 # combines collections that are not null or empty, but does not remove null values
 # from collections so e.g. combineNonNull @(@(1,$null), @(1,2,3), $null, $null, 10)
 # returns 1, $null, 1, 2, 3, 10
@@ -72,7 +53,6 @@ function combineNonNull_ ($Array) {
         }
     }
 }
-
 
 filter selectNonNull_ {
     param($Collection)
@@ -121,7 +101,6 @@ function notDefined_ {
     $null -eq ($ExecutionContext.SessionState.PSVariable.GetValue($Name))
 }
 
-
 function Get-CIDebugFlag {
     # Returns $true when a known CI system has its debug/verbose logging switch enabled.
     # Mirrors the CI detection used for Output.CIFormat so more CI systems can be added here later.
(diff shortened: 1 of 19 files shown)


src/functions/Pester.Debugging.ps1
```diff
@@ -1,4 +1,4 @@
-﻿function Count-Scopes {
+function Count-Scopes {
```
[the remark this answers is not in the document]
fflaten: Time to give in and rename? 😄

fflaten:
Quick reminder that we support `[ExcludeFromCodeCoverage()]` if we want to keep some of these while fixing CC.

> New-OneTimeBlockSetup

Both `OneTimeBlockSetup/-Teardown` functions + public properties in `Pester.Block` are dead code AFAIK. `EachBlockSetup` and `EachBlockTeardown` are only called by runtime tests also. Any plans for them or should they've been removed in 6.0.0 (should be fine in 6.1.0 as well - very minor breaking change)

src/functions/Pester.Debugging.ps1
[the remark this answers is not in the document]
nohwnd: Given in 🙂 renamed to `Get-ScopeCount`.

nohwnd:
Yeah they're dead. The `New-*Block*` functions are only called from the runtime tests, so the four `Pester.Block` properties are never set in a real run and the two reads in `Invoke-Test` are no-ops. Removed all of it in #2908 (stacked on this), the functions, the properties and the reads. Kept the plugin block lifecycle steps (`OneTimeBlockSetupStart` etc.), those are live. All truly dead so I just deleted them instead of using `[ExcludeFromCodeCoverage()]`.

src/functions/Mock.ps1
```diff
@@ -2164,49 +2032,6 @@ function Get-ScriptBlockAST {
 }
 
 # TODO: Remove?
```
nohwnd: remove the todo

nohwnd:
```suggestion
```

src/Pester.Utility.ps1
```diff
@@ -437,23 +346,6 @@ function Contain-AnyStringLike ($Filter, $Collection) {
 }
 
 # TODO: Remove?
```
nohwnd:
```suggestion
```

Issue #1219, opened 2019-01-19
v5: BeforeEach/AfterEach -Context?


nvarscar:
## 1. General summary of the issue

I tend to write a chain of different `It` tests that expect a certain resource to be available during the whole execution of the `Context` in persistent state and then get removed once the tests in this particular context are over. Right now, I'm using `BeforeAll` and `AfterAll` for each `Context` to create/remove such resources. It would be much easier to have a `BeforeEach -Context` so that I could define it on a `Describe` level just once.

## 2. Describe Your Environment

Powershell 5.1/Core, Windows and Linux

## 3. Desired Behavior

```powershell
Describe "my tests" {
  BeforeEach -Context {
    # create new file here
  }
  AfterEach -Context {
    # remove file here
  }
  Context "context 1" {
    It "it 1" {
      # doing some activities on the file
      $true | Should -Be $true
    }
    It "it 2" {
      # continuing working on the same file and expecting it to have results from "it 1"
      $true | Should -Be $true
    }
  }
  Context "context 2" {
    It "it 1" {
      # doing some activities on the file
      $true | Should -Be $true
    }
    It "it 2" {
      # continuing working on the same file and expecting it to have results from "it 1"
      $true | Should -Be $true
    }
  }
}
```

## 4.Current Behavior

```powershell
Describe "my tests" {
  Context "context 1" {
    BeforeAll {
      # create new file here
    }
    AfterAll {
      # remove file here
    }
    It "it 1" {
      # doing some activities on the file
      $true | Should -Be $true
    }
    It "it 2" {
      # continuing working on the same file and expecting it to have results from "it 1"
      $true | Should -Be $true
    }
  }
  Context "context 2" {
    # defining all the same logic again
    BeforeAll {
      # create new file here again
    }
    AfterAll {
      # remove file here again
    }
    It "it 1" {
      # doing some activities on the file
      $true | Should -Be $true
    }
    It "it 2" {
      # continuing working on the same file and expecting it to have results from "it 1"
      $true | Should -Be $true
    }
  }
}
```

## 5. Possible Solution

As suggested, add something like a `-Context` switch to the BeforeEach/AfterEach blocks so that they only trigger on Contexts

## 6. Context

Want to get rid of excessive code and making changes on 10 different places instead of a single block.

nohwnd:
[...]

PR #2908, opened 2026-07-18
Remove vestigial block setup/teardown

nohwnd:
Draft, stacked on #2907 (which is stacked on #2906). Follow up to Frode's comment on #2907.

The block level setup/teardown is dead. `New-EachBlockSetup`, `New-EachBlockTeardown` and `New-OneTimeBlockTeardown` (and the never set `New-OneTimeBlockSetup` that #2907 already removed) are only ever called from the runtime tests. Real `BeforeAll` / `BeforeEach` / `AfterAll` / `AfterEach` go through the `*Test*` variants (`New-OneTimeTestSetup`, `New-EachTestSetup`, ...), so in a real run the `Block.EachBlockSetup` / `OneTimeBlockSetup` / `EachBlockTeardown` / `OneTimeBlockTeardown` properties are always null and the two reads in `Invoke-Test` are no-ops.

Removes:
- the three functions
- the four `ScriptBlock` properties on `Pester.Block`
- the two always null reads in `Invoke-Test`
- the runtime tests that exercised them. The two data availability tests keep their teardown check by switching to `New-OneTimeTestTeardown`.

The plugin block lifecycle (`OneTimeBlockSetupStart`, `EachBlockTeardownEnd`, ...) is a separate, live mechanism and is untouched.

These are genuinely dead so I deleted them rather than reaching for `[ExcludeFromCodeCoverage()]`.

Minor breaking change, fine for 6.1.0.

## Verification

Full `test.ps1` run passes.

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -26,10 +26,6 @@ else {
 # 'New-EachTestTeardown'
 # 'New-OneTimeTestSetup'
 # 'New-OneTimeTestTeardown'
-# 'New-EachBlockSetup'
-# 'New-EachBlockTeardown'
-# 'New-OneTimeBlockSetup'
-# 'New-OneTimeBlockTeardown'
 # 'Invoke-Test',
 # 'Find-Test',
 # 'Invoke-PluginStep'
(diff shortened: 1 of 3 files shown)


fflaten:
About OneTimeBlock*, see #1219. Is `BeforeAll -Context` still planned? If so, we can't remove them after all.

nohwnd:
Good point, let's keep them. The `EachBlockSetup`/`EachBlockTeardown` side is actually already executed in `Invoke-Test` (the outer setup/teardown around each child block), it just has no public way to set it, so it is scaffolding for #1219 rather than dead. `OneTimeBlock*` is not wired at runtime yet. Closing this, I'll look at wiring up #1219 (`BeforeEach`/`AfterEach -Context`) instead.

Issue #2657, opened 2025-08-21
Pester clears test drives inefficiently


splatteredbits:
### Checklist

- [x] Issue has a meaningful title
- [x] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [x] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### What is the issue?

After each Context block runs, Pester clears the test drive. Unfortunately, it does this really inefficiently by calling `Remove-Item` on each item in the test drive. One of my test fixtures downloads and unpackages multiple copies of Node.js, each which contains over 5,000 files. The way Clear-TestDrive removes files, it takes about 30 seconds to delete all the files (on Linux specifically). If it called `Remove-Item` instead, it would take less than a second.

### Expected Behavior

I would expect the test drive to be cleared way more efficiently.

### Steps To Reproduce

This is the closest I can get to a reproduction. In this example, it takes pester around 4 seconds to clear the test drive. In my actual tests, it takes Pester around 25 to 30 seconds. 

```powershell

BeforeAll {
    function InitNode
    {
        param(
            [String] $In
        )

        Write-Verbose "$(Get-Date)  ${In}" -Verbose
        $pkgPath = Join-Path -Path $In -ChildPath 'node-v22.18.0-linux-x64.tar.xz'
        Invoke-WebRequest -Uri 'https://nodejs.org/dist/v22.18.0/node-v22.18.0-linux-x64.tar.xz' `
                          -OutFile $pkgPath
        tar -xJf $pkgPath -C $In

        $pkgPath = Join-Path -Path $In -ChildPath 'node-v16.20.2-linux-x64.tar.xz'
        Invoke-WebRequest -Uri 'https://nodejs.org/dist/v16.20.2/node-v16.20.2-linux-x64.tar.xz' `
                          -OutFile $pkgPath
        tar -xJf $pkgPath -C $In

        $numv22 = 1
        for ($i = 0; $i -lt $numv22; ++$i)
        {
            $testDirPath = Join-Path -Path $In -ChildPath $i
            New-Item -Path $testDirPath -ItemType Directory
            $linkPath = Join-Path -Path $testDirPath -ChildPath '.node'
            $linkTarget = Join-Path -Path $In -ChildPath 'node-v22.18.0-linux-x64'
            New-Item -Path $linkPath -ItemType SymbolicLink -Value $linkTarget
[...]

PR #2917, opened 2026-07-19
Clear TestDrive with .NET IO instead of per-item Remove-Item

nohwnd:
`Clear-TestDrive` ran `Remove-Item -Path $item -Force -Recurse -ErrorAction Ignore` for every new root item after each block. Two problems with that:

- `Remove-Item` costs ~470us per item where `[IO.Directory]::Delete` / `[IO.File]::Delete` cost ~85us. On a TestDrive-heavy suite the cleanup was the single hottest line in the profile.
- `-Path` treats the path as a wildcard pattern. A file named like `file[1].txt` never matched its own name, was never deleted, and leaked across block teardowns. The reverse is also possible, a name with `*` in it could over-delete. The .NET calls treat the path literally, so this is also a small bug fix.

Symlink semantics stay the same: recursive `Directory.Delete` removes reparse points without following them into their target, which is what `Remove-TestDrive` has already been doing for the whole drive with the exact same call since 2021. I verified this empirically on macOS with dir symlinks, file symlinks, dangling links, the PowerShell/PowerShell#621 problematic link set, and a symlink cycle, the linked target content survives in all cases. Anything that throws (read-only files on Windows, locked files) falls back to `Remove-Item -LiteralPath -Force -Recurse -ErrorAction Ignore`, keeping the old behavior for those, just slower on that rare path.

A ride-along in `New-RandomTempRegistry`: `Join-Path` replaced with plain string interpolation, registry provider paths always use backslash and the base key has no trailing separator, so the result is identical and it skips a cmdlet invocation per new registry key.

One heads-up: the symlink tests in `TestDrive.Tests.ps1` are entirely commented out (they needed admin on Windows), so nothing in CI exercises symlink cleanup today. Might be worth re-enabling them at least on Linux/macOS runners in a separate PR.

Verification: full suite green on macOS (inline and dot-sourced build), `TestDrive.Tests.ps1` including the Clear-TestDrive #2657 regression suite and `Pester.TestDrive.ts.ps1` pass.

Related perf PRs: #2914, #2915, #2916, #2918. Measured together they cut ~21% from four representative workloads (inline build, macOS).

🤖

### DIFF
--- a/src/functions/TestRegistry.ps1
+++ b/src/functions/TestRegistry.ps1
@@ -95,7 +95,8 @@ function Invoke-TestRegistryWithRetry {
 function New-RandomTempRegistry {
     do {
         $tempPath = Get-TempRegistry
-        $Path = & $SafeCommands['Join-Path'] -Path $tempPath -ChildPath ([IO.Path]::GetRandomFileName().Substring(0, 4))
+        # registry provider paths always use backslash; plain interpolation avoids a per-call cmdlet invocation
+        $Path = "$tempPath\$([IO.Path]::GetRandomFileName().Substring(0, 4))"
         # Test-Path is a read operation and can throw the same transient IOException as the
         # New-Item write below, so retry it the same way.
         $keyAlreadyExists = Invoke-TestRegistryWithRetry { & $SafeCommands['Test-Path'] -Path $Path -PathType Container }
--- a/src/functions/TestDrive.ps1
+++ b/src/functions/TestDrive.ps1
@@ -131,12 +131,29 @@ function Clear-TestDrive {
             [System.StringComparer]::OrdinalIgnoreCase)
 
         # Only delete "root" new items (those whose parent directory is not also a new item).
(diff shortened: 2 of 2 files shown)


src/functions/TestRegistry.ps1
```diff
@@ -95,7 +95,8 @@ function Invoke-TestRegistryWithRetry {
 function New-RandomTempRegistry {
     do {
         $tempPath = Get-TempRegistry
-        $Path = & $SafeCommands['Join-Path'] -Path $tempPath -ChildPath ([IO.Path]::GetRandomFileName().Substring(0, 4))
+        # registry provider paths always use backslash; plain interpolation avoids a per-call cmdlet invocation
+        $Path = "$tempPath\$([IO.Path]::GetRandomFileName().Substring(0, 4))"
```
fflaten, line 99: Not really a hot path, but sure 👍

nohwnd: got some free fable tokens, told it to optimize (especially the hot path), looks like most of the optimizations are microscopic.

PR #2918, opened 2026-07-19
Skip unused work in discovery post-processing and output

nohwnd:
Three changes on the discovery and post-processing side:

- The WriteScreen plugin flattened the whole discovered tree with `View-Flat` on `DiscoveryEnd`, but the result is only read when `Run.SkipRun` is set or verbosity is `Detailed`/`Diagnostic`. In the default run it was computed and thrown away, the cost grows with the number of discovered tests. It is now computed only when one of those two guarded blocks can actually read it.
- The RSpec decoration walk (`Add-RSpecTestObjectProperties` per test, run once per container after it finishes) and `Remove-RSpecNonPublicProperties` used `Fold-Container`/`Fold-Run`, which pay a scriptblock dispatch per item plus an advanced-function recursion per block. Both are now direct recursive loops that visit the same items in the same pre-order (tests of a block before its child blocks). The decoration body itself is unchanged and stays in one place, the walker just calls it.
- `Format-Nicely2` called the `Is-DecimalNumber`, `Is-ScriptBlock`, `Is-Collection` and `Is-Value` helpers on every formatted value, and it runs for every `<template>` name expansion of data-driven tests. The type checks are inlined (`Is-Collection` is the same `LanguagePrimitives::GetEnumerator` test), the helpers stay defined for their other callers.

Verification: full suite green on macOS (inline and dot-sourced build), `Pester.RSpec.ts.ps1`, `Pester.RSpec.Parallel.ts.ps1` (covers the parallel decoration path, which changed too) and `Pester.RSpec.Output.ts.ps1` pass in fresh processes. Result object shape is covered by `Pester.RSpec.ResultObject.ts.ps1`, also green.

Related perf PRs: #2914, #2915, #2916, #2917. Measured together they cut ~21% from four representative workloads (inline build, macOS).

🤖

### DIFF
--- a/src/Format2.ps1
+++ b/src/Format2.ps1
@@ -122,11 +122,13 @@ function Format-Nicely2 ($Value, [switch]$Pretty, [int]$Depth = 0) {
         return Format-Type2 -Value $Value
     }
 
-    if (Is-DecimalNumber -Value $Value) {
+    # Inlined Is-DecimalNumber / Is-ScriptBlock to avoid a function call per formatted value on this
+    # hot dispatch path (used for every <template> name expansion). The helpers stay defined for other callers.
+    if ($Value -is [float] -or $Value -is [single] -or $Value -is [double] -or $Value -is [decimal]) {
         return Format-Number2 -Value $Value
     }
 
-    if (Is-ScriptBlock -Value $Value) {
+    if ($Value -is [ScriptBlock]) {
         return Format-ScriptBlock2 -Value $Value
     }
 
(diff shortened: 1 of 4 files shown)


src/Pester.RSpec.ps1
fflaten, line 1: Just curious so I can learn, what's the big benefit here besides clearer names? We skip maybe 1-2 light function calls, but all the scriptblocks are reference types and passed on from the initial call.

src/Format2.ps1
```diff
@@ -122,11 +122,13 @@ function Format-Nicely2 ($Value, [switch]$Pretty, [int]$Depth = 0) {
         return Format-Type2 -Value $Value
     }
 
-    if (Is-DecimalNumber -Value $Value) {
+    # Inlined Is-DecimalNumber / Is-ScriptBlock to avoid a function call per formatted value on this
+    # hot dispatch path (used for every <template> name expansion). The helpers stay defined for other callers.
+    if ($Value -is [float] -or $Value -is [single] -or $Value -is [double] -or $Value -is [decimal]) {
```
fflaten, line 127: `Is-DecimalNumber` is only used here and `Format.ps1`. Maybe update both?

PR #2927, opened 2026-07-31
Show a compact string-diff message under ConciseView

### LINKED ISSUE #2872: String difference error broken when ErrorView is ConciseView

nohwnd:
Fix #2872

The string difference message from `Should-BeString` and `Should -Be`/`-BeExactly` is multi-line: the expected and actual length, the index where they differ, and a caret pointing at it. Under the default `ConciseView` this becomes an unreadable single line, because ConciseView collapses all the newlines. There is no message shape that survives it, I checked.

So under `ConciseView` we now emit a compact single line instead:

```
Expected strings to be the same, but they differ at index 2. Expected: 'ab' (length 2), but was: 'abc' (length 3).
```

Under `NormalView` the rich multi-line diff with the caret is unchanged. The new and the classic assertion produce the same message now.

Also fixes the caret in `Should-BeString`, it pointed one column to the left of the first difference. Now it points at the first differing character, same as `Should -Be` and as xunit/nunit/mstest.

Note: the collection and equivalence messages have the same ConciseView problem, that is a separate follow-up.

🤖

### DIFF
--- a/src/functions/assert/String/Should-BeString.ps1
+++ b/src/functions/assert/String/Should-BeString.ps1
@@ -143,6 +143,21 @@ function Get-StringDifferenceMessage {
 
     $because = if ($Because) { " because $Because," } else { "" }
 
+    $expectedExpanded = Expand-SpecialCharacters -InputObject $Expected
+    $actualExpanded = Expand-SpecialCharacters -InputObject $Actual
+
+    # ConciseView (the default $ErrorView since PowerShell 7.2) reflows the error message: it
+    # collapses every newline into a space and re-wraps to the console width. That turns the
+    # aligned multi-line diff, including the caret, into an unreadable single line (#2872). No
+    # message shape survives that reflow, so under ConciseView we emit a compact single-line
+    # message instead, which stays readable once flattened.
+    if ('ConciseView' -eq $ErrorView) {
+        if ($Expected.Length -ne $Actual.Length) {
+            return "Expected strings to be the same,$because but they differ at index $differenceIndex. Expected: '$expectedExpanded' (length $($Expected.Length)), but was: '$actualExpanded' (length $($Actual.Length))."
+        }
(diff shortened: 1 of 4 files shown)


nohwnd:
Closing, see #2872. This only affects running the assertion outside of Pester, inside a run we render the error ourselves and it looks fine. Keeping the caret fix separately in #2928.

PR #2947, opened 2026-07-31
Add -End switch to Mock for pipeline aggregation

### LINKED ISSUE #2154: Can't mock a function that takes pipeline input
## Describe your environment

Pester 5.3.1 on PSv5 and v7 on Win 10

## Expected Behavior

Both tests pass

## Current Behavior

Mock returns 1 object for each piped input

Steppable pipeline. For sample code, run `[System.Management.Automation.ProxyCommand]::Create([System.Management.Automation.CommandMetadata]::new((Get-Command New-Item)))`

nohwnd:
Fix #2154

Mocking a command that takes pipeline input and returns once from its end block never
worked. `MockWith` runs once per piped item, so a command that aggregates the pipeline
and returns a single object returns one object per input instead, and `Should -Invoke`
counts one call per item.

```powershell
function outer { 1..3 | inner }
function inner {
    param ([Parameter(ValueFromPipeline)] $InputObject)
    "RETURN"   # implicit end block, runs once
}

Mock inner { "FOO" }
@(outer).Count   # 3, should be 1
```

## The change

Add `-End` to `Mock`. With it the process block only collects the pipeline, and `MockWith`
runs once from the end block over all the input.

```powershell
Mock inner { "FOO" } -End
@(outer).Count   # 1
```

The collected input is piped into `MockWith`, so it is there as `$input` and `MockWith` can
have its own `begin`/`process`/`end` blocks, just like the real command.

```powershell
Mock Save-State { "saved $(@($input).Count) items" } -End
1..3 | Save-State   # "saved 3 items"
```

`Should -Invoke` counts one call per pipeline, and the parameter filter sees the whole
collected input bound to the pipeline parameter, so you can assert on all of it at once.

```powershell
Mock inner { "FOO" } -End
$null = outer
Should -Invoke inner -Times 1 -Exactly
Should -Invoke inner -Times 1 -Exactly -ParameterFilter { $InputObject -contains 2 }
```

## Design

The idea: one mock type per command.

- A per-call mock (today) runs `MockWith` per item and records one call per item.
- An `-End` mock collects the pipeline and records one call per pipeline.

All mocks for a single command must be the same type. Mixing the two would put two different
granularities in the call history and `Should -Invoke` could not count it coherently, so it is
rejected at definition time. That is what keeps `Should -Invoke` a plain counter, it does not
have to know about the two types, it just counts the history entries. Per-call mocks behave
exactly as before, `-End` is purely additive.

Because the type carries the granularity, the filter naturally sees the aggregate for an `-End`
mock, so there is no separate `-EndParameterFilter`.

## Verification

262 existing Mock tests still pass, plus 16 new ones in `Mock.EndBlock.Tests.ps1` (output shape,
per-pipeline counting, filter over the aggregate, `-Verifiable`, module scope, the homogeneity
guard). Built inline and ran on PS 7.5.

## Open questions, @fflaten LMK before I polish

- [ ] Name. I went with `-End` because that is what we said, but it reads a bit odd as a bare
      switch (`Mock inner { } -End`). `-Aggregate` / `-PipelineEnd` / `-OnEnd` are alternatives.
- [ ] Homogeneity is checked in the current scope only. A parent-scope `-End` mock and a
      child-scope per-call mock for the same command would slip through the definition-time check.
      Do we care, or add a resolution-time guard too?
- [ ] Native commands. Piping to a mocked application (e.g. `pwsh`, see #2154 powercode comment)
      has no `ValueFromPipeline` parameter, so nothing binds by name for the filter. The input is
      still collected and piped into `MockWith`, but you cannot assert on it by parameter name.
      Probably fine for v1, but worth calling out.
- [ ] Should we fold the tests into `Mock.Tests.ps1` instead of a separate file?

🤖

### DIFF
--- a/src/functions/Mock.ps1
+++ b/src/functions/Mock.ps1
@@ -16,7 +16,8 @@ function New-MockBehavior {
         [Parameter(Mandatory)]
         $Hook,
         [string[]]$RemoveParameterType,
-        [string[]]$RemoveParameterValidation
+        [string[]]$RemoveParameterValidation,
+        [Switch] $Aggregate
     )
 
     [PSCustomObject] @{
@@ -26,6 +27,9 @@ function New-MockBehavior {
         IsDefault   = $null -eq $ParameterFilter
         IsInModule  = -not [string]::IsNullOrEmpty($ContextInfo.TargetModule)
         Verifiable  = $Verifiable
+        # When set the whole pipeline is collected and MockWith runs once from the end block over
+        # all the input, instead of once per piped item. See the -End switch on Mock. (#2154)
+        Aggregate   = [bool]$Aggregate
         Executed    = $false
         ScriptBlock = $MockWith
         Hook        = $Hook
@@ -948,6 +952,32 @@ function Resolve-Command {
     }
 }
 
+function Get-PipelineParameterName {
+    # Return the names of the parameters that take pipeline input (ValueFromPipeline or
+    # ValueFromPipelineByPropertyName). Used by -End (aggregate) mocks to bind the whole collected
+    # pipeline to the pipeline parameter, so a parameter filter can assert on all the input. Native
+    # applications have no metadata and no such parameters, so this returns nothing for them. (#2154)
+    param ($Metadata)
+
+    if ($null -eq $Metadata -or $null -eq $Metadata.Parameters) {
+        return
+    }
+
+    foreach ($parameter in $Metadata.Parameters.Values) {
+        $takesPipeline = $false
+        foreach ($parameterSet in $parameter.ParameterSets.Values) {
+            if ($parameterSet.ValueFromPipeline -or $parameterSet.ValueFromPipelineByPropertyName) {
+                $takesPipeline = $true
+                break
+            }
+        }
+
+        if ($takesPipeline) {
+            $parameter.Name
+        }
+    }
+}
+
 function Invoke-MockInternal {
     [CmdletBinding()]
     param (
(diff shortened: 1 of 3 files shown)


fflaten:
This is great! 🥳 Maybe even default behavior some day. 

> * [ ]  Name. I went with `-End` because that is what we said, but it reads a bit odd as a bare
>   switch (`Mock inner { } -End`). `-Aggregate` / `-PipelineEnd` / `-OnEnd` are alternatives.

Yeah. I think even mentioning `End` is wrong when users can do `{ begin {} process {} end {} }`. It's closer to `-WithPipeline/AsPipeline/Advanced/AdvancedFunction` even though it's not 1:1. `-Aggregate` best of the above.

> * [ ]  Homogeneity is checked in the current scope only. A parent-scope `-End` mock and a
>   child-scope per-call mock for the same command would slip through the definition-time check.
>   Do we care, or add a resolution-time guard too?

No need if there's no runtime conflict in my opinion. Looks like the biggest side effect of mixing modes at different levels is confusing invocation count when using e.g. `Should -Invoke -Scope Describe`. The new `Should -Invoke` error with invocation history shows `-End` call with array value vs individual for normal, so I think users will figure it out.

> * [ ]  Native commands. Piping to a mocked application (e.g. `pwsh`, see [Can't mock a function that takes pipeline input #2154](https://github.com/pester/Pester/issues/2154) powercode comment)
>   has no `ValueFromPipeline` parameter, so nothing binds by name for the filter. The input is
>   still collected and piped into `MockWith`, but you cannot assert on it by parameter name.
>   Probably fine for v1, but worth calling out.

👍 

> * [ ]  Should we fold the tests into `Mock.Tests.ps1` instead of a separate file?

Separate - don't see any benefits of merging. `Mock.Tests.ps1` is already massive.

Issue #2425, opened 2024-02-27
Support shuffling / randomized test runs


twhiting:
### Checklist

- [X] Feature request has a meaningful title
- [X] I have searched the existing issues. See [all issues](https://github.com/pester/Pester/issues?q=is%3Aissue)
- [X] I have tested using the latest version of Pester. See [Installation and update guide](https://pester.dev/docs/introduction/installation).

### Summary of the feature request

Google test has a feature where tests can be shuffled randomly per test run. This is beneficial in integration tests scenarios when one test might fail as a result of another previously ran test. Frankly it helps find obscure bugs.

### How should it work?

See google test shuffle for a good description: http://google.github.io/googletest/advanced.html#shuffling-the-tests

nohwnd:
Interesting. That would a bit difficult (and inefficient) to provide, because Pester relies on execution that is "top-down", so if we mix tests that come from 2 different files we have to "restore" the state for each test. And run all setups and teardowns again, breaking some of the life cycles that people probably rely on. This would be specially painful for mocks I think. 

I am not against it completely, I would just like a bit more data showing why doing this is worth it.

fflaten:
Yes it would be limited to shuffling tests (`It`) inside the current block (`Context/Describe`). Is that enough to be useful?

Shuffling containers/files is also an option, though that can already be achieved manually.

twhiting:
Yes google test shuffles within a "test suite" which is the behavior that @fflaten describes.

twhiting:
It is hard to explain why it's worth it without being super abstract. I for example have a powershell module that interfaces with a windows service. It is possible that test A can cause a state within the service that test B exposes. It is VERY hard to track down these cases without a reliable randomizer.

The second part of this is needing to be able to run the tests again in the same shuffled order.

Google test can shuffle based on a seed. So say in CI a test fails when shuffled with seed X. I can then manually trigger a new run with seed X and get the same exact test run order for all tests.

twhiting:
For a good description of the seed mechanism see the google test link I posted above.

nohwnd:
Makes sense. If we randomize within the same container, in a top-down way (that is we randomize on each level, but dont jump up and down), then it is hardly any change from the current way of running tests, all you need to do is shuffle the tests in the discovered tree (in a deterministic way). I think there is even an “order” list already in the discovered tree.

@twhiting do you want to make a PR for this? Even if its just a proof of concept.

twhiting:
[...]

PR #2948, opened 2026-07-31
Add Run.Shuffle to shuffle test execution order

nohwnd:
Implements #2425.

`Run.Shuffle` shuffles the order things run in. It reorders the files, the Describes in a file, the Describes and Contexts in a Describe, and the Its in a block. Items only ever move within their own level.

```powershell
$config.Run.Path = 'tests'
$config.Run.Shuffle = $true
```

The order is repeatable. Each run picks a seed and prints it:

```
Shuffling execution order using seed 1738685315. Set 'Run.ShuffleSeed = 1738685315' to repeat this order.
```

Set the seed to replay the same order:

```powershell
$config.Run.ShuffleSeed = 1738685315
```

You don't have to read it off the screen. The seed that was used is on the result too, so you can grab it after a run and replay:

```powershell
$r = Invoke-Pester -Configuration $config -PassThru
$r.Configuration.Run.ShuffleSeed.Value   # 1738685315, even when you did not set one
```

## Opting out

Some tests are ordered on purpose, for example mock history that adds up across sibling `It`s. A file keeps its order with a comment, the same way it opts out of parallel:

```powershell
# pester:no-shuffle

Describe 'Deploy' {
    It 'builds'  { ... }
    It 'deploys' { ... }
}
```

Its blocks and tests then run in declaration order even when `Run.Shuffle` is on. A block level switch like `Describe -Ordered { }` could follow, keeping that out of this PR for now.

## How it works

The seed is resolved once in `Invoke-Pester`, so it is reported once, shared by the parallel workers, and returned on the result. The files are shuffled in `Invoke-Test`, and each block's children in `PostProcess-DiscoveredBlock`, where `.Blocks` and `.Tests` are rebuilt from the shuffled order before `First`/`Last` are marked, so the one time setup and teardown still wrap the real first and last item.

## Tried it on Pester itself

I ran Pester's own tests with `Run.Shuffle = $true`. It found a few order coupled tests, all the expected kind, mock history that counts calls across sibling `It`s, and tests that check the exact setup and teardown sequence. Marking `tst/functions/Mock.Tests.ps1` with `#pester:no-shuffle` takes it from 6 failures back to 0 under a shuffled run, so the opt-out does its job.

Tests are in `tst/Pester.RSpec.Shuffle.ts.ps1`.

Note: the opt-out keeps a file's own order, but the file still moves in the file order shuffle. And in parallel each worker shuffles its own slice of files, so cross file order is not globally reproducible there. The within file shuffle and the seed are.

🤖


nohwnd:
@fflaten curious what you think about this one, mainly about the opt-out.

Right now a whole file can opt out with `#pester:no-shuffle`, same style as `#pester:no-parallel`. But shuffle happens at every level, not just per file, so a file level opt-out feels too coarse on its own. I am thinking a block level switch as a follow up, something like `Describe -Ordered { }` / `Context -Ordered { }` to keep just that block's children in declaration order. Does that direction feel right to you, or would you do it differently? Keeping it out of this PR for now.

🤖

tst/Pester.RSpec.Shuffle.ts.ps1
```diff
@@ -0,0 +1,317 @@
...
+            # Reduce each entry to its top-level Describe and keep the order they first appear in.
+            $shuffled = Get-ExecutionOrder -Shuffle -Seed 42
+            $topLevel = @($shuffled.Order | ForEach-Object { ($_ -split '\.')[0] } | Select-Object -Unique)
+            (($topLevel -join ',') -ne 'A,B,C') | Verify-True
+        }
+
+        t "reorders the Its inside a block" {
+            # Seeds are chosen so the C block's tests are not in declaration order.
+            $found = $false
+            foreach ($seed in 1..20) {
```
fflaten: Use a known seed instead of brute force? 🙂

tst/Pester.RSpec.Shuffle.ts.ps1
```diff
@@ -0,0 +1,317 @@
...
+                $shuffled = Get-ExecutionOrder -Shuffle -Seed $seed
+                $cTests = @($shuffled.Order | Where-Object { $_ -like 'C.*' })
+                if (($cTests -join ',') -ne 'C.c1,C.c2,C.c3') { $found = $true; break }
+            }
+            $found | Verify-True
+        }
+
+        t "reorders the Its inside a nested Context" {
+            $found = $false
+            foreach ($seed in 1..20) {
```
fflaten: Same as above

tst/Pester.RSpec.Shuffle.ts.ps1
```diff
@@ -0,0 +1,317 @@
...
+# pester:no-shuffle
+Describe 'Ordered' {
+    It 'o1' { $global:__order2.Add('o1') }
+    It 'o2' { $global:__order2.Add('o2') }
+    It 'o3' { $global:__order2.Add('o3') }
+    It 'o4' { $global:__order2.Add('o4') }
+}
+'@
+            foreach ($n in 'Free1', 'Free2', 'Free3') {
+                Set-Content -Path (Join-Path $folder "$n.Tests.ps1") -Value "Describe '$n' { It 'a' { 1 | Should -Be 1 }; It 'b' { 1 | Should -Be 1 } }"
```
fflaten: Assert that at least one of them was shuffled? As a implementation regression test so a single no-shuffle directive doesn't disable globally.

fflaten (changes requested):

fflaten:
> @fflaten curious what you think about this one, mainly about the opt-out.
> 
> Right now a whole file can opt out with `#pester:no-shuffle`, same style as `#pester:no-parallel`. But shuffle happens at every level, not just per file, so a file level opt-out feels too coarse on its own. I am thinking a block level switch as a follow up, something like `Describe -Ordered { }` / `Context -Ordered { }` to keep just that block's children in declaration order. Does that direction feel right to you, or would you do it differently? Keeping it out of this PR for now.
> 
> 🤖

I think we should pick one, not both. Either we support:
- Block level granularity (with inheritance) using `-Ordered/-NoShuffle`. For most users could be a single top-level `Describe`
- Or container level only. User could split their tests out in a separate container. The feature is opt-in, so requiring a separate container is not unreasonable for highly coupled tests.

Leaning towards block level for flexibility. The directive syntax is new (experimental) and might confuse as we still shuffle the container order - unlike no-parallel which runs it last IIRC.

nohwnd:
Problem with the opt out on block level is that the blocks themselves can be shuffled, so there is no clear place where we should run the unshuffled tests? maybe at the start of the file? But what if the opt-out is only on the second level of multi level describe, then what, should the whole parent not shuffle? 

This is why I chose to implement it on file level, because I don't have good answers for these questions.

tst/Pester.RSpec.Shuffle.ts.ps1
nohwnd: Good point, switched to a fixed seed (7) that puts the C block out of declaration order, no more brute-force loop.

🤖

tst/Pester.RSpec.Shuffle.ts.ps1
nohwnd: Same here, seed 7 also reorders the nested Context, so it asserts the exact order now.

🤖

tst/Pester.RSpec.Shuffle.ts.ps1
nohwnd: Added. The Free files now record their own order and the test asserts they actually shuffled, so a single `#pester:no-shuffle` disabling shuffle for the whole run would fail here.

🤖

nohwnd:
Let's keep just the file-level `#pester:no-shuffle` opt-out for now. Block level is where the semantics get unclear: the blocks themselves get shuffled, so there is no obvious place to run an opted-out block, and a directive on a nested block raises the question whether the parent should stop shuffling too. I don't have good answers for those yet.

The whole feature is experimental (marked it as such now), so we can add block granularity later if it turns out we need it, without breaking anyone.

🤖

fflaten:
> Problem with the opt out on block level is that the blocks themselves can be shuffled, so there is no clear place where we should

The idea was that `-NoShuffle` affects the child nodes, not the block itself. A single root-level `Describe 'd' -NoShuffle` would then disable at file-level

nohwnd:
Yes but what I tried to outline is: 
```
Describe top {
    Describe a { }
    Describe b -NoShuffle { 
       Describe c { 
        }
    }
}
```

Will `b` run first or second? Because neither `a` or `top` describes have `-NoShuffle`.

That is what makes it difficult to provide consistent behavior.

fflaten:
Understood, and expected a/b order to be random in that case.

A clear description for `-NoShuffle` like "Always guarantee execution order for child blocks and tests. This block and data order can still be shuffled.". Low cost fix for users to add a wrapper block like `top` if they need them fixed, considering this is all opt-in to begin with.

Whether we should shuffle `-ForEach` data or not is likely more confusing. Some might expect at least `b1, b2, b3, a`, similar to JUnit and GTest which according to Google AI overview keeps parameter order.

We can try it out and ask for input. I would just like to avoid multiple controls for the same feature in the end - should we support block-level exclusion 🙂

PR #2954, opened 2026-08-06
Add -NormalizeNewline to Should-BeString

### LINKED ISSUE #2952: Add -NormalizeLineEndings to Should-BeString
### Checklist

- [x] Feature request has a meaningful title

For tests that will run on multiple platforms, it would be nice to ignore the difference between `\n` and `\r\n` (and maybe `\r` and Unicode line endings). The `-IgnoreWhitespace` switch won't work for tests that need to confirm the location of the newlines, and preserve indents. Git settings may or may not convert the source (reference) value literal to the native line endings of the current platform.

### How should it work?

When used, the reference value and difference value are compared as if `-replace '\r\n|\r|\p{Zl}',([char]0xA)` had been used on both first.

nohwnd:
Fix #2952

Adds `-NormalizeNewline` to `Should-BeString`. When set, both sides are run through `-replace '\r\n|\r|\p{Zl}', "`n"` before comparing, so `\r\n`, `\r` and the Unicode line separator all count as `\n`. Unlike `-IgnoreWhitespace` it keeps the newlines and their positions, so indentation and blank lines are still compared. Added the same switch to `Should-NotBeString`, to keep the pair consistent with `-IgnoreWhitespace` and `-TrimWhitespace`.

@fflaten I am not happy with the name yet, which of these do you like most:

- `-NormalizeNewline` (what I used here)
- `-NormalizeLineEnding`

Names I avoided, and why:

- `-IgnoreNewline` / `-IgnoreLineEnding`, reads as "drop the newlines", which is what `-IgnoreWhitespace` already does (`-replace '\s'`). Here we keep them, we only unify the style, so `Ignore` is misleading.
- plural `-NormalizeNewlines` / `-NormalizeLineEndings`, the existing switches use singular collective nouns (`Whitespace`), so singular fits better.

Between the two I lean slightly to `-NormalizeLineEnding` because it clearly covers `\r` and the Unicode separator too, but `-NormalizeNewline` is shorter and still clear. Happy to rename before merge.

Tests cover CRLF, CR, LF and the Unicode line separator comparing equal, and that content differences and extra blank lines still fail.

🤖

### DIFF
--- a/src/functions/assert/String/Should-BeString.ps1
+++ b/src/functions/assert/String/Should-BeString.ps1
@@ -4,11 +4,17 @@
         $Actual,
         [switch]$CaseSensitive,
         [switch]$IgnoreWhitespace,
-        [switch]$TrimWhitespace
+        [switch]$TrimWhitespace,
+        [switch]$NormalizeNewline
     )
 
     if ($Actual -isnot [string]) {
         return $false
+    }
+
+    if ($NormalizeNewline) {
+        $Expected = $Expected -replace '\r\n|\r|\p{Zl}', "`n"
+        $Actual = $Actual -replace '\r\n|\r|\p{Zl}', "`n"
     }
 
     if ($IgnoreWhitespace) {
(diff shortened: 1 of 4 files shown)


nohwnd:
merging so I can release, still interested in your opinion

fflaten:
Prefer `-NormalizeLineEnding` as more generic and correct, but just nitpicking. Everyone will understand either way.

PR #2956, opened 2026-08-06
Hint that -ExpectedMessage uses wildcards when the message looks identical

### LINKED ISSUE #1793: ExpectedMessage comparison fails
# General summary of the issue

Error message comparison fails.

## Describe your environment

```
Pester version     : 5.1.0-beta1 C:\Users\Chris\Documents\PowerShell\Modules\Pester\5.1.0\Pester.psm1
PowerShell version : 7.1.0
```

## Steps to reproduce

```ps
{ @(1, 2) - 1 } | Should -Throw -ExpectedMessage "Method invocation failed because [System.Object[]] does not contain a method named 'op_Subtraction'."
```

## Expected Behavior

The assertion should be successful.

Using try / catch to perform the same comparison is successful:

```ps
try {
    @(1, 2) - 1
} catch {
    $_.Exception.Message | Should -Be "Method invocation failed because [System.Object[]] does not contain a method named 'op_Subtraction'."
}
```

## Current Behavior

The assertion fails.

nohwnd:
`Should -Throw -ExpectedMessage` matches with `-like`, so `[ ] * ?` are wildcards. When the actual message is identical to the expected one except for those characters, the match fails, but the failure message printed the same text on both sides:

```
Expected an exception with message like 'value is [1]' to be thrown, but the message was 'value is [1]'.
```

Same text on both sides, no hint why it failed. This bites people migrating from v4, where the message match used `.Contains`, and it is the cause of a few issues and StackOverflow questions.

Now that case adds a note:

```
Note: -ExpectedMessage matches using wildcards (-like). The messages are identical except for the
wildcard characters [ ] * ? in -ExpectedMessage. Escape them with a backtick (`[) or use
[System.Management.Automation.WildcardPattern]::Escape() to match them literally.
```

The note only shows when the actual message equals the expected one treated literally, so it does not fire on genuine mismatches or on intentional wildcards that just did not match.

Fixes #1793

🤖

### DIFF
--- a/src/functions/assertions/PesterThrow.ps1
+++ b/src/functions/assertions/PesterThrow.ps1
@@ -97,11 +97,18 @@
     }
 
     $filterOnMessage = -not [string]::IsNullOrWhitespace($ExpectedMessage)
+    $messageFailedOnWildcard = $false
     if ($filterOnMessage) {
         $unescapedExpectedMessage = [System.Management.Automation.WildcardPattern]::Unescape($ExpectedMessage)
         $filters += "message like $(Format-Nicely $unescapedExpectedMessage)"
         if ($actualExceptionWasThrown -and (-not (Get-DoValuesMatch $actualExceptionMessage $ExpectedMessage))) {
             $buts += "the message was $(Format-Nicely $actualExceptionMessage)"
+            # -ExpectedMessage matches with -like. When the actual message is identical to the expected
+            # one treated literally, the only reason the match failed is unescaped wildcard characters
+            # ([ ] * ?) in -ExpectedMessage. Flag it so the failure message is not baffling (#1793).
+            if ($actualExceptionMessage -eq $unescapedExpectedMessage) {
+                $messageFailedOnWildcard = $true
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/functions/assertions/PesterThrow.ps1
```diff
@@ -120,7 +127,13 @@
     if ($buts.Count -ne 0) {
         $filter = Join-And $filters
         $but = Join-And $buts
-        $failureMessage = "Expected an exception$(if($filter) { " with $filter" }) to be thrown,$(Format-Because $Because) but $but. $actualExceptionLine".Trim()
+        $wildcardHint = if ($messageFailedOnWildcard) {
+            "$([System.Environment]::NewLine)    Note: -ExpectedMessage matches using wildcards (-like). The messages are identical except for the wildcard characters [ ] * ? in -ExpectedMessage. Escape them with a backtick (``[) or use [System.Management.Automation.WildcardPattern]::Escape() to match them literally."
```
fflaten, line 131: Use same format as hints in v6 assertions?
```powershell
"$formattedMessage`n`nHint: $hint"
```

fflaten (commented):
We need the same for `Should-Throw`. Maybe make `Hint` a optional parameter in `Fail` so we can override the default `Get-AssertionGotcha` in special cases like this.

PR #2970, opened 2026-08-10
Hint at unescaped wildcards on Should-Throw -ExceptionMessage

### LINKED ISSUE #2966: Use same format as hints in v6 assertions?
Use same format as hints in v6 assertions?
```powershell
"$formattedMessage`n`nHint: $hint"
```

_Originally posted by @fflaten in https://github.com/pester/Pester/pull/2956#discussion_r3731521796_

### LINKED ISSUE #2968: Expand ExpectedMessage wildcard hint to Should-Throw
We need the same for `Should-Throw`. Maybe make `Hint` a optional parameter in `Fail` so we can override the default `Get-AssertionGotcha` in special cases like this.

_Originally posted by @fflaten in https://github.com/pester/Pester/pull/2956#pullrequestreview-4877815683_

nohwnd:
`Should-Throw` matches `-ExceptionMessage` with `-like`, so `[ ] * ?` are wildcards. When the actual message is identical to the expected one treated literally, the match fails only because those characters were not escaped, and the failure shows two messages that look the same. This is the v6 counterpart of #2956, which added the same hint to the v5 `Should -Throw`.

To carry the hint I added an optional `Hint` key to the `ShouldAssertion.Fail` data (issue #2968 suggested making `Hint` overridable). When present it overrides the default `Get-AssertionGotcha` hint and is appended with the standard `Hint: <text>` format the other v6 assertions already use, so #2966 is covered without a new formatting path.

```
Expected an exception, with message like 'value is [1]' to be thrown, but the message was 'value is [1]'.

Hint: -ExceptionMessage matches using wildcards (-like). The messages are identical except for the wildcard characters [ ] * ? in -ExceptionMessage. Escape them with a backtick (`[) or use [System.Management.Automation.WildcardPattern]::Escape() to match them literally.
```

Fixes #2968
Fixes #2966

🤖

### DIFF
--- a/src/functions/assert/Common/New-ShouldAssertion.ps1
+++ b/src/functions/assert/Common/New-ShouldAssertion.ps1
@@ -74,18 +74,23 @@ class ShouldAssertion {
             $actual = $this.Collected['Actual']
         }
         $because = [string] $Data['Because']
+        # An explicit Hint overrides the default Get-AssertionGotcha hint, for assertions that have a
+        # more specific thing to say about their own failure (e.g. Should-Throw explaining that the
+        # -ExceptionMessage filter failed only because of unescaped wildcard characters). Like the
+        # other reserved keys it is not turned into a <key> token.
+        $hintOverride = [string] $Data['Hint']
 
-        # Everything except the three special keys becomes a <key> token in the message.
+        # Everything except the reserved keys becomes a <key> token in the message.
         $extra = @{}
         foreach ($key in $Data.Keys) {
-            if ($key -ne 'Expected' -and $key -ne 'Actual' -and $key -ne 'Because') {
+            if ($key -ne 'Expected' -and $key -ne 'Actual' -and $key -ne 'Because' -and $key -ne 'Hint') {
                 $extra[$key] = $Data[$key]
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/functions/assert/Common/New-ShouldAssertion.ps1
```diff
@@ -74,18 +74,23 @@ class ShouldAssertion {
...
+        # -ExceptionMessage filter failed only because of unescaped wildcard characters). Like the
+        # other reserved keys it is not turned into a <key> token.
+        $hintOverride = [string] $Data['Hint']
 
-        # Everything except the three special keys becomes a <key> token in the message.
+        # Everything except the reserved keys becomes a <key> token in the message.
         $extra = @{}
         foreach ($key in $Data.Keys) {
-            if ($key -ne 'Expected' -and $key -ne 'Actual' -and $key -ne 'Because') {
+            if ($key -ne 'Expected' -and $key -ne 'Actual' -and $key -ne 'Because' -and $key -ne 'Hint') {
```
fflaten:
```suggestion
            if ($key -notin 'Expected', 'Actual', 'Because', 'Hint') {
```

PR #2973, opened 2026-08-12
Make Run.ParallelThrottleLimit = 1 run sequentially in the current session

nohwnd:
Comes out of #2971, where @DarkLite1 described the pattern he uses in his own code: one function that runs a scriptblock over items, and when you set `-ThrottleLimit 1` it drops to a plain `foreach` on the main thread so you can step through it. Same knob for concurrency and for debugging.

`Run.ParallelThrottleLimit = 1` did not do that. It ran one file at a time, but still inside a worker runspace, so you got no speedup and no debugging: breakpoints are per runspace, a breakpoint set in the session that calls `Invoke-Pester` never hits inside a worker.

Now 1 falls back to the normal sequential path, next to the fallbacks we already have (Windows PowerShell 5.1, `ScriptBlock` containers, `Run.SkipRemainingOnFailure = 'Run'`). The banner also stops saying "in parallel", because the run is not parallel.

No new configuration option, and no new switch on `Invoke-Pester`. Turn the throttle you already have down to 1 and you can debug the file.

Before:

```powershell
Invoke-Pester -Configuration @{ Run = @{ Parallel = $true; ParallelThrottleLimit = 1 } }
# file 1 ran in runspace 2, breakpoint hits: 0
```

After:

```powershell
Invoke-Pester -Configuration @{ Run = @{ Parallel = $true; ParallelThrottleLimit = 1 } }
# file 1 ran in runspace 1, breakpoint hits: 1
```

Tests added in `tst/Pester.RSpec.Parallel.ts.ps1`: the breakpoint hits, the banner does not claim parallel, and the counts match a parallel run. `test.ps1 -File tst/Pester.RSpec.Parallel.ts.ps1` passes 25/25, `PesterConfiguration.Tests.ps1` and `Help.Tests.ps1` pass, and `Invoke-ScriptAnalyzer` on `src/Main.ps1` reports the same 9 pre-existing findings as `main`.

Open question: someone might want 1 file at a time but still isolated in a runspace, to tell an isolation problem apart from a concurrency problem. That is what this takes away. I think the debugging case is worth more, LMK if you disagree.

🤖


nohwnd:
Not the direction we want to take. Enabling Run.Parallel means the files run in runspaces, throttling to 1 just means one runspace at a time. If you don't want parallel, don't enable it.

🤖

PR #2994, opened 2026-08-22
Run Pester.BeforeContainer.ps1 for test files only

nohwnd:
`Pester.BeforeContainer.ps1` is a setup for test files, dot-sourced before each one, and a ScriptBlock container is not a file. It was running for those too.

That makes the cost grow with something nobody expects. A test that runs `Invoke-Pester` over a scriptblock pays for the whole repo's setup on every nested run, and a suite that does that a few hundred times pays it a few hundred times. Ours does: with a `Pester.BeforeContainer.ps1` in the repo root the file ran 621 times instead of 97, and cost 5.4s of a 17s unit run. With this it is back to 11.1s against 10.7s without the file at all.

The stray output test used a ScriptBlock container to get the repo-root setup to emit into the success stream. The container type is incidental to what it checks, so it uses a file now.

Found while trying to turn `Run.Parallel` on for our own suite, which needs a `Pester.BeforeContainer.ps1` to work at all.

🤖

### DIFF
--- a/src/Pester.Runtime.ps1
+++ b/src/Pester.Runtime.ps1
@@ -1,4 +1,4 @@
-# PESTER_BUILD
+﻿# PESTER_BUILD
 if (-not (Get-Variable -Name "PESTER_BUILD" -ValueOnly -ErrorAction Ignore)) {
     . "$PSScriptRoot/Pester.Utility.ps1"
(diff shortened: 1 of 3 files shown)


nohwnd:
this changes product to cater to our tests, nope.

PR #2995, opened 2026-08-22
Add -Because to It, Describe and Context to say why a test is skipped

### LINKED ISSUE #2049: Feature: Add a -Reason parameter when using the -Skip switch

Hey, I wonder if you can add What is the reason a test was skipped

the current return object you will get if you using ```-PassThru``` doesn't contain any info on this
maybe you can  add parameter name ```-Reason``` 'the reason it skipped'

```powershell
PS D:\Github\pwshchakraui> $results.Skipped[0]             


Name              : Has help comment block
Path              : {New-Tab, Comment Base Help, Has help comment block}
Data              :
ExpandedName      : Has help comment block
ExpandedPath      : New-Tab.Comment Base Help.Has help comment block
Result            : Skipped
ErrorRecord       : {}
StandardOutput    :
Duration          : 00:00:00.0054751
ItemType          : Test
Id                :
ScriptBlock       :
                                    $FilePath | Should -FileContentMatch "<#"
                                    $FilePath | Should -FileContentMatch "#>"

Tag               :
Focus             : False
Skip              : True
Block             : [!] Comment Base Help
First             : True
Last              : False
Include           : False
Exclude           : False
Explicit          : False
ShouldRun         : True
StartLine         : 9
Executed          : True
ExecutedAt        : 8/14/2021 11:43:21 AM
Passed            : True
Skipped           : True
UserDuration      : 00:00:00
FrameworkDuration : 00:00:00.0054751
PluginData        :
FrameworkData     :
```  

```powershell

Describe "describe1" {
    Context "with one skipped test" {
        It "test 1" -Skip -Reason 'i want it to skip' {
            1 | Should -Be 2
        }

        It "test 2" {
            1 | Should -Be 1
        }
    }

    Describe "that is skipped" -Skip -Reason 'i want it to skip all the tests' {
        It "test 3" {
            1 | Should -Be 2
        }
    }

    Context "that is skipped and has skipped test" -Skip -Reason 'i want all the It tests to skip' {
        It "test 3" -Skip {
            1 | Should -Be 2
        }

        It "test 3" {
            1 | Should -Be 2
        }
    }
}

```

So the return object will look something like this.
```powershell
Name              : Has help comment block
Path              : {New-Tab, Comment Base Help, Has help comment block}
Data              :
ExpandedName      : Has help comment block
ExpandedPath      : New-Tab.Comment Base Help.Has help comment block
Result            : Skipped
ErrorRecord       : {}
StandardOutput    :
Duration          : 00:00:00.0054751
ItemType          : Test
Id                :
ScriptBlock       :
                                    $FilePath | Should -FileContentMatch "<#"
                                    $FilePath | Should -FileContentMatch "#>"

Tag               :
Focus             : False
Skip              : True
Block             : [!] Comment Base Help
First             : True
Last              : False
Include           : False
Exclude           : False
Explicit          : False
ShouldRun         : True
StartLine         : 9
Executed          : True
ExecutedAt        : 8/14/2021 11:43:21 AM
Passed            : True
Skipped           : True
```
```powershell
SkippedReason  : I want it to skip
```
```powershell
UserDuration      : 00:00:00
FrameworkDuration : 00:00:00.0054751
PluginData        :
FrameworkData     :
```

nohwnd:
Fix #2049

Picks up @corbob's work from #2607 and finishes it. His commit is kept as-is, the rest is on top.

```powershell
It 'talks to the API' -Skip -Because 'the API is down' { }
Describe 'integration' -Skip -Because 'needs Azure' { }
```

```
$r.Skipped[0].Reason  ->  the API is down
```

The parameter is `-Because` rather than `-Reason`, so the codebase uses one word for why: `Set-ItResult -Because` and `Should -Because` already do. The result property stays `.Reason`, which is the term NUnit and JUnit use.

## What changed on top of #2607

`Set-ItResult` keeps building its message the way it did. #2607 replaced the whole message with the reason, which dropped the `is skipped, because ...` shape and is what turned its CI red. The reason is pulled back out of the message for `.Reason`, so it holds the reason on its own no matter which way the test was skipped, and `Set-ItResult -Inconclusive` records one too.

The reason reaches all three report formats, not only NUnit2.5. An explicit reason wins over the error message in NUnit3's `reason` element and in JUnit's `skipped` element. Verified all four cases land in `NUnitXml`, `NUnit3` and `JUnitXml`.

Tests for the result object on every path: `It -Skip -Because`, `-Skip` with no reason, a skipped `Describe` or `Context` passing its reason down, and both `Set-ItResult` kinds. Help added for the new parameter.

## Open

@fflaten noted in #2049 that NUnit3-console also writes `<property name="_SKIPREASON" ... />` next to the `reason` element. Not done here, say the word and I will add it.

🤖


nohwnd:
Not super shure about this chanbge. @fflaten yes or no?

PR #3014, opened 2026-08-29
Bump to 6.2.0-alpha1 and start docs/NEXT-RELEASE.md

nohwnd:
Release notes for the version in development live in the repo now, in `docs/NEXT-RELEASE.md`. The idea is that every PR that changes something a user can notice adds its entry there in the same commit, so the notes are versioned next to the change instead of being reconstructed from the commit log at the end. The file grows through the alphas and is copied into the release body as it is, the maintainer block at the top is an HTML comment so it does not render.

Filled it with the notes for 6.2.0-alpha1, which is everything since the 6.1.0 tag:

- the `Pester.BeforeContainer.ps1` cascade from #2993, with a full folder example and the three behavior changes it brings
- the `Should-BeString` region diff from #3010
- strict configuration values and the unknown-key warning from #2982
- error locations and assertion names from #2981, and the real error from a throwing `BeforeAll` from #3005
- parallel on Windows PowerShell 5.1 from #2987, and the lost-file and aborted-run fixes from #3012 and #3004

Every example in the notes was run against this build, the output blocks are pasted from the real run.

Version: `ModuleVersion` 6.2.0, `Prerelease` alpha1, `RequiredAssemblyVersion` 6.2.0, and `ReleaseNotes` points at the 6.2.0-alpha1 tag. Minor bump because there are features in there, not only fixes.

## Verification

`./build.ps1 -Clean` then `./test.ps1` on PS 7.5 macOS: P phase 23 blocks 0 failed, RSpec 2921 passed, 0 failed, 3 skipped. `Pester v6.2.0-alpha1` in the banner, assembly 6.2.0.0, informational version 6.2.0-alpha1.

🤖


docs/NEXT-RELEASE.md
```diff
@@ -0,0 +1,469 @@
...
+This is a prerelease. Install it with `Install-Module Pester -AllowPrerelease`.
+
+`Pester.BeforeContainer.ps1` grew from one file at the repository root into a chain that
+follows your folder structure, so unit tests and integration tests can each have their own
+setup without repeating it in every test file. `Should-BeString` prints a real diff when a
+long string does not match. The configuration now tells you when you handed it a value it
+cannot use, instead of quietly ignoring it. And the experimental parallel runner works on
+Windows PowerShell 5.1, which is the slowest edition and the one that needed it most.
+
+Pester 6 runs on **Windows PowerShell 5.1** and **PowerShell 7.4+**.
```
nohwnd: no change here, is more confusing than helpful

docs/NEXT-RELEASE.md
```diff
@@ -0,0 +1,469 @@
...
+is applied, outermost first:
+
+```
+reporoot/Pester.BeforeContainer.ps1                 <- applies to everything
+reporoot/tests/Pester.BeforeContainer.ps1           <- applies to tests/ and below
+reporoot/tests/unit/Pester.BeforeContainer.ps1      <- applies to tests/unit only
+reporoot/tests/integration/Pester.BeforeContainer.ps1
+```
+
+The root file holds what everything needs:
```
nohwnd: this applies only to tests as well, so it is better to say "holds common setups and teardowns that every test needs"

docs/NEXT-RELEASE.md
```diff
@@ -0,0 +1,469 @@
...
+Fix [#3011](https://github.com/pester/Pester/issues/3011). The import race that killed the
+worker in the first place is the other half of that issue and is not fixed yet, I could not
+reproduce it locally in 288 staggered imports, 1600 worker imports across 200 real parallel
+runs, or 400 imports timed onto the five second `validVerbs` removal.
+
+## <a id="6.2.0-behavior-changes"></a>Behavior changes
+
+Three of these come with the `Pester.BeforeContainer.ps1` cascade above.
+
+- **Two `BeforeAll` (or `AfterAll`, `BeforeEach`, `AfterEach`) in one block no longer throw.**
```
nohwnd: describe it in a positive way. saying that you can use multiple same blocks, not that they no longer throw.

docs/NEXT-RELEASE.md
```diff
@@ -0,0 +1,469 @@
...
+  a run where every file opts out with `#pester:no-parallel`.
+- A `TestRegistry` retry that recovers no longer prints
+  `IO exception during a TestRegistry operation, retrying.`. The operations it wraps are
+  idempotent, and when the retry does not recover the exception propagates, which is the real
+  signal. It is a debug message now.
+- `Should-BeFasterThan` and `Should-BeSlowerThan` no longer flake on shared CI machines. Both
+  assertions bound the measured time from above, and measuring a scriptblock times compiling
+  it, GC, and whatever else the machine is doing, which has no upper bound on a shared runner.
+
+## <a id="6.2.0-repository"></a>Repository
```
nohwnd: this whole part is not relevant to users. remove

docs/NEXT-RELEASE.md
nohwnd: Removed.

🤖

docs/NEXT-RELEASE.md
nohwnd: Reworded to "The root file holds common setups and teardowns that every test needs".

🤖

docs/NEXT-RELEASE.md
nohwnd: Rewritten as a capability, "A block can have as many BeforeAll, AfterAll, BeforeEach and AfterEach as you want", with an example that groups setup by what it sets up. The note about why it used to throw moved below the example.

🤖

docs/NEXT-RELEASE.md
nohwnd: Removed the whole section, kept the Full Changelog link at the end of Other improvements and fixes.

🤖

PR #3018, opened 2026-08-30
Run CI on a pull request whatever branch it targets

nohwnd:
Both workflows filtered `pull_request` by base branch, `ci.yml` to `main`, `rel/*` and `dev/*`, and `code-analysis.yml` to `main`. A pull request opened against anything else matched neither, so it got no checks at all. A pull request with no checks does not look broken, it looks fine, which is worse than a red one.

That is what happened to #3015. It was stacked on the `release-6.2.0-alpha1` branch so it could add its entry to `docs/NEXT-RELEASE.md`, and it sat there with zero signal until I dispatched both workflows by hand against the branch.

## Why drop the filter instead of adding a pattern

Adding another pattern, `stack/*` or similar, only helps when you remember to name the branch correctly, and forgetting is silent. That is the same trap. Any pull request is worth testing, whatever it targets.

`push` keeps its filter. That one is about which branches are worth building on their own, and `main` plus `rel/*` is still the right answer there. So no new runs from pushes, only from pull requests that previously got nothing.

## Verification

`pull_request:` with no value parses as "every pull request":

```
ci.yml            -> {'push': {'branches': ['main', 'rel/*'], ...}, 'pull_request': {'paths-ignore': [...]}, 'workflow_dispatch': None}
code-analysis.yml -> {'push': {'branches': ['main']}, 'pull_request': None, 'workflow_dispatch': None}
```

`paths-ignore` on `ci.yml` is untouched.

I am also opening a throwaway pull request stacked on this branch to confirm the checks actually appear, and will close it once it has reported. Result in a comment below.

## Separate, not fixed here

A pull request that only touches `docs/**` or `**/*.md` still matches `paths-ignore` and runs nothing, so `Done` never reports. If `Done` is a required check that would block a docs-only pull request from merging. I have not checked the branch protection settings, so I do not know whether it actually bites.

🤖


nohwnd:
Verified. #3019 was a throwaway pull request based on `ci/pr-trigger-any-base` rather than on `main`, which is exactly the shape that got nothing before. Checks appeared on it within 15 seconds:

```
Build            pending
PSScriptAnalyzer pending
```

Both workflows fired, so both file changes work, not just the `ci.yml` one. For a `pull_request` event GitHub reads the workflow file from the merge commit, so the branch under test activates its own trigger change, which is what makes this testable before merging.

Closed #3019 and deleted its branch rather than letting the full 8 leg matrix run on a blank line.

🤖

PR #3019, opened 2026-08-30
TEMPORARY: prove a stacked PR gets checks

nohwnd:
Throwaway. Base is ci/pr-trigger-any-base, not main, which is exactly the case that got no checks before #3018. If checks appear on this PR, the fix works. Closing as soon as it reports.

🤖


nohwnd:
Did its job, checks appeared within 15 seconds. Proof recorded on #3018.

🤖

PR #3020, opened 2026-08-30
Set LASTEXITCODE when Invoke-Pester fails internally

pathei-kosmos:
`Invoke-Pester` could overwrite the exit status of an internal error with `$null` when the run was not created yet, or with `0` when a plugin failed after an otherwise successful run.

This change tracks the process exit code separately from the test failure count so internal errors keep the existing `-1` exit code without changing how test failures are reported.

Regression tests cover both failure paths.


nohwnd:
Ty!

PR #3022, opened 2026-09-05
Render test results in the CI run instead of a privileged workflow

nohwnd:
Removes the one place where the output of a fork pull request reaches a writable token, so we can stop approving fork runs.

## Problem

`test-report.yml` runs on `workflow_run`. That means it runs in the base repository with a writable token (`checks: write`), and what it reads are artifacts that the fork pull request produced, parsed by a third-party action. A fork controls `ci.yml` inside its own pull request, so it controls the artifact names and the XML that this privileged job consumes.

We added that workflow recently, and it is the only reason a fork pull request has to be approved before it runs. The rest is already fine: `ci.yml` is `contents: read` and gets a read-only token from a fork, no secrets reach fork runs, there is no `pull_request_target`, every runner is GitHub-hosted, and `backport.yml` verifies the commenter has write access before it does anything.

Approving runs is not a real boundary anyway. Someone can send a few good pull requests, earn the trust, and then send a bad one. And approval is a click, not a diff review.

## Changes

- Delete `test-report.yml`. The `Done` job already downloads every `results-*` artifact for the coverage table, so it renders the failed tests there too. Writing to `GITHUB_STEP_SUMMARY` needs no token, so a fork pull request gets the same summary we get.
- Pin every action to a commit SHA. `dotnet/arcade` backport-base was on `@main` and was the loosest reference we had, with `contents: write`. It needs a manual bump now.
- Skip the SARIF upload in `code-analysis.yml` for fork pull requests. A read-only token cannot write code scanning results, so once fork runs start without approval that step would fail on every one of them. The analysis itself still runs.

## What is lost

The per-leg check runs named `Tests - ps7-ubuntu-latest` in the checks list. Failures are now a table plus a collapsible list per leg in the CI run summary, next to the coverage table.

## Verification

Ran the new summary step locally with pwsh against NUnit3 fixtures: normal failures, a leg with no failures, a malformed XML file, and no results at all. A malformed leg prints `error` in its row instead of throwing, and the step never fails the job, so `All test legs passed` is still what fails the build. All three workflow files parse.

## After this merges

Settings, Actions, General: set fork pull request approval to the weakest option, and check that sending write tokens and sending secrets to fork pull requests are both off.

@fflaten what do you think? The check runs were nice and I am trading them for not having a writable token near fork output. If you would rather keep the reporting, the other option is running the reporter inside `ci.yml`, which works for our pull requests and quietly does nothing for forks.

🤖


.github/workflows/ci.yml
```diff
@@ -207,19 +214,77 @@ jobs:
...
           pattern: results-*
-          path: coverage-reports
+          path: results
+
+      # Render the failed tests into this run's summary. This is what test-report.yml
+      # used to do as check runs; doing it here keeps the token read-only and works
+      # for fork pull requests, which cannot create check runs.
+      # Only failures are listed. A full pass list is thousands of lines and GitHub
+      # truncates the summary.
+      - name: Test results
```
nohwnd, line 244: script into file?

nohwnd: Done. Both summary scripts are in `.github/scripts` now and take `-Path` and `-SummaryPath`, so they run against downloaded artifacts locally. Verified both against the artifacts from the last green run here.

The sparse checkout in Done has to run before the artifact download, because checkout cleans the working directory and would delete the results otherwise.

Left the per-leg coverage summary inline, those jobs have no checkout on purpose, they run against the build artifact.

🤖

PR #3033, opened 2026-09-05
Check out YAML with LF so Dependabot stops rewriting every line

nohwnd:
Our global rule is `* text=auto eol=crlf`: the repository stores LF, the working tree gets CRLF. Dependabot checks out the working-tree form, edits it, and commits those bytes without re-normalizing, so its blobs come back CRLF against our LF and every line reads as changed.

Both Dependabot pull requests we have ever had look like this:

| PR | diff | real changes |
|---|---|---|
| #2986 (closed) | 328 insertions | 11 |
| #3027 (open) | whole file | 4 |

There is no Dependabot setting for line endings, and `text eol=crlf` cannot store CRLF anyway, since `text` always normalizes to LF on the way in. So the fix has to be on our side.

`eol=lf` for YAML makes the checkout and the stored form agree. **No blob changes**, they are already LF, and none of the 11 YAML files are PowerShell or test fixtures. For reference, pester/vscode-adapter and PowerShell/PSScriptAnalyzer already use `eol=lf` repo-wide, and PowerShell/PowerShell uses the same targeted-override pattern for `*.sh` and a few scripts.

Verified with `git ls-files --eol`: after a fresh checkout the workflow files are `i/lf w/lf`, and `git add --renormalize .` touches nothing.

Once this lands, `@dependabot recreate` on #3027 regenerates it with a clean diff.

🤖


fflaten:
Related bugs:
- dependabot/dependabot-core#8693
- dependabot/dependabot-core#10193

PR #3035, opened 2026-09-05
Install PSScriptAnalyzer directly instead of via psmodulecache

nohwnd:
Same change as #3034 on main.

`sha_pinning_required` applies to actions referenced **inside** composite actions. `potatoqualitee/psmodulecache` does not pin its own `actions/cache` reference - v6.2.1 uses `@v4.2.0`, v6.3 uses `@v5.0.5` - so the job fails at Set up job before running anything.

This branch matters because its workflows trigger on `rel/*`, so they actually run and would break exactly the way main did. rel/6.0.x has the same line but its code-analysis only triggers on `main`, so it never fires there.

The step only installed PSScriptAnalyzer and ConvertToSARIF, which take seconds from PSGallery.

Upstream fix sent as potatoqualitee/psmodulecache#68, but we should not depend on it landing.

🤖


nohwnd:
Parking this as a draft. `sha_pinning_required` is off again so nothing is blocked, and the module cache step predates this work, so I want to discuss it before removing it.

The reason enforcement and psmodulecache cannot coexist is in the description above, and the upstream fix is potatoqualitee/psmodulecache#68.

🤖

nohwnd:
Closing in favour of #3036, where the reasoning and the options are written up.

Removing the module cache is only one of four ways to handle this, and it is not obviously the right one until we know what the cache was solving. The policy is off, so nothing is blocked.

🤖

